PR #36, opened 2021-01-29
Switch to Github Actions CI

Derollez:
First unoptimized attempt to use Github actions for CI Pipeline instead of Travis CI. Uses make commands for now. Can be further optimized to use actions and docker containers directly in main.yml file.


Derollez:
TBR

PR #45, opened 2021-07-20
fix: J4 propagation

robinpdm:


lucas-bremond (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Models/Kepler.cpp
```diff
@@ -742,16 +745,16 @@ State                           Kepler::CalculateJ4StateAt                  (
         1.0 + expr * (2.0 + eccentricityAtEpochSquared / 2.0 - 2.0 * sqrtBeta
```
Derollez, line 745: @robinpdm just checking but the Beta here is not meant to be the Beta variable defined in Vallado correct ?

robinpdm: Here, `sqrtBeta = std::sqrt(1.0 - eccentricityAtEpoch * eccentricityAtEpoch)`

PR #48, opened 2021-11-07
chore: update OSTk Physics

lucas-bremond:


lucas-bremond (commented):
Update Mathematics and Physics.

PR #56, opened 2022-05-24
chore: remove the old notebooks under docs that were previously used for unit testing

apaletta3:
As discussed with lucas, remove deprecated notebooks


Derollez:
@apaletta3 @robinpdm These notebooks were going deep into the available methods on the ostk objects in python for the new user. Today, tutorials in the base repository of ostk are rather high level. Open question: would it make sense to add an updated version of some of these notebooks into the `tutorials` folder of the ostk base repo? Or do you think they don't bring too much information? (we should check that we can tab-complete and get proper info on the methods in python using `__help__` maybe?)

PR #69, opened 2022-09-22
misc: suggestions propagated

Derollez:


apaletta3 (commented):
Agree with all the changes, but would like to modify one thing to make the unit tests pass cause right now its not passing since there is a seg fault due to something that @Derollez you changed

PR #80, opened 2022-11-10
feat: add new Propagator intermediate class and overloaded method to calculate_state(s)_at

vishwaLoft:
Reasoning:
* The current calculate_state(s)_at methods update a cache state array. This concept works great for when we're propagating a single trajectory like a set of post-fit states to get the most accurate state in the fastest time possible.
* However, if we're using the propagator with multiple different initial conditions, (like in shooting methods or monte carlos etc.) then we're populating this cached state array that is "polluted" with multiple sets of trajectories.

Solution:
* A clean method that propagates from the initial state to the final time in a more general way.
* Add `Propagator` class to hold `SatelliteDynamics` and `NumericalSolver` data.
* Move (initial) `State` from `SatelliteDynamics` in `Propagated` (constructed with a `State` or an `Array<State>`. 
* Add fundamental methods for `calculate_state(s)_at` to `Propagator`.


Derollez (approved):
See suggestions commits on top! LGTM

PR #87, opened 2022-12-27
feat: add CCSDS CDM class and utilities

Derollez:
- Implement a general (not super flexible) definition of Conjunction Data Message (CDM). This class and utilities provide a general toolbox to create and extract data from multiple input formats and frame them within the CCSDS standards
- `Dictionary` function relies on strings values (as this is the output format given by SpaceTrack client in python). This **should** be generalized in the future. 
- The current Schema is not really flexible and only takes care of mandatory fields in the CCSDS white paper. 
- Some additional structures (`Covariance`, `Object`, `Organization`) can be added in the future to organize and extract data in a more robust and clear way. 
- Goal is to merge this first flaky implementation and test it out with a variety of data input


include/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+        CDM::Header             getHeader                                   ( ) const ;
+
+        CDM::RelativeMetadata   getRelativeMetadata                         ( ) const ;
+
+        Array<CDM::Metadata>    getMetadataArray                            ( ) const ;
+
+        Array<CDM::Data>        getDataArray                                ( ) const ;
+
+        CDM::Metadata           getObjectMetadataAt                         (   const   Size&                       anIndex                                     ) const ;
```
lucas-bremond, line 233: There's an Index type you can use

include/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+        Length                  getMissDistance                             ( ) const ;
+
+        Position                getRelativePosition                         ( ) const ;
+
+        Velocity                getRelativeVelocity                         ( ) const ;
+
+        Real                    getCollisionProbability                     ( ) const ;
+
+        String                  getCollisionProbabilityMethod               ( ) const ;
```
lucas-bremond, line 261: Just checking: no enum for this?

include/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.hpp
```diff
@@ -0,0 +1,321 @@
...
+
+        // Object Metadata getters. TBI: Create Resident Space Object (RSO) class and add these methods under it
+
+        Integer                 getObjectDesignator                         (   const   Size&                       anIndex                                     ) const ;
+
+        String                  getObjectName                               (   const   Size&                       anIndex                                     ) const ;
+
+        String                  getObjectInternationalDesignator            (   const   Size&                       anIndex                                     ) const ;
+
+        ObjectType              getObjectType                               (   const   Size&                       anIndex                                     ) const ;
```
lucas-bremond, line 271:
```suggestion
        CDM::ObjectType         getObjectType                               (   const   Index&                      anIndex                                     ) const ;
```

src/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.cpp
```diff
@@ -0,0 +1,747 @@
...
+
+Integer                         CDM::getObjectDesignator                    (   const   Size&                       anIndex                                     ) const
+{
+
+    if (!this->isDefined())
+    {
+        throw ostk::core::error::runtime::Undefined("CDM") ;
+    }
+
+    CDM::Metadata objectMetadata = this->getObjectMetadataAt(anIndex) ;
```
lucas-bremond, line 268:
```suggestion
    const CDM::Metadata objectMetadata = this->getObjectMetadataAt(anIndex) ;
```
 And batch apply

src/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.cpp
```diff
@@ -0,0 +1,747 @@
...
+        Real::Undefined(),
+        Real::Undefined(),
+        Real::Undefined(),
+        Real::Undefined(),
+        Mass::Undefined(),
+        Real::Undefined(),
+        Real::Undefined(),
+        Real::Undefined(),
+        Real::Undefined(),
+        State(
```
lucas-bremond, line 585:
```suggestion
        State
        (
```
 if we want to keep Allman consistency + apply to rest

src/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.cpp
```diff
@@ -0,0 +1,747 @@
...
+            Instant::DateTime(DateTime::Parse(aDictionary["TCA"].getString(), DateTime::Format::ISO8601), Scale::UTC),
+            Position::Meters(  // TBI: Add Position::Kilometers
+                {
+                    Real::Parse(aDictionary["SAT1_X"].accessString()) * 1000.0,
+                    Real::Parse(aDictionary["SAT1_Y"].accessString()) * 1000.0,
+                    Real::Parse(aDictionary["SAT1_Z"].accessString()) * 1000.0
+                },
+                Frame::ITRF()
```
lucas-bremond, line 593: Do you confirm this is the right frame to use here? (Since another frame seems used for another Position field)

test/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.test.cpp
```diff
@@ -0,0 +1,862 @@
...
+//     //             Angle::Degrees(3.0),
+//     //             Angle::Degrees(4.0),
+//     //             123.456
+//     //         }
+//     //     }
+//     // } ;
+
+// }
+
+// TEST_F (OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Messages_CCSDS_CDM, IsDefined)
```
lucas-bremond, line 218: Why not enabling?

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Conjunction/Messages/CCSDS/CDM.cpp
```diff
@@ -0,0 +1,330 @@
...
+        .def_static("dictionary", &CDM::Dictionary, arg("dictionary"))
+        .def_static("parse", &CDM::Parse, arg("string"))
+        .def_static("load", &CDM::Load, arg("file"))
+        .def_static("object_type_from_string", &CDM::ObjectTypeFromString, arg("string"))
+
+    ;
+
+    enum_<CDM::ObjectType>(cdm, "ObjectType")
+
+    .value("PAYLOAD", CDM::ObjectType::Payload)
```
lucas-bremond, line 88: Padding + enum style?

lucas-bremond (approved):

include/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.hpp
Derollez: Not right now indeed. From Annex E in https://public.ccsds.org/Pubs/508x0b1e2c2.pdf, it seems like there are a couple of methods supported. Yet, unfortunately a lot of users don't necessarily respect the format, so I think keeping it general with a `String` is fine, especially because this is informational for now (at least in the first MVP to fetch and display Conjunction)

src/OpenSpaceToolkit/Astrodynamics/Conjunction/Messages/CCSDS/CDM.cpp
Derollez: So it is not necessarily `ITRF` in the general case indeed. What we should do is check based on the value of the other `refFrame` field.

lucas-bremond: Yeah let's, and just fail at loading time if the frame is not something OSTK can support as of today.

Derollez: My plan is to do a separate pass on those Frame/Position/Velocity/Covariance fields (might need to add other Frame functions) and adapt where needed. I'll experiment with the other fields for now

lucas-bremond: Right, my only ask is to not have implicit hardcoding because we may/will forget to revisit, and this can lead to subtle bugs. So if we only support ITRF for now, totally fine, but let's blatantly fail if the CDM is expressed in a different frame.

Derollez: Yep okay, that makes total sense. I'll modify this :)

Derollez: Adapted to support `ITRF`, `GCRF` and `J2000` based on options enumerated on standard at https://public.ccsds.org/Pubs/508x0b1e2s.pdf.

PR #88, opened 2023-01-03
chore:  use spaces to pad revolution number for TLE sanitization

Derollez:
- Spaces and Zeroes are both valid to pad for revolution number. Aligning to use spaces to match SpaceTrack convention. 
- Zeroes are used for International Designator and Epoch. Already verified.


lucas-bremond (approved):

vishwa2710 (approved):
Thanks for adding extra tests!

robinpdm (approved):

PR #105, opened 2023-05-19
build: update OSTk dependencies to latest

Derollez:


Derollez:
Using last release from Physics https://github.com/open-space-collective/open-space-toolkit-physics/pull/104 before merging

bindings/python/requirements.txt
```diff
@@ -1,3 +1,3 @@
-# Apache License 2.0 
+# Apache License 2.0
 
-open-space-toolkit-physics~=0.5,>=0.5.27
+open-space-toolkit-physics~=0.8.0
```
lucas-bremond, line 3: This is locking the deps to `open-space-toolkit-physics~=0.8,<0.9`. Which makes sense until we are in the `0.x.y` version realm.

Maybe we should actually move to a `1.x.y` range?

Derollez: > Which makes sense until we are in the 0.x.y version realm

Yeah that was the goal for now

> Maybe we should actually move to a 1.x.y range?

Yeah I think we should probably transition to see soon

PR #106, opened 2023-05-21
Draft: feat: add dummy maneuvers and add forward prop at Propagator level

Derollez:
-  Very WIP. 
- Add dummy maneuvers and add forward prop at Propagator level, not optimized, WIP + jupyter notebook on 3.10. 
- Just playing with concepts a bit.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Maneuver.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Maneuver.hpp
@@ -1,0 +1,77 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Trajectory_Maneuver__
+#define __OpenSpaceToolkit_Astrodynamics_Trajectory_Maneuver__
+
+#include <OpenSpaceToolkit/Core/Types/Shared.hpp>
+
+#include <OpenSpaceToolkit/Mathematics/Objects/Vector.hpp>
+
+#include <OpenSpaceToolkit/Physics/Coordinate/Frame.hpp>
+#include <OpenSpaceToolkit/Physics/Coordinate/Position.hpp>
+#include <OpenSpaceToolkit/Physics/Coordinate/Velocity.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace trajectory
+{
+
+using ostk::core::types::Shared;
+
+using ostk::physics::coord::Frame;
+using ostk::physics::coord::Velocity;
+using ostk::physics::time::Instant;
+using ostk::physics::time::Duration;
+
+/// @brief                      Trajectory maneuver
+///                             Dummy Model, to be improved
+
+class Maneuver
+{
+   public:
+
+    // Add Type or Child Class
+
+    Maneuver(const Instant& anInstant, const Duration& aDuration, const Velocity& aVelocityDelta);
+
+    bool operator==(const Maneuver& aManeuver) const;
+
+    bool operator!=(const Maneuver& aManeuver) const;
+
+    friend std::ostream& operator<<(std::ostream& anOutputStream, const Maneuver& aManeuver);
+
+    bool isDefined() const;
+
+    const Instant& accessInstant() const;
+
+    const Duration& accessDuration() const;
+
+    const Velocity& accessVelocityDelta() const;
+
+    Instant getInstant() const;
+
+    Duration getDuration() const;
+
+    Velocity getVelocityDelta() const;
+
+    // Maneuver inFrame(const Shared<const Frame>& aFrameSPtr) const;
+
+    void print(std::ostream& anOutputStream, bool displayDecorator = true) const;
+
+    static Maneuver Undefined();
+
+   private:
+    Instant instant_;
+    Duration duration_;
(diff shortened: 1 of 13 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Maneuver.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+using ostk::physics::coord::Frame;
+using ostk::physics::coord::Velocity;
+using ostk::physics::time::Instant;
+using ostk::physics::time::Duration;
+
+/// @brief                      Trajectory maneuver
+///                             Dummy Model, to be improved
+
+class Maneuver
```
vishwa2710, line 32: Worth making this an abstract parent, we should have impulsive and finite burn maneuvers.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -69,6 +72,18 @@ class Propagator
...
+    ///
+    /// @code
+    ///                     Propagator propagator = { aSatelliteDynamics, aNumericalSolver, aManeuverArray } ;
+    /// @endcode
+    ///
+    /// @param              [in] aSatelliteDynamics A satellite dynamics object
+    /// @param              [in] aNumericalSolver A numerical solver
+    /// @param              [in] aManeuverArray A maneuver array
+
+    Propagator(const SatelliteDynamics& aSatelliteDynamics, const NumericalSolver& aNumericalSolver, const Array<Maneuver>& aManeuverArray);
```
vishwa2710, line 85: This is the way.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -127,13 +135,74 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
         }
     }
 
+    // TBI: Dummy implementation at Propagator level, it should probably not live at this level
+    // Check size of this->maneuverArray_ before continuing
+
+    Array<Maneuver> forwardInstantManeuvers;
+    Array<Maneuver> backwardInstantManeuvers;
+
+    for (const Maneuver& aManeuver : this->maneuverArray_)
```
vishwa2710, line 144: I think this logic should be broken out, we should make a struct called PropagatedChunks or something which basically is a start time, endtime/duration, initial state, and maneuver (potentially undefined maneuver)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -127,13 +135,74 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
...
     Array<SatelliteDynamics::StateVector> propagatedForwardStateVectorArray;
-    if (!forwardInstants.isEmpty())
+    Instant startInstant = aState.getInstant();
+    if (!forwardInstantPairs.isEmpty())
     {
-        propagatedForwardStateVectorArray = numericalSolver_.integrateStatesAtSortedInstants(
-            startStateVector, aState.getInstant(), forwardInstants, satelliteDynamics_.getDynamicalEquations()
-        );
+
+        for (const auto& instantManeuverPair: forwardInstantPairs)
```
vishwa2710, line 178: Over here then we can just look through each PropagatorChunk (name TBD ofcourse). Which can be a separate class method. I think that separates the logic well and makes it testable

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -127,13 +135,74 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
...
+                );
+
+                startStateVector = propagationVector;
+                startInstant = propagationInstant;
+
+                propagatedForwardStateVectorArray.add(propagationVector);
+            }
+            else
+            {
+                Vector3d deltaV = maneuver.getVelocityDelta().getCoordinates();  // Check Frame, link with attitude provider
```
vishwa2710, line 199: This logic should be abstracted out into a "Dynamics" class object rather than added directly so that we can leverage finite burn vs impulsive in an easy way.

vishwa2710 (commented):
Just left some comments, I know this just a spike so I'm sure you're already thinking about all the stuff I've mentioned but just thought I'd put it down so we can discuss and see if we are aligned. Great stuff!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
Derollez: I like that comment.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
Derollez: Definitely, we can make the logic in there modular

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
Derollez: I see what you mean: that makes sense

PR #108, opened 2023-06-01
style: small whitespace fixes after headers

kyle-cochran:


lucas-bremond:
Thanks!! Are you sure about the removal of `.gitkeep`? (not against, just making sure this was intentional)

lucas-bremond (approved):

PR #118, opened 2023-06-19
feat: split gravitational dynamics

vishwa2710:
I think this is a better interface but wanted to get some thoughts first.
Might also be worth adding a PositionIntegrator (name TBD) like Pau mentioned in FDEP-003 to integrate positions.


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/System/Dynamics.cpp
```diff
@@ -15,7 +15,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_System_Dynamics(pybind11::mod
     using ostk::astro::flight::system::Dynamics;
 
     class_<Dynamics, Shared<Dynamics>>(aModule, "Dynamics")
-        .def_static("get_dynamical_equations", &Dynamics::GetDynamicalEquations, arg("instant"), arg("dynamics_array"))
+        .def_static("get_dynamical_equations", &Dynamics::getDynamicalEquations, arg("instant"), arg("dynamics_array"))
```
lucas-bremond, line 18: C++ static methods start with capital letter I think no?

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics.hpp
```diff
@@ -61,21 +61,35 @@ class Dynamics
 
     virtual void print(std::ostream& anOutputStream, bool displayDecorator = true) const = 0;
 
-    /// @brief              Obtain dynamical equations function wrapper (pure virtual)
+    /// @brief              Get dynamical equations function wrapper (pure virtual)
     ///
     /// @return             std::function<void(const std::vector<double>&, std::vector<double>&, const double)>
 
+    static DynamicalEquationWrapper getDynamicalEquations(
```
lucas-bremond, line 68: Because it's static, move down and capitalize method name?

Derollez: Oups! Yeah totally, that's a typo

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+using ostk::physics::env::obj::Celestial;
+using ostk::physics::time::Instant;
+
+using ostk::astro::flight::system::SatelliteSystem;
+using ostk::astro::flight::system::Dynamics;
+
+/// @brief                      Define the acceleration experienced by a point mass due to atmospheric drag
+
+class AtmosphericDrag : public Dynamics
```
Derollez, line 31: So I know it will be heavy but while re-adding those, I was thinking about `AtmosphericDragDynamics` vs `AtmosphericDrag` and `CentralBodyGravityDynamics` etc to show the type of object right away (and know it's really a Dynamics and not a Force or else) and avoid confusing for end users. But it would indeed make the interface simpler to use this current naming. Thoughts? @vishwa2710 @robinpdm @lucas-bremond

vishwa2710: Yeah fair point, I removed the `*Dynamics` because for other inherited objects we don't really use the parent name.
Like for Orbits, we don't have `KeplerOrbit`, `PropagatedOrbit` etc.

Same with Earth in the various models etc.

IMO if there are conflicts in naming we can always name with the `using A = *` in C++ and `import B as A` in python.

But I'm fine either way as long as we're consistent across the repo.

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics.hpp
```diff
@@ -61,21 +61,35 @@ class Dynamics
...
+        const Instant& anInstant, const Array<Shared<Dynamics>>& aDynamicsArray
+    );
+
+    /// @brief              Update the state derivative (pure virtual)
+    ///
+    /// @param              [in] x A state vector
+    /// @param              [in] dxdt A state derivative vector
+    /// @param              [in] anInstant An instant
+
     virtual void update(const StateVector& x, StateVector& dxdt, const Instant& anInstant) = 0;
```
Derollez, line 78: Thoughts on naming here? Fine by me but just want to be sure it conveys the right concept

vishwa2710: Made a poll

PR #128, opened 2023-06-29
refactor: flexible state size

phc1990:


include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.hpp
```diff
@@ -99,6 +100,12 @@ class AtmosphericDrag : public Dynamics
 
     SatelliteSystem getSatelliteSystem() const;
 
+    /// @brief              Declares the coordinates involved in the contribution calculation
+    ///
+    /// @param              [in] coordinatesBroker The coordinate broker mapping state coordinates and indexes
+
+    virtual void declareCoordinates(const Shared<CoordinatesBroker>& coordinatesBroker) override;
```
kyle-cochran, line 107:
```suggestion
    void declareCoordinates(const Shared<CoordinatesBroker>& coordinatesBroker) override;
```

nit. Does this need to be virtual?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -189,6 +189,7 @@ class Propagator
     );
 
    private:
+    Shared<CoordinatesBroker> coordinatesBroker_ = std::make_shared<CoordinatesBroker>();
```
kyle-cochran, line 192:
```suggestion
    const Shared<CoordinatesBroker> coordinatesBroker_ = std::make_shared<CoordinatesBroker>();
```

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/PositionDerivative.hpp
```diff
@@ -60,6 +60,12 @@ class PositionDerivative : public Dynamics
 
     virtual bool isDefined() const override;
 
+    /// @brief              Declares the coordinates involved in the contribution calculation
+    ///
+    /// @param              [in] coordinatesBroker The coordinate broker mapping state coordinates and indexes
+
+    virtual void declareCoordinates(const Shared<CoordinatesBroker>& coordinatesBroker) override;
```
kyle-cochran, line 67:
```suggestion
    virtual void declareCoordinates(const Shared<CoordinatesBroker>& coordinatesBroker) const override;
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
```diff
@@ -0,0 +1,82 @@
...
+    CoordinatesBroker();
+
+    /// @brief              Adds a coordinates subset to be considered, returning the starting index it will occupy (or
+    /// that it occupies if it was already added) in the state coordinates
+    ///
+    /// @param              [in] aCoordinatesSubset a coordinate subset to be considered
+    ///
+    /// @return             The starting index of the subset in the state coordinates
+
+    Integer addSubset(const CoordinatesSubset& aCoordinatesSubset);
```
kyle-cochran, line 42: Consider using the `Size` type from `core` for indices since it's unsigned
```suggestion
    Size addSubset(const CoordinatesSubset& aCoordinatesSubset);
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
```diff
@@ -0,0 +1,80 @@
...
+    /// @brief              Constructor
+    ///
+    /// @code
+    ///                     CoordinatesSubset coordinatesSubset("POSITION", 3) ;
+    /// @endcode
+    ///
+    /// @param              [in] anId The coordinate subset unique identifier
+    /// @param              [in] aSize The coordinate subset size
+
+    CoordinatesSubset(const String& anId, const Integer& aSize);
```
kyle-cochran, line 33:
```suggestion
    CoordinatesSubset(const String& anId, const Size& aSize);
```

Same here. Consider using the `Size` type

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    bool operator==(const CoordinatesSubset& aCoordinatesSubset) const;
+
+    bool operator!=(const CoordinatesSubset& aCoordinatesSubset) const;
+
+    bool isDefined() const;
+
+    String getId() const;
+
+    Integer getSize() const;
```
kyle-cochran, line 43:
```suggestion
    Size getSize() const;
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    static CoordinatesSubset Position();
+
+    static CoordinatesSubset Velocity();
+
+    static CoordinatesSubset Mass();
+
+   private:
+    String id_;
+    Integer size_;
```
kyle-cochran, line 55:
```suggestion
    Size size_;
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    /// @brief              Returns the total number of coordinate subsets
+    ///
+    /// @return             The total number of coordinate subsets
+
+    Integer getNumberOfSubsets() const;
+
+   private:
+    Integer nextIndex_;
+    std::unordered_map<CoordinatesSubset, Integer> map_;
```
kyle-cochran, line 75: There's a `Map` type in Core that you may be able to leverage here:
https://github.com/open-space-collective/open-space-toolkit-core/blob/main/include/OpenSpaceToolkit/Core/Containers/Map.hpp 

I think it would let you avoid the hashing stuff in CoordinatesBroker.hpp

kyle-cochran (commented):
A few nits. Looking good dude!

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.hpp
vishwa2710: It needs to be virtual since we're overriding it explicitly I believe.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
```diff
@@ -0,0 +1,82 @@
...
+    Integer getNumberOfCoordinates() const;
+
+    /// @brief              Returns the total number of coordinate subsets
+    ///
+    /// @return             The total number of coordinate subsets
+
+    Integer getNumberOfSubsets() const;
+
+   private:
+    Integer nextIndex_;
```
vishwa2710, line 74: Worth using `Index` here from ostk::core::types::Index

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
vishwa2710, line 75: Let's call this something less generic like `coordinateIndexMap`

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -34,6 +34,10 @@ Propagator::Propagator(const NumericalSolver& aNumericalSolver, const Array<Shar
     : dynamics_(aDynamicsArray),
       numericalSolver_(aNumericalSolver)
 {
+    for (Shared<Dynamics> aDynamics : this->dynamics_)
+    {
+        aDynamics->declareCoordinates(coordinatesBroker_);
+    }
```
vishwa2710, line 40: Instead of storing the various indices in the dynamics objects. Should we return the contribution for each dynamic and then sum it in the propagator? That way the dynamics don't need to keep track of what the individual indices etc. are.

We can update the function signatures so that it accepts a `State` instead of `x` and `dxdt`. That way it can access the relevant values from the `State` object and return a coordinate vector of accelerations or positions or whatever it is (with the relevant coordinate subset).

vishwa2710 (commented):
Let me know if you want to discuss my suggestion further

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.hpp
phc1990: It is being overriden yes
(copied the virtual func pattern from `applyContribution`)

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/PositionDerivative.hpp
phc1990: Cannot make it const as it might modify the instance (for instance, storing internal indexes, see `AtmosphericDrag` example implementation)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
phc1990: Oh God, thanks for this 👍

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
phc1990: Ok 👍 
See if this would make sense to you/align with what you suggested:

1. Declare coordinates would still exist, and that particular for loop above, would stay as is. At the end of the day, this is a way for each `Dynamics` (or `CoordinatesResolver` for that matter in the future) to state "ey, these are the subsets I play with (either read/write direction)". This happens only once, whenever a `Dynamics` (or `CoordinatesResolver` eventually) is added, and it defines the size of the `State` (so that a dev working at a `Propagator` level doesn't have to worry).

2. `addSubset` function of the `CoordinatesBroker` would no longer return an index and therefore `Dynamics` would no longer store any indexes.

3. Update the `applyCoordinates` signature to use `State`, and map State to/from std::vector inside the `Propagator`, so that it complies with boost's API.

This is aligned with the direction of travel I was working on, which was to:

1. Make the `CoordinatesBroker` a shared pointer (as there will be one per Propagator).

2. Update the current `State` constructor `State(instant, coordinates, frame);` (which currently holds only because there is a hard-coded assumption that coordinates is {pos, vel}) to `State(instant, coordinates, frame, broker);`. At that point in time the `State` instance has all it needs to remain generic: the `coordinates` as well as the `broker` to interpret them.

3. The optimisation is then pushed on to the `State` implementation to make the best use of the `broker` to extract coordinates subsets (for instance, you do not want to fetch for the position index everytime you do `state.getPosition()`) -> but this I will not worry for now.

vishwa2710: How about this?
1. Dynamics has an Array of CoordinateSubsets, `readCoordinateSubsets` and `writeCoordinateSubsets`. These are hardcoded static values for the class. For example, `AtmosphericDrag` will have a `readCoordinateSubsets` = `{Position, Velocity}` and  `writeCoordinateSubsets` = `{Velocity}`
2. We update `Dynamics` from `applyContribution` to `getContribution`. Where `getContribution` is defined as follows
VectorXd getContribution(const State& aState);
We can access the Position and Velocity from the State object. And the instant as well. (This might not be very performant... but we'll have to see. It's a nice clean interface, but we should make sure our propagation is performant as that is the bread and butter to everything.)
3. We can then put together the contributions in the DynamicalEquations function.
```
void Dynamics::DynamicalEquations(
    const Dynamics::StateVector& x,
    Dynamics::StateVector& dxdt,
    const double& t,
    const Array<Shared<Dynamics>>& aDynamicsArray,
    const Instant& anInstant
)
{
    for (Index i = 0; i < dxdt.size(); ++i)
    {
        dxdt[i] = 0;
    }

    const Instant nextInstant = anInstant + Duration::Seconds(t);
    const State newState = {nextInstant, x, gcrfSPtr_};  // TBI: This could be whatever frame we're integrating in, in the future.

    for (const Shared<Dynamics>& dynamics : aDynamicsArray)
    {
        const VectorXd contributions = dynamics->getContribution(newState);
        for (const CoordinateSet& coordinateSet : dynamics->getWriteContributions)
            Index subsetIndex = CoordinateBroker->getSubsetIndex(coordinateSet);
            for (Index i = 0; i < coordinateSet.getSize(); ++i)
            {
                 dxdt[i + subsetIndex] += contributions[i];
            }
    }
}
```
Something like that, if that makes sense.
Additionally, before running `calculateStatesAt` or `calculateStateAt` we confirm that the State has all the subset coordinates required by the unique set of each dynamics' `readCoordinateSubsets` so that we're sure all the information to propagate exists.

Sorry if this is confusing, github isn't the best place to discuss, we can catch up on this next week!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
phc1990: done!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
phc1990: done!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
phc1990: Done!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
phc1990: Updated to also use `Size` and `Index` for all functions where applicable.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
phc1990: Updated both to meaningful names:

image

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
phc1990: done!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
phc1990: I've tried, but I think it forces me to implemente the `<` operator between Subsets (which cannot be sorted as there is no notion of lesser/greater than between them).

Could it be that `Map.hpp` assumes an underlying sorted map?

phc1990: (I think it does include sorting):

image

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
phc1990: Thanks Viswha, ok I see 👍 
Your approach makes sense and I think it should work.

The only concern I have is performance (as you mentioned).

- For the `readCoordinatesSubsets`, it's fine, it's only called once at the beginning of the propagation.

It's the writing part that I'm more concerned about:

- Every time we do `getContribution(newState)`, each dynamic will have to extract the information that it's interested it from the state. For instance, if it needs position, it's going to do something like 'newState.getPosition()`. Now, that is assuming that there is a `getPosition()` (which exists now, it sort of breaks the 'general state concept' but it's ok as it's too much change to removed `getPosition()` and/or `getVelocity()`. Then what happens inside the `State` instance when `getPosition()` is called? Was it previously stored as a `Vector3d` already, or does it use it's internal `CoordinateBroker` to fetch and extract from the `VectorXd` coordinates? (this is why I was saying that a lot of the optimization would be pushed on to the `State`, nothing wrong with, I just don't know we will come up with something that will at least match the performance of 'not even calling these every time'.
- Every time we do `getWriteContributions` what will happen? Are we returning always a new array or do we create it once and always return the same. If we want to be performant, maybe we created it once, then store it in the dynamics and always return it.
- Every time we do `CoordinateBroker->getSubsetIndex` a map search is performed. 

So, that was my concern. In summary:
- `getContribution(newState)` -> a lot happens here, and it will happen `n x m` times, where `n` is the number of steps and `m` is the number of dynamics.
- `getWriteContributions` -> also happening `n x m` times, but since it always yield the same result (this to me is quite key), we can circumvent by storing once at the beginning and just returning the same value.
- `CoordinateBroker->getSubsetIndex` -> this happens roughly `n x m x p` times where `p` is the number of `write coordinates subsets` each dynamics has. And just like the one above it always returns the same values.

Sorry about the long message. I agree with you maybe this is not the best to discuss, so I'll wait so that we can discuss over a call. But I struggle to see how we are going to beat the 'storing indexes' approach. Maybe not the most beautiful one, but remember indexes stay hidden inside the `Dynamic` implementation. Nobody else knows about it.

I can also update the other 3 dynamics to use indexes and we see how much it takes and measure against the other approach. But on paper it looks like it would at most match it, but I am no Cpp expert so... 🤷

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State.test.cpp
```diff
@@ -14,17 +16,33 @@ using ostk::physics::time::Instant;
...
     }
 
+    // {
+    //     const Instant instant = Instant::DateTime(DateTime(2018, 1, 1, 0, 0, 0), Scale::UTC);
+    //     std::vector<double> coordinates = {1.0, 2.0, 3.0, 4.0, 5.0, 6.0};
+    //     const Shared<const CoordinatesBroker> broker = std::make_shared<const CoordinatesBroker>(
+    //         CoordinatesBroker({CoordinatesSubset::Position(), CoordinatesSubset::Velocity()})
+    //     );
+
+    //     EXPECT_NO_THROW(State::fromStdVector(instant, coordinates, Frame::GCRF(), broker));
```
phc1990, line 43: I added 5 tests to check the static constructor but I had to comment out as the compiler didnt like them. 

I wasn't able to make anything out of the error message. So they are commented out for now. Any idea what I might be doing wrong?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
phc1990: I've opened a separate tix for this as it likely involves a discussion on its own : )

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp
kyle-cochran: > Could it be that Map.hpp assumes an underlying sorted map?

Perhaps, but we also [`OrderedMap`](https://github.com/open-space-collective/open-space-toolkit-core/blob/main/include/OpenSpaceToolkit/Core/Containers/OrderedMap.hpp), so I think this one would naturally be unsorted.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+    for (CoordinatesSubset coordinatesSubset : aCoordinatesSubsetArray)
+    {
+        this->addSubset(coordinatesSubset);
+    }
+}
+
+Index CoordinatesBroker::addSubset(const CoordinatesSubset& aCoordinatesSubset)
+{
+    auto search = this->coordinatesSubsetIndexMap_.find(aCoordinatesSubset);
```
Derollez, line 32:
```suggestion
    const auto search = this->coordinatesSubsetIndexMap_.find(aCoordinatesSubset);
```

I would make the name more explicit for `search`, like `coordinatesSubsetSearchIndex`

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.cpp
```diff
@@ -0,0 +1,77 @@
...
+    if (search != this->coordinatesSubsetIndexMap_.end())
+    {
+        return search->second;
+    }
+
+    Index preAdditionNextCoordinatesSubsetIndex = this->nextCoordinatesSubsetIndex_;
+    this->coordinatesSubsetIndexMap_.insert({aCoordinatesSubset, this->nextCoordinatesSubsetIndex_});
+    this->nextCoordinatesSubsetIndex_ += aCoordinatesSubset.getSize();
+
+    return preAdditionNextCoordinatesSubsetIndex;
```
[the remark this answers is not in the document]
Derollez: Also what's the value in returning `preAdditionNextCoordinatesSubsetIndex` and not `nextCoordinatesSubsetIndex_`? 🤔

test/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.test.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        EXPECT_NO_THROW(CoordinatesSubset("ID", 1));
+    }
+
+    {
+        EXPECT_NO_THROW(CoordinatesSubset("ID", 10));
+    }
+
+    {
+        EXPECT_NO_THROW(CoordinatesSubset("ID", -1));
```
Derollez, line 22: Is that an issue?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.test.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+        // Add subset
+        EXPECT_EQ(0, broker.addSubset(subset));
+
+        EXPECT_EQ(1, broker.getNumberOfCoordinates());
+        EXPECT_EQ(1, broker.getNumberOfSubsets());
+        EXPECT_TRUE(broker.hasSubset(subset));
+        EXPECT_TRUE(broker.hasSubset(duplicate));
+        EXPECT_EQ(0, broker.getSubsetIndex(subset));
+        EXPECT_EQ(0, broker.getSubsetIndex(duplicate));
```
Derollez, line 98: Nice, maybe also add a test with
```
CoordinatesSubset subset = CoordinatesSubset("S", 1);
CoordinatesSubset duplicate = CoordinatesSubset("S", 2);
```

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics.hpp
```diff
@@ -63,6 +69,12 @@ class Dynamics
 
     String getName() const;
 
+    /// @brief              Declares the coordinates involved in the contribution calculation
```
Derollez, line 72:
```suggestion
    /// @brief              Declare the coordinates involved in the contribution calculation
```

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/CentralBodyGravity.hpp
```diff
@@ -85,6 +88,12 @@ class CentralBodyGravity : public Dynamics
 
     Shared<const Celestial> getCelestial() const;
 
+    /// @brief              Declares the coordinates involved in the contribution calculation
```
Derollez, line 91:
```suggestion
    /// @brief              Declare the coordinates involved in the contribution calculation
```

Fix elsewhere

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/CentralBodyGravity.hpp
```diff
@@ -104,6 +113,8 @@ class CentralBodyGravity : public Dynamics
 
    private:
     Shared<const Celestial> celestialObjectSPtr_;
+    Integer positionIndex_ = Integer::Undefined();
```
Derollez, line 116: Why do we switch from using `Index` (implicitly ~ `size_t`) to using `Integer` in `Dynamics`?

include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/CentralBodyGravity.hpp
```diff
@@ -104,6 +113,8 @@ class CentralBodyGravity : public Dynamics
...
+    Integer velocityIndex_ = Integer::Undefined();
```
Derollez, line 117: I would make it even more explicit with `positionCoordinatesIndex_` and `velocityCoordinatesIndex_`, what do you think?

src/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.cpp
```diff
@@ -95,23 +95,31 @@ SatelliteSystem AtmosphericDrag::getSatelliteSystem() const
...
+    this->velocityIndex_ = coordinatesBroker->addSubset(CoordinatesSubset::Velocity());
+}
+
 void AtmosphericDrag::applyContribution(
     const Dynamics::StateVector& x, Dynamics::StateVector& dxdt, const Instant& anInstant
 ) const
 {
     (void)anInstant;
 
+    Vector3d positionCoordinates = Vector3d(x[positionIndex_], x[positionIndex_ + 1], x[positionIndex_ + 2]);
```
Derollez, line 110: Use size from the `coordinatesSubset` (i.e. 3 here) and slicing to make it more generic? That might be part of the next step, if so feel free to discard

src/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.cpp
```diff
@@ -95,23 +95,31 @@ SatelliteSystem AtmosphericDrag::getSatelliteSystem() const
...
+    Vector3d velocityCoordinates = Vector3d(x[velocityIndex_], x[velocityIndex_ + 1], x[velocityIndex_ + 2]);
```
Derollez, line 111: same

src/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/PositionDerivative.cpp
```diff
@@ -42,16 +42,22 @@ std::ostream& operator<<(std::ostream& anOutputStream, const PositionDerivative&
...
     const Dynamics::StateVector& x, Dynamics::StateVector& dxdt, const Instant& anInstant
 ) const
 {
     (void)anInstant;
 
     // Integrate position states
-    dxdt[0] = x[3];
-    dxdt[1] = x[4];
-    dxdt[2] = x[5];
+    dxdt[positionIndex_] += x[velocityIndex_];
```
Derollez, line 58:
```suggestion
    dxdt[positionIndex_] += x[velocityIndex_];
```
Do we really need `+=` here? 🤔

Derollez (commented):
Really nice design! I think we should rebase on main to clearly see the addition of this PR wrt step 1 (merged earlier). Also follow-up question: `coordinatesBrokerSPtr_` is now an attribute from `State` and `Propagator`, what is the flow making the link between the two here?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
```diff
@@ -3,55 +3,69 @@
...
+        const Instant& anInstant,
+        const VectorXd& allCoordinates,
+        const Shared<const Frame>& fromFrame,
+        const Shared<const CoordinatesBroker>& aCoordinatesBroker,
+        const Shared<const Frame>& toFrame
+    ) const = 0;
 
-    static CoordinatesSubset Mass();
+    // virtual VectorXd operator+(const CoordinatesSubset& aCoordinatesSbuset) const = 0;
+    // static CoordinatesSubset Undefined();
```
Derollez, line 64: TBR?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubsets/CartesianPosition.hpp
```diff
@@ -0,0 +1,63 @@
...
+        const Instant& anInstant,
+        const VectorXd& allCoordinates,
+        const Shared<const Frame>& fromFrame,
+        const Shared<const CoordinatesBroker>& aCoordinatesBroker,
+        const Shared<const Frame>& toFrame
+    ) const override;
+
+    static CartesianPosition OneDimensional();
+
+    static CartesianPosition TwoDimensional();
```
Derollez, line 50:
```suggestion
    static CartesianPosition OneDimensional();

    static CartesianPosition TwoDimensional();
```

Agreed with @vishwa2710's comment to focus on 3D for starter

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubsets/Invariant.hpp
```diff
@@ -0,0 +1,54 @@
...
+using ostk::core::types::String;
+
+using ostk::physics::time::Instant;
+
+using ostk::astro::trajectory::CoordinatesSubset;
+using ostk::astro::trajectory::CoordinatesBroker;
+
+/// @brief  Invariant (frame-indepdentent) coordinate subset.
+
+class Invariant : public CoordinatesSubset
```
Derollez, line 26:
```suggestion
class FrameInvariant : public CoordinatesSubset
```

Should we make it more explicit?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubset.hpp
phc1990: N/A

include/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubsets/Invariant.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace trajectory
+{
+namespace coordinatessubsets
```
kyle-cochran, line 14:
```suggestion
namespace coordinatessubset
```
To match the naming of the class. What do think?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.cpp
```diff
@@ -0,0 +1,156 @@
...
+Index CoordinatesBroker::getSubsetIndex(const Shared<const CoordinatesSubset>& aCoordinatesSubsetSPtr) const
+{
+    return this->getSubsetIndex(aCoordinatesSubsetSPtr->getId());
+}
+
+VectorXd CoordinatesBroker::extract(const VectorXd& allCoordinates, const CoordinatesSubset& aCoordinatesSubset) const
+{
+    Size size = aCoordinatesSubset.getSize();
+    VectorXd subsetCoordinates = VectorXd(size);
+    Index indexOffset = this->getSubsetIndex(aCoordinatesSubset.getId());
```
kyle-cochran, line 110:
```suggestion
    const Index indexOffset = this->getSubsetIndex(aCoordinatesSubset.getId());
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesBroker.cpp
```diff
@@ -0,0 +1,156 @@
...
+    return search != this->coordinateSubsetsIndexMap_.end();
+}
+
+Index CoordinatesBroker::getSubsetIndex(const String& anId) const
+{
+    auto search = this->coordinateSubsetsIndexMap_.find(anId);
+
+    if (search == this->coordinateSubsetsIndexMap_.end())
+    {
+        throw ostk::core::error::runtime::Wrong("Coordinates subset not found");
```
kyle-cochran, line 148:
```suggestion
        throw ostk::core::error::Runtime("Coordinates subset {} not found", anId);
```

pedantic, but I would use generic Runtime error here. Feel free to ignore this one, but I do think we should print the id of the subset either way.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubsets/CartesianPosition.cpp
```diff
@@ -0,0 +1,60 @@
...
+    const Instant& anInstant,
+    const VectorXd& allCoordinates,
+    const Shared<const Frame>& fromFrame,
+    const Shared<const CoordinatesBroker>& aCoordinatesBroker,
+    const Shared<const Frame>& toFrame
+) const
+{
+    VectorXd positionCoordinates = aCoordinatesBroker->extract(allCoordinates, *this);
+
+    Vector3d toFrameCoordinates =
```
kyle-cochran, line 44:
```suggestion
    const Vector3d toFrameCoordinates =
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/CoordinatesSubsets/CartesianVelocity.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    if (this->getSize() != 3)
+    {
+        throw ostk::core::error::runtime::ToBeImplemented("Frame Transformation");
+    }
+
+    const VectorXd positionCoordinates = aCoordinatesBroker->extract(allCoordinates, this->cartesianPositionSPtr_);
+    const VectorXd velocityCoordinates = aCoordinatesBroker->extract(allCoordinates, *this);
+
+    Vector3d toFrameCoordinates =
```
kyle-cochran, line 48:
```suggestion
    const Vector3d toFrameCoordinates =
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -106,10 +112,7 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
         Dynamics::GetDynamicalEquations(this->dynamics_, aState.getInstant())
     );
 
-    return {
-        anInstant,
-        Position::Meters({endStateVector[0], endStateVector[1], endStateVector[2]}, gcrfSPtr),
-        Velocity::MetersPerSecond({endStateVector[3], endStateVector[4], endStateVector[5]}, gcrfSPtr)};
+    return State::fromStdVector(anInstant, endStateVector, gcrfSPtr, coordinatesBrokerSPtr_);
```
kyle-cochran, line 115:
```suggestion
    return State::StdVector(anInstant, endStateVector, gcrfSPtr, coordinatesBrokerSPtr_);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -190,10 +193,7 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
     for (const Dynamics::StateVector& stateVector :
          (propagatedBackwardStateVectorArray + propagatedForwardStateVectorArray))
     {
-        State propagatedState = {
-            anInstantArray[k],
-            Position::Meters({stateVector[0], stateVector[1], stateVector[2]}, gcrfSPtr),
-            Velocity::MetersPerSecond({stateVector[3], stateVector[4], stateVector[5]}, gcrfSPtr)};
+        State propagatedState = State::fromStdVector(anInstantArray[k], stateVector, gcrfSPtr, coordinatesBrokerSPtr_);
```
kyle-cochran, line 196:
```suggestion
        State propagatedState = State::StdVector(anInstantArray[k], stateVector, gcrfSPtr, coordinatesBrokerSPtr_);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -195,10 +243,23 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+    Index i = 0;
 
-    return {this->instant_, position, velocity};
+    for (const Shared<const CoordinatesSubset> subset : this->coordinatesBrokerSPtr_->getSubsets())
+    {
+        const VectorXd subsetInFrame = subset->inFrame(
+            this->instant_, this->coordinates_, this->frameSPtr_, this->coordinatesBrokerSPtr_, aFrameSPtr
+        );
+
+        for (int j = 0; j < subsetInFrame.size(); j++)
```
kyle-cochran, line 255:
```suggestion
        for (Size j = 0; j < subsetInFrame.size(); j++)
```

kyle-cochran (commented):
hell of a job dude, this is awesome! Few comments on naming/const-ness

PR #135, opened 2023-07-17
fix: user permissions issues with debug jupyterlab and update jupyterlab image to python3.11 

apaletta3:


Derollez:
@apaletta3 We might need to bump OSTk Physics version to be in sync between python and cpp from the error we see here in python tests

vishwa2710:
I think the rebase should fix this 🤞

Derollez:
@apaletta3 Let's run the formatting for python and we should be good! :)

Makefile
```diff
@@ -299,14 +299,17 @@ debug-jupyter-notebook: build-release-image-jupyter ## Debug jupyter notebook
 	docker run \
 		-it \
 		--rm \
+		--user=root \
 		--publish="$(jupyter_notebook_port):8888" \
 		--volume="$(CURDIR)/bindings/python/docs:/home/jovyan/docs" \
 		--volume="$(CURDIR)/tutorials/python/notebooks:/home/jovyan/tutorials" \
```
kyle-cochran, line 305: This also gets chown'd, no? Need to clean this up as well?

Makefile
```diff
@@ -299,14 +299,17 @@ debug-jupyter-notebook: build-release-image-jupyter ## Debug jupyter notebook
...
 		--volume="$(CURDIR)/lib/libopen-space-toolkit-$(project_name).so.$(project_major_version):/opt/conda/lib/python$(jupyter_python_version)/site-packages/ostk/$(project_name)/libopen-space-toolkit-$(project_name).so.$(project_major_version):ro" \
```
kyle-cochran, line 306:
```suggestion
		--volume="$(CURDIR)/lib/libopen-space-toolkit-$(project_name).so.$(project_major_version):/opt/conda/lib/python$(jupyter_python_version)/site-packages/ostk/$(project_name)/libopen-space-toolkit-$(project_name).so.$(project_major_version):ro" \
		--group-add=$(shell id -g) \
```

Makefile
```diff
@@ -299,14 +299,17 @@ debug-jupyter-notebook: build-release-image-jupyter ## Debug jupyter notebook
...
 		--volume="$(CURDIR)/lib/$(jupyter_project_name_python_shared_object).so:/opt/conda/lib/python$(jupyter_python_version)/site-packages/ostk/$(project_name)/$(jupyter_project_name_python_shared_object).so:ro" \
 		--workdir="/home/jovyan" \
 		$(docker_release_image_jupyter_repository):$(docker_image_version) \
-		bash -c "start-notebook.sh --ServerApp.token=''"
+		bash -c "chown -R jovyan:users /home/jovyan ; start-notebook.sh --ServerApp.token=''"
+
+	bash -c "sudo chown -R $(shell id -u):$(shell id -g) $(CURDIR)/bindings/python/docs"
```
kyle-cochran, line 312:
```suggestion
		bash -c "chmod -R g+w /home/jovyan ; start-notebook.sh --ServerApp.token=''"
```


Not sure how to replicate your permissions situation, but you may be able to do something like this (+ the group add above) to avoid having a cleanup step.

Not tested, but the intent is to add the container user to the same group as the shell user, and just give group write permissions.

Makefile
apaletta3: I'm not sure what you mean by this? This only executes the docker run command as root so that I'm allowed to chown the home/jovyan folder in the first place right? I'm not sure it does anything else that is "lasting" and needs to be cleaned up? Lmk if I'm wrong

Makefile
apaletta3: @kyle-cochran This suggestion along with the --group-add doesn't seem to work for my purposes. I'm not sure exactly what you were intending with it.

Makefile
kyle-cochran: Oops I totally commented the wrong line 🤦 

I meant this second volume:
"$(CURDIR)/tutorials/python/notebooks:/home/jovyan/tutorials"

Since we `chown` the entire jovyan dir, this dir will be `chown`ed as well.

Makefile
kyle-cochran: (this is pedantic, if it doesn't work, feel free to resolve this comment)

apaletta3: Yes correct that's the issue at hand, but your suggestion didn't seem to give jovyan write access to the jupyter notebooks inside the jovyan dir

Makefile
apaletta3: Oh I see what you're saying. Yes that's correct, and I addressed it with another cleanup command after the docker image is stopped.

Makefile
kyle-cochran: Hmm.. the idea is to add group write access to the folder so that any user in the same group as the owner can write, then add jovyan to the same group as the owner. (rather than changing ownership)

I totally didn't test it though, so if something is fishy feel free to resolve this and 🚢

kyle-cochran (approved):

Makefile
apaletta3: Yeah I get the idea, its just that the current implementation you proposed doesn't seem to work. So if you want to try it real quick to see if you can get your implementation to work (run the make recipe with your changes and try to save an edited jupyter notebook from the browser) then go for it. Otherwise I think its probs good enough for now. @kyle-cochran

apaletta3: resolved over slack

PR #144, opened 2023-07-20
chore: suggestions for using a struct, draft

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics.hpp
```diff
@@ -51,6 +51,14 @@ class Dynamics
     typedef std::vector<double> StateVector;  // Container used to hold the state vector
     typedef std::function<void(const StateVector&, StateVector&, const double)> DynamicalEquationWrapper;
 
+    struct DynamicsInformation
```
vishwa2710, line 54: Name TBD

PR #146, opened 2023-07-21
feat: numerical solver add conditional stopping

vishwa2710:
TODO:
- Python bindings


vishwa2710:
This is closed in favor of #150

PR #149, opened 2023-07-21
feat: add EventCondition base class

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -0,0 +1,136 @@
...
+        case EventCondition::Criteria::NegativeCrossing:
+            return [](const Real& currentValue, const Real& previousValue) -> bool
+            {
+                return previousValue > 0.0 && currentValue < 0.0;
+            };
+
+        case EventCondition::Criteria::AnyCrossing:
+            return [](const Real& currentValue, const Real& previousValue) -> bool
+            {
+                return ((currentValue > 0.0) == (previousValue < 0.0));
```
kyle-cochran, line 116:
```suggestion
                return (currentValue > 0.0) == (previousValue < 0.0);
```

nit. idk what the formatter thinks about this though

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+class EventCondition
+{
+   public:
+    enum class Criteria
+    {
+        PositiveCrossing,
+        NegativeCrossing,
+        AnyCrossing,
+        PositiveOnly,
```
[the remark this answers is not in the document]
kyle-cochran: Looks like it does. What do you think about `PositiveValue` and `NegativeValue`? That feels a little more intuitive to me

kyle-cochran (commented):
This is so cool dude! Will give it another look tomorrow

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include <OpenSpaceToolkit/Core/Types/Real.hpp>
+#include <OpenSpaceToolkit/Core/Types/String.hpp>
+
+#include <OpenSpaceToolkit/Mathematics/Objects/Vector.hpp>
+
+namespace ostk
+{
+namespace astro
+{
```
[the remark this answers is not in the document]
Derollez: I think it makes sense yeah. But I think Vishwa added them at the root for now (similar to NumericalSolver) because will move it to math (hopefully soon)

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -0,0 +1,122 @@
...
+    virtual bool isSatisfied(const Real& currentValue, const Real& previousValue) const;
+
+    /// @brief              Evaluate the Event Condition
+    ///
+    /// @param              [in] aStateVector The current state vector
+    /// @param              [in] aTime The current time
+    ///
+    /// @return             Real number representing the evaluation result of the Event Condition
+
+    virtual Real evaluate(const VectorXd& aStateVector, const Real& aTime) const = 0;
```
Derollez, line 100: Neat, that way we can use VectorXd directly because NumericalSolver uses it too (with other PR), right?

Derollez (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    bool isSatisfied(const Real& currentValue, const Real& previousValue) const override
+    {
+        PYBIND11_OVERRIDE(bool, EventCondition, isSatisfied, currentValue, previousValue);
+    }
+
+    Real evaluate(const VectorXd& aStateVector, const Real& aTime) const override
+    {
+        PYBIND11_OVERRIDE_PURE(Real, EventCondition, evaluate, aStateVector, aTime);
+    }
```
Derollez, line 33: I guess there will be a usage when adding more later. Do we need it in here? (and the associated binding function)?

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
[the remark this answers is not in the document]
kyle-cochran: Oohhh yeah that makes sense to me. Vishwa, if you confirm feel free to close this comment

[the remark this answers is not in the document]
vishwa2710: Yeah precisely, there is a namespace for the different conditions but not for the parent object. Once we move it to Math we can figure what we want to do with it.

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
vishwa2710: Let me try!

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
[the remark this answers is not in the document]
vishwa2710: Yeah that's true. Every condition is defined by this mapping:
StateVector + Time -> a Real value which resolves to the targeted value

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
vishwa2710: This is required if we want to be able to define new EventConditions in Python

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
vishwa2710: Precisely!

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
vishwa2710: Works!

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+        case EventCondition::Criteria::Undefined:
+            return [](const Real& currentValue, const Real& previousValue) -> bool
+            {
+                (void)currentValue;
+                (void)previousValue;
+
+                throw ostk::core::error::runtime::Undefined("Comparator");
+
+                return false;
```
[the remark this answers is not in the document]
kyle-cochran: Well that's silly. Okay, resolving!

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    /// @brief              Convert criteria to string
+    ///
+    /// @param              [in] aCriteria An enum representing the criteria
+    ///
+    /// @return             String representing the given criteria
+
+    static String StringFromCriteria(const Criteria& aCriteria);
```
phc1990, line 108: Should this function belong to the `Criteria` class as a non-static func?

`String Criteria::toString()` or similar?

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+std::function<bool(const Real&, const Real&)> EventCondition::getComparator(const EventCondition::Criteria& aCriteria)
+{
+    switch (aCriteria)
+    {
+        case EventCondition::Criteria::StrictlyPositive:
+            return [](const Real& currentValue, const Real& previousValue) -> bool
+            {
+                (void)previousValue;
```
phc1990, line 90: Is this to avoid unusued warnings?

If so I have been using `[[maybe_unused]]`

vishwa2710: Yeah. It's also good practice to void unused parameters as the compiler can optimize it out.

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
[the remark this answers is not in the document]
phc1990: I know we've already talked about this @vishwa2710 so sorry to bring it back but the more I think we are boxing ourselves.

I really think this should be `evaluate (const State& aState)`

I know the first step is just to numerically propagate to this conditions but there are a bunch of problems that will arise from using `VectorXd` and `Real` as a time offset.

- Time based event conditions (which are many, almost all of them) will now need to carry over a private property storing initial time so that they can convert the offset to an instant.
- No frame-related information. One gets a `VectorXd` but no reference frame attached to this. So eventually this gets either passed or it needs to be 'assumed` by conditions (i.e. hardcoded)

Those are just details, but it overall defeats the purpose of having a `State` concept that already does all this for us. You can pass it a coordinate subset and it will give it to you, in the desired frame (since it has all the info inside).

I think the `EventCondition` interface is great, but this particular signature boxes us into really low level, when we don't really need to.

In the numerical propagator, just create a `State` off of your `VectorXd`, instant and integration reference frame, and pass it along to the EventCondition. Maybe just a tiny bit slowe but for sure a much better dev experience imo.

kyle-cochran (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
Derollez: Aaaaah, I see that makes sense! thanks for the clarification

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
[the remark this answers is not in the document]
vishwa2710: Okay I discussed this with @phc1990 offline and we agree that for now we should move forward as is. Once we land his State + CoordinateBroker MR we can figure out what's the best way to leverage this in the conditions.

The conditions will always need a VectorXd + Double/Real interface as the NumericalSolver is aimed to be agnostic of the Astro stuff once it moves to Math.

In the future if we want, we can subclass the solver in astro and make it state aware if required.

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
vishwa2710: Since `Criteria` is an `Enum` we can't provide such functionality to it and we have to rely on these static methods :'(

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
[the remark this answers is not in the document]
Derollez: > Once we land his State + CoordinateBroker MR we can figure out what's the best way to leverage this in the conditions.

That seems reasonable!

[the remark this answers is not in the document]
Derollez: > The conditions will always need a VectorXd + Double/Real interface as the NumericalSolver is aimed to be agnostic of the Astro stuff once it moves to Math.

👍

PR #150, opened 2023-07-21
feat: add event conditions to numerical solver

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -202,7 +208,30 @@ class NumericalSolver
         const SystemOfEquationsWrapper& aSystemOfEquations
     );
 
-    /// @brief              Perform numerical integration for a certain duration
+    /// @brief              Perform numerical integration from a start time till a condition is met, or end time is
+    /// reached
+    ///
+    /// @code
+    ///                     StateVector stateVector = numericalsolver.integrateTime(stateVector,
```
Derollez, line 215:
```suggestion
    ///                     StateVector stateVector = numericalSolver.integrateTime(stateVector,
```

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -202,7 +208,30 @@ class NumericalSolver
...
+    ///                     aStartTime, anEndTime, SystemofEquations, anEventCondition) ;
```
Derollez, line 216:
```suggestion
    ///                     aStartTime, anEndTime, SystemOfEquations, anEventCondition) ;
```

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -230,14 +259,35 @@ class NumericalSolver
...
-    /// @return             std::vector<double>
+    /// @return             Array<Solution>
 
     Array<Solution> integrateDuration(
         const StateVector& anInitialStateVector,
         const Array<Real>& aDurationArray,
         const SystemOfEquationsWrapper& aSystemOfEquations
     );
 
+    /// @brief              Perform numerical integration from a start time till a condition is met, or end time is
```
Derollez, line 270:
```suggestion
    /// @brief              Perform numerical integration from a start time till a condition is met, or duration is
```

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -230,14 +259,35 @@ class NumericalSolver
...
+    /// reached
+    ///
+    /// @code
+    ///                     StateVector stateVector = numericalsolver.integrateDuration(stateVector,
```
Derollez, line 274:
```suggestion
    ///                     StateVector stateVector = numericalSolver.integrateDuration(stateVector,
```

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -230,14 +259,35 @@ class NumericalSolver
...
+    ///                     aStartTime, aDurationInSeconds, SystemofEquations, anEventCondition) ;
```
Derollez, line 275:
```suggestion
    ///                     aStartTime, aDurationInSeconds, SystemOfEquations, anEventCondition) ;
```

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -230,14 +259,35 @@ class NumericalSolver
...
+    /// @endcode
+    /// @param              [in] anInitialStateVector An initial n-dimensional state vector to begin integrating at
+    /// @param              [in] aDurationInSeconds A duration to integrate for
+    /// @param              [in] aSystemOfEquations An std::function wrapper with a particular signature that
+    ///                     boost::odeint accepts to perform numerical integration
+    /// @param              [in] anEventCondition An event condition
+    /// @return             Solution
+
+    Solution integrateDuration(
+        const StateVector& anInitialStateVector,
```
Derollez, line 285: Should we augment this to take into account `aStartTime`?

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+NumericalSolver::Solution NumericalSolver::integrateDuration(
+    const StateVector& anInitialStateVector,
+    const Real& aDurationInSeconds,
+    const SystemOfEquationsWrapper& aSystemOfEquations,
+    const Shared<EventCondition>& anEventCondition
+)
+{
+    if (aDurationInSeconds.isZero())
+    {
+        return {anInitialStateVector, 0.0};
```
Derollez, line 404:
```suggestion
        return {anInitialStateVector, 0.0};
```

Should we use `anInitialStateVector` corresponding instant? for `solution.second` here? (see comment above)

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+    }
+
+    // TBI: Share information upstream on if the condition was satisfied
+    if (!conditionSatisfied)
+    {
+        stepper.calc_state(aDurationInSeconds, currentState);
+        return {currentState, aDurationInSeconds};
+    }
+
+    // Search for the exact time of the condition change
```
Derollez, line 472: So here we now the condition was met and we go further into determining the exact time with dichotomy because the integration scheme gives us an estimate based on stepper configuration, agreed?

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+        }
+
+        // TBI: Make tolerance a parameter
+        if (std::abs(midValue) < 1e-6)
+        {
+            break;
+        }
+    }
+
+    // TBI: Share information upstream on the number of iterations + success
```
Derollez, line 509: YES, that'd be amazing

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+        currentState = stepper.current_state();
+
+        currentValue =
+            anEventCondition->evaluate(Eigen::Map<const Eigen::VectorXd>(currentState.data(), stateSize), currentTime);
+
+        conditionSatisfied = anEventCondition->isSatisfied(currentValue, previousValue);
+
+        if (conditionSatisfied)
+        {
+            break;
```
Derollez, line 457: Random thought: that'd be good to assess how much overhead we had with this logic in the case of duration/time-based eventConditions, just to get a ball park

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+        }
+        else
+        {
+            // root lies between midTime and currentTime
+            // update previous -> mid
+            previousTime = midTime;
+            previousValue = midValue;
+        }
+
+        // TBI: Make tolerance a parameter
```
Derollez, line 502:
```suggestion
        // TBI: Make tolerance a parameter
```

Neat, should it be added to `EventCondition` derived classes, yeah?

Derollez: actually, maybe even base class?

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
Derollez: Do we want to store this into an additional object for logging? or do you literally mean adding to console here?

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -373,6 +392,131 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+        if (std::abs(midValue) < 1e-6)
+        {
+            break;
+        }
+    }
+
+    // TBI: Share information upstream on the number of iterations + success
+    // if (iterationCount == maxIterationCount)
```
Derollez, line 510: `maxIterationCount` should be flexible as well, 💯 

What do you think about adding a quick additional class for `RootFindingSolver > Dichotomy` and leveraging this in here? (same, we'd move this to `math` asap)

Derollez (commented):
Amazing, really awesome! Along the same lines as the `RootFinding` addition suggestion: Even if we might not want to add those now, I would still try to compartimentalize the content of `IntegrateDuration` core logic using private functions or so (for instance) just to make it clearer and more modular. Down the line additional classes for root finding and other processes are probably the way to go

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
Derollez: So I understand this is not a big deal for now for us but just thinking about somebody using `integrateDuration` directly with a time dependent set of equations (anyway, feel free to ignore, this is not the purpose of this PR)

vishwa2710 (commented):
Thanks for the excellent feedback!

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -221,7 +228,30 @@ class NumericalSolver
...
+    ///                     boost::odeint accepts to perform numerical integration
+    /// @param              [in] anEventCondition An event condition
+    /// @return             Solution
+
+    Solution integrateTime(
+        const StateVector& anInitialStateVector,
+        const Real& aStartTime,
+        const Real& anEndTime,
+        const SystemOfEquationsWrapper& aSystemOfEquations,
+        const Shared<EventCondition>& anEventCondition
```
kyle-cochran, line 251: What's the reason for passing this in as a Shared pointer instead of plain old passing by reference? Do we need it to be allocated on the heap for some reason? This seems like a user level thing and having to wrap your inputs in a shared pointer can be a little cumbersome IMO.

vishwa2710: Since `EventCondition` is an abstract class, we have to pass it via a shared pointer.
The flow is:
Base Abstract `EventCondition` class
|_> Child classes that implement the various conditions via the evaluate method
     |_> evaluate

Therefore for the polymorphism this has to be a shared pointer.

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -371,6 +391,133 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+
+    if (aDurationInSeconds.isZero())
+    {
+        return {anInitialStateVector, 0.0};
+    }
+
+    NumericalSolver::StateVector aStateVector = anInitialStateVector;
+
+    // Ensure that the time step is the correct sign
+    const double signedTimeStep = getSignedTimeStep(aDurationInSeconds);
```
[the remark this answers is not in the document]
vishwa2710: Odeint doesn't know how to play with Reals so it has to be a double.

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -371,6 +391,133 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+
+    // TBI: Adapt this to any dense stepper type
+    auto stepper = make_dense_output(absoluteTolerance_, relativeTolerance_, dense_stepper_type_5());
+
+    // initialize stepper
+    double currentTime = 0.0;
```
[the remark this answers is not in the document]
vishwa2710: Same as above.

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -371,6 +391,133 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+    stepper.initialize(aStateVector, currentTime, signedTimeStep);
+
+    // do first step
+    double previousTime;
```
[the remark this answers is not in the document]
vishwa2710: Samesies.

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
kyle-cochran: Maybe I'm misunderstanding the scope you're talking about, but I was a little skeptical because that seems like a decent restriction on polymorphism.
So I tried it out:
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/155

Changing it to a non-pointer type seems to work with at least one of the tests.

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
vishwa2710: Typical integration takes ~10 ms while this takes ~100 ms. Which is on par with what I would expect as we are doing a lot more operations to identify the exact crossing.

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -221,7 +226,30 @@ class NumericalSolver
         const SystemOfEquationsWrapper& aSystemOfEquations
     );
 
-    /// @brief              Perform numerical integration for a certain duration
+    /// @brief              Perform numerical integration from a start time until either a condition or the end time is
+    /// met reached
```
kyle-cochran, line 230:
```suggestion
    /// @brief              Perform numerical integration from a start time until either a condition or the end time is reached
    ///
```
typo?

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
```diff
@@ -249,14 +277,35 @@ class NumericalSolver
...
+    /// @return             Array<Solution>
 
     Array<Solution> integrateDuration(
         const StateVector& anInitialStateVector,
         const Array<Real>& aDurationArray,
         const SystemOfEquationsWrapper& aSystemOfEquations
     );
 
+    /// @brief              Perform numerical integration from a start time until either a condition or duration is met
+    /// reached
```
kyle-cochran, line 289:
```suggestion
    ///
```

kyle-cochran (approved):

include/OpenSpaceToolkit/Astrodynamics/NumericalSolver.hpp
vishwa2710: Kyle was right! We can pass this via reference as we are only using the parent methods so there are no issues with object slicing.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -133,18 +136,29 @@ class Propagator
...
+    /// @brief              Calculate the state subject to an Event Condition, given initial state and maximum end time
+    /// @code
+    ///                     State state = propagator.calculateStateAt(aState, anInstant, anEventCondition);
+    /// @endcode
+    /// @param              [in] aState An initial state
+    /// @param              [in] anInstant An instant
+    /// @param              [in] anEventCondition An event condition
+    /// @return             State
+
+    State calculateStateAt(const State& aState, const Instant& anInstant, const EventCondition& anEventCondition) const;
```
phc1990, line 156: So, eventually when we want to propagate to the first of many viable conditions, we can just group them under the OR condition and pass it here right?

Just a question, resolve if what I said is correct : )

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -133,18 +136,29 @@ class Propagator
...
-    ///                     State state = propagator.calculateStateAt(aState, anInstant) ;
+    ///                     State state = propagator.calculateStateAt(aState, anInstant);
     /// @endcode
     /// @param              [in] aState An initial state
     /// @param              [in] anInstant An instant
     /// @return             State
 
     State calculateStateAt(const State& aState, const Instant& anInstant) const;
 
+    /// @brief              Calculate the state subject to an Event Condition, given initial state and maximum end time
```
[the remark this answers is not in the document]
phc1990: sorry I guess it's the `anInstant`
nevermind, please resolve : )

src/OpenSpaceToolkit/Astrodynamics/NumericalSolver.cpp
```diff
@@ -371,6 +389,133 @@ Array<NumericalSolver::Solution> NumericalSolver::integrateDuration(
...
+    const auto checkTimeLimit = [&currentTime, &aDurationInSeconds]()
+    {
+        if (aDurationInSeconds > 0.0)
+        {
+            return currentTime < aDurationInSeconds;
+        }
+        return currentTime > aDurationInSeconds;
+    };
+
+    while (checkTimeLimit())
```
phc1990, line 444: Maybe we don't need this, we can maybe just check the condition, since there might always be one

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
vishwa2710: Yes that is indeed the case. In a separate MR I will add the Conjunctive/Disjunctive classes to enable this.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -111,6 +111,36 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
     };
 }
 
+State Propagator::calculateStateAt(
```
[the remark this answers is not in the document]
vishwa2710: I could.... but it's slower to do it via the DurationCondition than regular integration. In regular integration you just get the last step, and then do a step that is end_time - current_time.
In event condition finding, you do some kind of root solver. So it's not ideal.

PR #153, opened 2023-07-24
feat: add Semi Major Axis condition

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/EventCondition/SemiMajorAxisCondition.hpp
```diff
@@ -0,0 +1,83 @@
...
+using ostk::math::obj::VectorXd;
+
+using ostk::physics::units::Derived;
+using ostk::physics::units::Length;
+
+using ostk::astro::EventCondition;
+
+/// @brief                      A semi major axis based event condition
+
+class SemiMajorAxisCondition : public EventCondition
```
Derollez, line 35: How much of this can be "templated" if we wanted to add "Inclination". Can we group those by type at least? like `AngleCondition`, or `LengthCondition`? Would that make sense?

vishwa2710: We could potentially add another parent class for OrbitalElementCondition or something.
But I don't see any common code. I'm trying to be computationally efficient here so we're only calculating the element we need, than generating a COE object.

src/OpenSpaceToolkit/Astrodynamics/EventCondition/SemiMajorAxisCondition.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+    return Length::Meters(smaInMeters_);
+}
+
+Derived SemiMajorAxisCondition::getGravitationalParameter() const
+{
+    return {gravitationalParameterInSI_, GravitationalParameterSIUnit};
+}
+
+Real SemiMajorAxisCondition::evaluate(const VectorXd& aStateVector, const Real& aTime) const
```
phc1990, line 43: This is what I meant about having to make some assumptions when operating one level below `State`

We are making 2 implicit assumptions here. Is going to work for us 99.9% of the time, but still I think if we can avoid them, it's a better dev experience. For instance, assumptions here are:

1) VectorXd coordinates are expressed w.r.t. an inertial frame (it is likely that this is always the case)
2) VectorXd coordinates are expressed w.r.t. a frame centered at the central body (it is likely that this is always the case)

Still, to me the robust way to go about it:
1. Provide not only the standard gravitational parameter in the constructor but also the inertial frame of choice centered at the desired central body (ideally we could only pass the `Celestial` body but I dont think it offers a `getInertialFrame`.
2. In the `evaluate` logic, you receive a `State`, extract pos/vel from it in the frame specified in the constructor, and then proceed to compute the semi-major axis.

As I said, not that it will fail, but I see this as a risk.
Sorry being a pain in the ass again.

vishwa2710: Not at all, this is great feedback. I think we have to find the balance between using high level objects and low level data structures. As we discussed offline, we are going to reassess this once we land your MR on the Coordinate Broker.

include/OpenSpaceToolkit/Astrodynamics/EventCondition/SemiMajorAxisCondition.hpp
kyle-cochran: Sorry for being late to this convo 🙏. My vote here is to go with the thing that makes us have to change our API less down the road. 
This class make sense to me. :) 

Maybe down the road it makes sense to expose a factory function:
`EventCondition:ConditionFromName("Semi Major Axis")` 

Just to make these a bit easier to work with

vishwaLoft:
Re-opening another MR for this.

PR #159, opened 2023-07-26
feat: suggestions add coordinates broker and subset

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesBroker.hpp
```diff
@@ -132,16 +132,6 @@ class CoordinatesBroker
     VectorXd extract(const VectorXd& allCoordinates, const Shared<const CoordinatesSubset>& aCoordinatesSubsetSPtr)
         const;
 
-    /// @brief              Create a shared pointer wiht the given coordinates subsets.
-    ///
-    /// @param              [in] aCoordinatesSubsetsArray the coordinates subsets to consider
-    ///
-    /// @return             A new instance
-
-    static Shared<const CoordinatesBroker> FromSubsets(
```
vishwa2710, line 141: This method is redundant as it can be replicated with a single line:
`std::make_shared<CoordinateBroker>(aCoordinateSubsetsArray);`

PR #160, opened 2023-07-26
feat: add CalculateStateAt with conditions for propagator

vishwa2710:


bindings/python/test/trajectory/test_propagator.py
```diff
@@ -12,6 +12,7 @@
 from ostk.physics.units import Mass
 from ostk.physics.time import Instant
 from ostk.physics.time import DateTime
+from ostk.physics.time import Duration
```
kyle-cochran, line 15:
```suggestion
```

I might be blind, but was this used in this file?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -41,6 +42,7 @@ namespace trajectory
 using ostk::core::ctnr::Array;
 using ostk::core::types::Integer;
 using ostk::core::types::Real;
+using ostk::core::types::Shared;
```
kyle-cochran, line 45:
```suggestion
```

idem. I might be blind, but don't see a Shared pointer added

kyle-cochran (approved):
The nittest of picks

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
vishwa2710: This is used by the dynamics class in the constructor that already exists.
We get away with it because we're importing a file that has the `using` defined but ideally we should be explicit per file.

PR #164, opened 2023-07-27
feat: add Root Solver

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/RootSolver.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+namespace ostk
+{
+namespace astro
+{
+
+using ostk::core::types::Size;
+using ostk::core::types::Real;
+
+/// @brief                      A Root Solver is a tool for finding the roots of equations.
```
Derollez, line 19:
```suggestion
/// @brief                      A Root Solver is a method for finding the roots (a.k.a. zeros) of an equation.
```

include/OpenSpaceToolkit/Astrodynamics/RootSolver.hpp
```diff
@@ -0,0 +1,113 @@
...
+class RootSolver
+{
+   public:
+    struct Solution
```
Derollez, line 23: Nice, similarly, I think we can have a RootSolver Settings (or Config) Struct (not necessarily now but later if you had more methods) to store all the settings and parameters for the rootSolvers. They usually have a bunch in common (maxIters, tols, etc)

src/OpenSpaceToolkit/Astrodynamics/RootSolver.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void RootSolver::print(std::ostream& anOutputStream, bool displayDecorator) const
+{
+    displayDecorator ? ostk::core::utils::Print::Header(anOutputStream, "Root Solver") : void();
+
+    ostk::core::utils::Print::Line(anOutputStream) << "Maximum Iterations Count:" << maximumIterationsCount_;
+    ostk::core::utils::Print::Line(anOutputStream) << "Number of Digits:" << numberOfDigits_;
+
+    displayDecorator ? ostk::core::utils::Print::Footer(anOutputStream) : void();
+}
```
Derollez, line 38: Move after getter implementation

Derollez: and after functions

include/OpenSpaceToolkit/Astrodynamics/RootSolver.hpp
```diff
@@ -0,0 +1,113 @@
...
+    /// @brief              Solve for root given a function, an initial guess and function direction
+    ///
+    /// @param              [in] aFunction A function
+    /// @param              [in] anInitialGuess An initial guess
+    /// @param              [in] isRising A boolean indicating whether the function is rising
+    ///
+    /// @return             The solution
+
+    Solution solve(
+        const std::function<double(const double&)>& aFunction, const double& anInitialGuess, const bool& isRising
```
Derollez, line 80: Just out of curiosity (let's not do that now), boost has some methods for multi-dimensional functions? see some good applications for multi-variate polynomials (laaaater)

Derollez (approved):
LGTM, thanks for making this a legit class !

include/OpenSpaceToolkit/Astrodynamics/RootSolver.hpp
vishwa2710: Interestingggg. I will do some research.

include/OpenSpaceToolkit/Astrodynamics/RootSolver.hpp
vishwa2710: Yeah for sure. I'm not sure how this class will grow so we can definitely do that in the future.

PR #171, opened 2023-08-01
feat: add logical connective event conditions

vishwa2710:
Note: This changes how we check for event conditions in the NumericalSolver.
Previously:
- calculate the value at the current step, use the previous value + current to check condition
- previousValue = currentValue
This was more efficient as it only calculates the condition once per step.

However, this is not (easily) possible with a _set_ of conditions. Therefore I am doing the (less efficient) check of isSatisfied at each step.


include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -90,6 +90,22 @@ class EventCondition
 
     virtual bool isSatisfied(const Real& currentValue, const Real& previousValue) const;
 
+    /// @brief              Check if the Event Condition is satisfied based on current value and previous value
```
kyle-cochran, line 93:
```suggestion
    /// @brief              Check if the Event Condition is satisfied based on current state and previous state.
```
Maybe add context on why time is needed here?

include/OpenSpaceToolkit/Astrodynamics/EventCondition/Conjunctive.hpp
```diff
@@ -0,0 +1,71 @@
+/// Apache License 2.0
```
kyle-cochran, line 1: I totally agree with being explicit and using the formal names `Conjunctive` and `Disjunctive`. However, in the future I think it would also be good to define an alias:
`using And = Conjunctive`
`using Or = Disjunctive`

Just to ease the burden for newcomers.

kyle-cochran: Or if we could work this into defining operators on the Event Condition class that would be 🤯

src/OpenSpaceToolkit/Astrodynamics/EventCondition/Conjunctive.cpp
```diff
@@ -0,0 +1,38 @@
...
+Conjunctive::~Conjunctive() {}
+
+bool Conjunctive::isSatisfied(
+    const VectorXd& currentStateVector,
+    const Real& currentTime,
+    const VectorXd& previousStateVector,
+    const Real& previousTime
+) const
+{
+    for (const Shared<EventCondition>& eventCondition : eventConditions_)
```
kyle-cochran, line 25: Not that important, but `std::all_of` could be used here and has closer semantics to the AND nature of this class. Could be slightly better readability:
```
return std::all_of(
  eventConditions_.begin(),
  eventConditions_.end(),
  [&](EventCondition& condition){return condition->isSatisfied(currentStateVector, currentTime, previousStateVector, previousTime)}
);
```
Totally a style thing, so your call. Feel free to close this comment without responding.

src/OpenSpaceToolkit/Astrodynamics/EventCondition/Disjunctive.cpp
```diff
@@ -0,0 +1,38 @@
...
+bool Disjunctive::isSatisfied(
+    const VectorXd& currentStateVector,
+    const Real& currentTime,
+    const VectorXd& previousStateVector,
+    const Real& previousTime
+) const
+{
+    for (const Shared<EventCondition>& eventCondition : eventConditions_)
+    {
+        if (eventCondition->isSatisfied(currentStateVector, currentTime, previousStateVector, previousTime))
```
kyle-cochran, line 27: idem here with `std::any_of`

test/OpenSpaceToolkit/Astrodynamics/EventCondition/Conjunctive.test.cpp
```diff
@@ -0,0 +1,173 @@
...
+        currentState << -0.5, 0.0;
+
+        VectorXd previousState(2);
+        previousState << -1.0, 3.0;
+
+        EXPECT_FALSE(conjuntionCondition_.isSatisfied(currentState, 0.0, previousState, 1.0));
+    }
+}
+
+// TEST_F(OpenSpaceToolkit_Astrodynamics_EventCondition_Conjunctive, isSatisfiedFromCache)
```
kyle-cochran, line 124: Should this be removed?

test/OpenSpaceToolkit/Astrodynamics/EventCondition/Disjunctive.test.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+    {
+        EXPECT_NO_THROW(Disjunctive Disjunctive({
+            std::make_shared<FirstCondition>(firstCondition_),
+            std::make_shared<SecondCondition>(secondCondition_),
+        }));
+    }
+}
+
+// TEST_F(OpenSpaceToolkit_Astrodynamics_EventCondition_Disjunctive, isSatisfied)
```
kyle-cochran, line 77: Seems like this one should be uncommented.

test/OpenSpaceToolkit/Astrodynamics/EventCondition/Disjunctive.test.cpp
```diff
@@ -0,0 +1,173 @@
...
+//         currentState << -0.5, 0.0;
+
+//         VectorXd previousState(2);
+//         previousState << -1.0, 3.0;
+
+//         EXPECT_TRUE(disjunctionCondition_.isSatisfied(currentState, 0.0, previousState, 1.0));
+//     }
+// }
+
+// TEST_F(OpenSpaceToolkit_Astrodynamics_EventCondition_Disjunctive, isSatisfiedFromCache)
```
kyle-cochran, line 124: Should this be removed?

include/OpenSpaceToolkit/Astrodynamics/EventCondition/Conjunctive.hpp
vishwa2710: Oh wow that's an interesting idea. Let me noodle with that.

vishwa2710: Okay I couldn't get this to work with operators :(
Too much shared pointer cloning weirdness. I agree that having a better interface here would be nice...

kyle-cochran: Ah well. Not that big of a deal.

include/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.hpp
```diff
@@ -0,0 +1,84 @@
...
+    virtual bool isSatisfied(const Real& currentValue, const Real& previousValue) const override;
+
+    /// @brief              Throw an error as the condition cannot be evaluated based on the state vector and time
+    ///
+    /// @param              [in] aStateVector The current state vector
+    /// @param              [in] aTime The current time
+    ///
+    /// @return             Real number representing the evaluation result of the Logical Connective Event Condition.
+
+    virtual Real evaluate(const VectorXd& aStateVector, const Real& aTime) const override;
```
kyle-cochran, line 74:
```suggestion
    virtual Real evaluate(const VectorXd& aStateVector, const Real& aTime) const override = 0;
```

Ditto

src/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+LogicalConnective::LogicalConnective(const String& aName, const Array<Shared<EventCondition>>& eventConditions)
+    : EventCondition(aName, EventCondition::Criteria::Undefined),
+      eventConditions_(eventConditions)
+{
+}
+
+LogicalConnective::~LogicalConnective() {}
+
+Array<Shared<EventCondition>> LogicalConnective::getEventConditions() const
```
kyle-cochran, line 19: If this is called a lot, might make sense to add an `accessEventConditions` and use that instead.

kyle-cochran (approved):

include/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.hpp
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    /// @code
+    ///                     LogicalConnective logicalCondition = {aName, eventConditions};
+    /// @endcode
+    ///
+    /// @param              [in] aName A string representing the name of the Logical Connective Event Condition
+    /// @param              [in] eventConditions An Array of shared pointers to EventCondition instances,
+    ///                                      representing the individual event conditions to be evaluated.
+
+    LogicalConnective(const String& aName, const Array<Shared<EventCondition>>& eventConditions);
```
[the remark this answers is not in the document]
vishwa2710: Hmmmm not sure why we would want to reference the pointer? The reference of the pointer should be the same memory footprint as the pointer yeah?

include/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+    /// @brief              Throw an error as the condition cannot be evaluated based on the current state vector and
+    /// time
+    ///
+    /// @param              [in] currentValue The current value
+    /// @param              [in] previousValue The previous value
+    ///
+    /// @return             Boolean value indicating if the condition is met.
+
+    virtual bool isSatisfied(const Real& currentValue, const Real& previousValue) const override;
```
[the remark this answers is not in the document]
vishwa2710: Ah. Yeah. So this method sort of stops meaning anything with respect to a combination of event conditions. So I want to throw an error in the case that it's ever called.
And I'm doing that here so that I don't have to implement it twice for the children (Disjunctive, Conjunctive)

include/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.hpp
vishwa2710: Idem

src/OpenSpaceToolkit/Astrodynamics/EventCondition/LogicalConnective.cpp
vishwa2710: Fair point. Currently it's not used at all, I feel like it will mostly just be used in python. But I'll keep that in mind and add it in the future if we required it.

PR #172, opened 2023-08-01
refactor: parametrize numerical solver tests to reduce boiler plate

vishwa2710:


test/OpenSpaceToolkit/Astrodynamics/NumericalSolver.test.cpp
```diff
@@ -98,6 +130,22 @@ class OpenSpaceToolkit_Astrodynamics_NumericalSolver : public ::testing::Test
     }
 };
 
+class OpenSpaceToolkit_Astrodynamics_NumericalSolver_Parametrized
+    : public OpenSpaceToolkit_Astrodynamics_NumericalSolver,
+      public ::testing::WithParamInterface<Tuple<NumericalSolver::StepperType>>
+{
+};
+
+INSTANTIATE_TEST_SUITE_P(
```
kyle-cochran, line 139: This is super cool. Didn't know it existed. 

Just to make sure I'm reading stuff right, it looks like we're now testing all of the numerical solvers against the same tolerance. Is that correct?

vishwa2710: Yes that's right. By reducing the step size on the RK54 we can compare everything with the same tolerance. Which is a small change from before. Great 👁️

kyle-cochran: Looks like the tolerance is also looser in some places. I think this is fine for this class, since they are just wrappers over the boost numerical solver classes. 

If we ever decide that we do need to test higher tolerances in some places, is that something we can include in the Google Test parameterization?

vishwa2710: Yeah for sure, we can just add it as an additional parameter in the parametrized template. Tagging you where.

test/OpenSpaceToolkit/Astrodynamics/NumericalSolver.test.cpp
```diff
@@ -98,6 +130,22 @@ class OpenSpaceToolkit_Astrodynamics_NumericalSolver : public ::testing::Test
     }
 };
 
+class OpenSpaceToolkit_Astrodynamics_NumericalSolver_Parametrized
+    : public OpenSpaceToolkit_Astrodynamics_NumericalSolver,
+      public ::testing::WithParamInterface<Tuple<NumericalSolver::StepperType>>
```
vishwa2710, line 135:
```suggestion
      public ::testing::WithParamInterface<Tuple<NumericalSolver::StepperType, Real>>
```
@kyle-cochran over here

kyle-cochran (approved):

test/OpenSpaceToolkit/Astrodynamics/NumericalSolver.test.cpp
kyle-cochran: Sick. This is cool w me then

PR #177, opened 2023-08-03
chore: improve development helpers

lucas-bremond:
This MR proposes a new/improved approach to development helpers.

Instead of mounting the helper scripts inside the `build/` folder, we now copy them into the development image, and add `ostk-` command to make them easier to invoke (for example, typing `ostk` + tab provides shell auto-completion).

On top, it makes running linting/checking from within the dev container, or from the outside, the same.

Finally, `make dev` is added as an alias to `make start-development`.


lucas-bremond:
Now using helpers from the base image!

vishwa2710 (approved):
Looks great!

PR #181, opened 2023-08-04
feat: add Temporal Condition Solver

lucas-bremond:
This PR adds a Temporal Condition Solver class, providing a generic algorithm to find the set of sub-intervals over which a set of conditions are met.

The Access Generator being a special case, it has been refactored to rely on this new solver under the hood.

In addition, this solver can be used to handle more complex cases, for example by returning time intervals over a certain zone while not being in eclipse, etc.

Here's an example of how this translates into Python (assuming `in_access` and `in_eclipse` are provided functions):

```py
temporal_condition_solver = TemporalConditionSolver(
    time_step=Duration.seconds(30.0),
    tolerance=Duration.milliseconds(1.0),
    maximum_iteration_count=500,
)

intervals: list[Interval] = temporal_condition_solver.solve(
    conditions=[
        lambda instant: in_access(instant),
        lambda instant: not in_eclipse(instant),
    ],
    interval=interval,
)
```

Note that an `in_access` function can now be obtained from the Access Generator class:

```py
access_generator = Generator(
    environment=Environment.default(),
)

in_access = access_generator.get_condition_function(
    from_trajectory=trajectory,
    to_trajectory=trajectory,
)
```

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Solvers/TemporalConditionSolver.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Solvers/TemporalConditionSolver.cpp
@@ -1,0 +1,192 @@
+/// Apache License 2.0
+
+#include <boost/math/tools/roots.hpp>
+
+#include <OpenSpaceToolkit/Core/Error.hpp>
+#include <OpenSpaceToolkit/Core/Utilities.hpp>
+
+#include <OpenSpaceToolkit/Astrodynamics/RootSolver.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Solvers/TemporalConditionSolver.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace solvers
+{
+
+TemporalConditionSolver::TemporalConditionSolver(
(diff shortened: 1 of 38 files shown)


src/OpenSpaceToolkit/Astrodynamics/Solvers/TemporalConditionSolver.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // Add interval if condition is true on the last iteration
+    if (inCondition)
+    {
+        intervals.add(Interval::Closed(conditionStartCache, instants.accessLast()));
+    }
+
+    return intervals;
+}
+
+// Instant TemporalConditionSolver::findSwitchingInstant(
```
lucas-bremond, line 101: Remove unused versions

vishwa2710 (approved):
Overall looks great man! Just some nits.

I think in the future we might want to refactor the generator a bit so that we're doing multiple calls for `getStateAt` as that can be expensive if this was a propagated Model. Definitely something not for this MR. I really like how slick the interface is now.

Another thing we can do is create an Access based EventCondition, and then generalize the generator to work with any condition.
Either way, great stuff!

PR #182, opened 2023-08-07
feat: add CesiumPy Viewer to Python utilities

lucas-bremond:
This PR adds a 3D viewer based on [CesiumPy](https://github.com/lucas-bremond/cesiumpy) to the Python utilities, to simplify displaying Profiles in Python.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -471,6 +471,7 @@ test-unit-python-standalone: ## Run Python unit tests (standalone)
 		/bin/bash -c "cmake -DBUILD_PYTHON_BINDINGS=ON -DBUILD_UNIT_TESTS=OFF .. \
 		&& $(MAKE) -j 4 && python3.11 -m pip install --root-user-action=ignore bindings/python/dist/*311*.whl \
 		&& python3.11 -m pip install plotly pandas \
+		&& python3.11 -m pip install git+https://github.com/lucas-bremond/cesiumpy.git#egg=cesiumpy \
 		&& cd /usr/local/lib/python3.11/site-packages/ostk/$(project_name)/ \
 		&& python3.11 -m pytest -sv ."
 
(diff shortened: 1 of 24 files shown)


Makefile
```diff
@@ -471,6 +471,7 @@ test-unit-python-standalone: ## Run Python unit tests (standalone)
 		/bin/bash -c "cmake -DBUILD_PYTHON_BINDINGS=ON -DBUILD_UNIT_TESTS=OFF .. \
 		&& $(MAKE) -j 4 && python3.11 -m pip install --root-user-action=ignore bindings/python/dist/*311*.whl \
 		&& python3.11 -m pip install plotly pandas \
+		&& python3.11 -m pip install git+https://github.com/lucas-bremond/cesiumpy.git#egg=cesiumpy \
```
vishwaLoft, line 474: Can do in a separate MR but should we just make a `test-requirements.in`?

bindings/python/requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Apache License 2.0
 
-open-space-toolkit-physics~=1.1.1
+open-space-toolkit-physics~=1.1
```
vishwaLoft, line 3: Is there a reason you're pinning to the major version? I think we had decided that we will pin to the minor version no?

vishwaLoft (approved):
This is freaking sick! Just some nits.

bindings/python/requirements.txt
kyle-cochran: +1 I believe this will allow Python to automatically update minor versions, which is different than how cpp dependencies get propagated. This caused some syncing issues between cpp/Python in the past.

kyle-cochran (approved):

Makefile
lucas-bremond: Yeah that'd be great!

bindings/python/requirements.txt
lucas-bremond: Reverted!

PR #185, opened 2023-08-08
build: fix format-python in makefile

vishwa2710:
This fixes the issue @kyle-cochran found with running `make format-python` locally.
I checked the /etc/group in the docker container and it doesn't contain our groups.

If you exec into the container it complains about this:
groups: cannot find name for group ID 1050


kyle-cochran (approved):
🙏  LGTM. As discussed, the formatter never adds file, so should be no danger of bad file permissions being created.

PR #187, opened 2023-08-14
fix: COECondition static construction

vishwa2710:
Because of how the member variables were getting initialized, this would produce a segfault as it was getting captured in the lambda before initialization.

### DIFF
--- a/bindings/python/test/event_condition/test_coe_condition.py
+++ b/bindings/python/test/event_condition/test_coe_condition.py
@@ -79,9 +79,12 @@ class TestCOECondition:
         ),
     )
     def test_static_constructors(
-        self, static_constructor, target, criteria, gravitational_parameter
+        self, static_constructor, target, criteria, gravitational_parameter, state_vector
     ):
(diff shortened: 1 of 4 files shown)


bindings/python/test/event_condition/test_coe_condition.py
```diff
@@ -79,9 +79,12 @@ def test_getters(self, condition, element, target, gravitational_parameter):
...
     )
     def test_static_constructors(
-        self, static_constructor, target, criteria, gravitational_parameter
+        self, static_constructor, target, criteria, gravitational_parameter, state_vector
     ):
-        assert static_constructor(criteria, target, gravitational_parameter) is not None
+        condition = static_constructor(criteria, target, gravitational_parameter)
+        assert condition is not None
+        
+        condition.evaluate(state_vector, 0.0) is not None
```
kyle-cochran, line 87:
```suggestion
        assert condition.evaluate(state_vector, 0.0) is not None
```
yeah?

kyle-cochran (approved):

PR #189, opened 2023-08-14
feat: add access methods to propagator

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -70,17 +70,33 @@ std::ostream& operator<<(std::ostream& anOutputStream, const Propagator& aPropag
...
+    {
+        throw ostk::core::error::runtime::Undefined("Propagator");
+    }
+
+    return this->numericalSolver_;
+}
+
+Size Propagator::getNumberOfCoordinates() const
+{
+    return this->coordinatesBrokerSPtr_->getNumberOfCoordinates();
```
kyle-cochran, line 99:
```suggestion
    return this->accessCoordinatesBroker()->getNumberOfCoordinates();
```

so it uses the `isDefined` check

kyle-cochran (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
vishwa2710: Ah nice yeah. That's a good call.

PR #192, opened 2023-08-17
draft: magic_enum + macros = joy

vishwa2710:
Just a spike trying out the idea. If we like this pattern I can move this over to Core. Saves some boiler plate. And it's kind of neat :)

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Utilities.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Utilities.hpp
@@ -1,0 +1,30 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Utilities__
+#define __OpenSpaceToolkit_Astrodynamics_Utilities__
+
+#include <magic_enum.hpp>
+#include <stdexcept>
+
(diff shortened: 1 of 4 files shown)


lucas-bremond (approved):

include/OpenSpaceToolkit/Astrodynamics/Utilities.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Utilities__
+#define __OpenSpaceToolkit_Astrodynamics_Utilities__
+
+#include <magic_enum.hpp>
+#include <stdexcept>
+
+#include <OpenSpaceToolkit/Core/Error.hpp>
+
+#define DEFINE_ENUM_CONVERSIONS(EnumClass)                                 \
```
kyle-cochran, line 11:
```suggestion
// Automatically define class <-> string conversions for enums
#define DEFINE_ENUM_CONVERSIONS(EnumClass)                                 \
```

vishwa2710:
Going to land this in OSTk core so that it can be leveraged by all repos

PR #193, opened 2023-08-21
feat: add compute virtual method to Event Condition

vishwa2710:
- Add a compute method that returns the computed value
- Add a "target" member variable to EventCondition
- evaluate is now a non-virtual parent method in EventCondition that calls compute and subtracts the target

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
@@ -2,11 +2,13 @@
 
 #include <OpenSpaceToolkit/Astrodynamics/EventCondition.hpp>
 
+#include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/BooleanEventCondition.cpp>
 #include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/COECondition.cpp>
 #include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/Conjunctive.cpp>
 #include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/Disjunctive.cpp>
 #include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/DurationCondition.cpp>
 #include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/LogicalConnective.cpp>
+#include <OpenSpaceToolkitAstrodynamicsPy/EventCondition/RealEventCondition.cpp>
 
 using namespace pybind11;
 
@@ -26,11 +28,6 @@ class PyEventCondition : public EventCondition
 
     // Trampoline (need one for each virtual function)
 
-    bool isSatisfied(const Real& currentValue, const Real& previousValue) const override
-    {
-        PYBIND11_OVERRIDE(bool, EventCondition, isSatisfied, currentValue, previousValue);
-    }
-
     bool isSatisfied(
         const VectorXd& currentStateVector,
         const Real& currentTime,
@@ -38,14 +35,16 @@ class PyEventCondition : public EventCondition
         const Real& previousTime
     ) const override
     {
-        PYBIND11_OVERRIDE(
-            bool, EventCondition, isSatisfied, currentStateVector, currentTime, previousStateVector, previousTime
+        PYBIND11_OVERRIDE_PURE_NAME(
+            bool,
+            EventCondition,
+            "is_satisfied",
+            isSatisfied,
+            currentStateVector,
+            currentTime,
+            previousStateVector,
+            previousTime
         );
-    }
-
-    Real compute(const VectorXd& aStateVector, const Real& aTime) const override
-    {
-        PYBIND11_OVERRIDE_PURE(Real, EventCondition, compute, aStateVector, aTime);
     }
 };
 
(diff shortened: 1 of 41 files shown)


src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -36,12 +37,23 @@ EventCondition::Criteria EventCondition::getCriteria() const
...
 }
 
+Real EventCondition::getTarget() const
+{
+    return target_;
+}
+
+Real EventCondition::evaluate(const VectorXd& aStateVector, const Real& aTime) const
+{
+    return compute(aStateVector, aTime) - target_;
```
Derollez, line 47: Does the order/sign of the substraction matter for any use case? 🤔

src/OpenSpaceToolkit/Astrodynamics/EventCondition/DurationCondition.cpp
```diff
@@ -10,22 +10,21 @@ namespace eventcondition
...
-    return Duration::Seconds(durationInSeconds_);
+    return Duration::Seconds(getTarget());
 }
 
-Real DurationCondition::evaluate(const VectorXd& aStateVector, const Real& aTime) const
+Real DurationCondition::compute(const VectorXd& aStateVector, const Real& aTime) const
 {
     (void)aStateVector;
-    return aTime - durationInSeconds_;
+    return aTime;
```
Derollez, line 27: Alright, should all be backward compatible as `evaluate` remains the method computing the difference between target and value, confirmed?

Derollez:
So `COECondition` already had `target`. Here we are moving that attribute to the base class `EventCondition`, I am assuming this is making our lives easier in general to have `target` defined at `EventCondition` level later on when manipulating them but jut curious if you have an example for this 👀

src/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.cpp
```diff
@@ -53,39 +47,14 @@ COE::Element COECondition::getElement() const
...
-        case COE::Element::TrueAnomaly:
-            return "True anomaly";
-        case COE::Element::MeanAnomaly:
-            return "Mean anomaly";
-        case COE::Element::EccentricAnomaly:
-            return "Eccentric anomaly";
-    }
-
-    throw ostk::core::error::runtime::Wrong("Element");
+    return evaluator_(positionVector, velocityVector, this->gravitationalParameter_);
```
Derollez, line 57: Should we rename the `evaluator_` and the `GetEvaluator` items to match the difference we are now making between `compute` and `evaluate` methods?

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
Derollez: Alright, LGTM. Checked different target values (positive, negative) at `isSatisfied` level

src/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.cpp
vishwa2710: Yes this is a good idea!

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
```diff
@@ -58,11 +58,22 @@ inline void OpenSpaceToolkitAstrodynamicsPy_EventCondition(pybind11::module& aMo
...
+                init<const String&, const EventCondition::Criteria&, const Real&>(),
+                arg("name"),
+                arg("criteria"),
+                arg("target")
+            )
 
             .def("__str__", &(shiftToString<EventCondition>))
             .def("__repr__", &(shiftToString<EventCondition>))
 
+            .def("get_name", &EventCondition::getName)
```
kyle-cochran, line 71: Not 100% about this MR but would be nice to have some kind of glossary (maybe as part of the doc strings) that say what all of these words mean. 

It's not very obvious what `target` or `criteria` means or what the difference is between `compute` and `evaluate`. Maybe adding a class-level docstring to Event conditions could help?

vishwa2710: Discussed this with Kyle and we agreed that the doxygen comments are sufficient for now...

phc1990 (approved):
Looks great!

Left just a couple of nit comments.

I see benchmark is complaining, lmk if you'd like me to take a look at it : )

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -41,7 +41,8 @@ class EventCondition
     /// @endcode
     ///
     /// @param                  [in] aName A string representing the name of the Event Condition
-    /// @param                  [in] aCriteria An enum indicating the criteria used to determine the Event Condition
+    /// @param                  [in] aCriteria An enum indicating the criteria used to determine if the Event Condition
```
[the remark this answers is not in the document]
vishwa2710: While technically correct I believe Criteria is used as a singular plural, like data. I think it's fine as it does not create ambiguity IMO

PR #196, opened 2023-08-23
refactor: CoordinateSubset class and methods

Derollez:
The goal of this refactor is to make CoordinateSubset instantiable (i.e. not a pure virtual class anymore) and move the `add`, `substract` and `inFrame` methods default implementation (default meaning invariant of the Frame) to the parent class `CoordinateSubset` and let children classes implement their own version when this is relevan


vishwa2710:
As we discussed on slack. Maybe it's best to just add these methods to the parent `CoordinateSubset` and then let others override it. This way we don't have an extra class. Otherwise looks great!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.cpp
```diff
@@ -54,6 +55,49 @@ Size CoordinatesSubset::getSize() const
...
+) const
+{
+    VectorXd coordinates = aCoordinatesBrokerSPtr->extractCoordinates(aFullCoordinatesVector, *this);
+
+    return coordinates;
+}
+
+Shared<const CoordinatesSubset> CoordinatesSubset::Mass()
+{
+    static const Shared<const CoordinatesSubset> mass = std::make_shared<CoordinatesSubset>("MASS", 1);
```
kyle-cochran, line 97: Does the mass value need to be static? I wonder if this could lead to weird behavior if you ever try to define two different `MASS` coordinate subsets.

Also might be worth removing the explicit `add` functions from other classes. Lemme know what you think about this:
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/202

Derollez: > Also might be worth removing the explicit add functions from other classes. Lemme know what you think about this

Absolutely 🙏

vishwa2710: So the reason this is static, is because we want to only create 1 pointer, and everytime this function is called, you just return the pointer to this (and thus increment the ownership by 1). Rather than creating a _new_ pointer and then allocating memory on the heap again. This is similar to how the Frame::GCRF() etc. works.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.cpp
```diff
@@ -54,6 +55,49 @@ Size CoordinatesSubset::getSize() const
...
+    const VectorXd& anotherFullCoordinatesVector,
+    [[maybe_unused]] const Shared<const Frame>& aFrameSPtr,
+    const Shared<const CoordinatesBroker>& aCoordinatesBrokerSPtr
+) const
+{
+    return aCoordinatesBrokerSPtr->extractCoordinates(aFullCoordinatesVector, *this) -
+           aCoordinatesBrokerSPtr->extractCoordinates(anotherFullCoordinatesVector, *this);
+}
+
+VectorXd CoordinatesSubset::inFrame(
```
vishwa2710, line 82: Adding a test for this would be nice.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.test.cpp
```diff
@@ -109,3 +80,80 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinatesSubset, Getter
...
+        VectorXd actual(3);
+
+        expected << 8.0, 5.0, 7.0;
+
+        actual = defaultCartesianPosition_.add(
+            Instant::Undefined(), input1, input2, Frame::Undefined(), defaultCoordinatesBroker_
+        );
+
+        EXPECT_EQ(expected, actual);
+    }
```
vishwa2710, line 116: IMO this test should be in CartesianPosition.test.cpp
When possible we don't want to have children objects required for instantiation in the parent file.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.test.cpp
```diff
@@ -109,3 +80,80 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinatesSubset, Getter
...
+        expected << -1.0;
+
+        actual = defaultCoordinateSubset_.subtract(
+            Instant::Undefined(), input1, input2, Frame::Undefined(), defaultCoordinatesBroker_
+        );
+
+        EXPECT_EQ(expected, actual);
+    }
+
+    {
```
vishwa2710, line 140: idem

vishwa2710 (approved):
Looks great man! Just 2 comments. Should be good to land after that.

vishwa2710 (approved):

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.hpp
phc1990, line 1: Maybe we want to add in the `Constructor` description that this is a frame invariant with linear +/- operations coordinate subset.

phc1990 (approved):
LGTM, left a small docs-related comment.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesSubset.hpp
Derollez: Yup, good point @phc1990, let me add this before pushing 🙏

Derollez: Done!

PR #208, opened 2023-08-29
feat: add TrajectorySegment class

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,143 @@
...
+using ostk::core::types::String;
+
+using ostk::physics::time::Instant;
+using ostk::physics::time::Duration;
+
+using ostk::astro::trajectory::State;
+using ostk::astro::flight::system::Dynamics;
+using ostk::astro::EventCondition;
+
+/// @brief                                  Represents a segment for astrodynamic calculations.
```
Derollez, line 35:
```suggestion
/// @brief                                  Represent a propagation segment for astrodynamics purposes
```
what do you think?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,143 @@
...
+    /// @brief                  Create a maneuvering segment.
+    ///
+    /// @param                  [in] anEventConditionSPtr An event condition.
+    /// @param                  [in] aThrusterDynamics Dynamics for the thruster.
+    /// @param                  [in] aNumericalSolver Numerical solver to be used. Defaults to undefined.
+    /// @return                 A TrajectorySegment for maneuvering.
+
+    static TrajectorySegment Maneuver(
+        const Shared<EventCondition>& anEventConditionSPtr,
+        const Shared<Dynamics>& aThrusterDynamics,  // TBM: This should be specifically ThrusterDynamics parent
```
Derollez, line 121: Yup agreed

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+    /// @brief                  Get dynamics.
+    /// @return                 Dynamics.
+
+    Array<Shared<Dynamics>> getDynamics() const;
+
+    /// @brief                  Get numerical solver.
+    /// @return                 Numerical solver.
+
+    NumericalSolver getNumericalSolver() const;
```
Derollez, line 75: Add `accessNumericalSolver` method and leverage in get?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,143 @@
...
+class TrajectorySegment
+{
+   public:
+    struct Solution
+    {
+       public:
+        Solution(const String& aName, const Array<Shared<Dynamics>>& aDynamicsArray, const Array<State>& aStates);
+
+        String name;                       /// Name of the segment.
+        Array<Shared<Dynamics>> dynamics;  /// List of dynamics used.
```
Derollez, line 45: Why do we need `dynamics` at Solution level?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,161 @@
...
+    ///
+    /// @param                  [in] aState Initial state for the segment.
+    /// @param                  [in] aDynamicsArray Array of dynamics. Defaults to an empty array.
+    /// @param                  [in] aNumericalSolver Numerical solver to be used. Defaults to undefined.
+    /// @param                  [in] maximumPropagationDuration Maximum duration for propagation. Defaults to 30 days.
+    /// @return                 A Solution representing the result of the solve.
+
+    Solution solve(
+        const State& aState,
+        const Array<Shared<Dynamics>>& aDynamicsArray = Array<Shared<Dynamics>>::Empty(),
```
kyle-cochran, line 103: Should this be called something like `extraDynamicsArray`? It looks like these get appended to whatever Dynamics the class is already holding.

Side-note: combining the dynamics and numerical solver (which are object bound things) with function input arguments feel a little weird to me. Seems like a strange mix of Object Oriented and Functional design. But maybe I'm missing something. What are your thoughts on reducing this to:
```
 Solution solve(
        const State& aState,
        const Duration& maximumPropagationDuration = Duration::Days(30.0)
    ) const;
```

And forcing dynamics/solver changes to use getters and setters?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
```diff
@@ -0,0 +1,161 @@
...
+    /// @param                  [in] anEventConditionSPtr An event condition.
+    /// @param                  [in] aDynamicsArray Array of dynamics. Defaults to an empty array.
+    /// @param                  [in] aNumericalSolver Numerical solver to be used. Defaults to undefined.
+    /// @return                 A TrajectorySegment for coasting.
+
+    static TrajectorySegment Coast(
+        const Shared<EventCondition>& anEventConditionSPtr,
+        const Array<Shared<Dynamics>>& aDynamicsArray = Array<Shared<Dynamics>>::Empty(),
+        const NumericalSolver& aNumericalSolver = NumericalSolver::Undefined()
+    );
```
kyle-cochran, line 126: Does it ever make sense to create a `Coast` arc without dynamics or a solver? I'm not quite seeing why these are optional. What am I missing?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.hpp
kyle-cochran: My biggest ick is that the numerical solver is an override here, while the dynamics are an addition, which is not very obvious IMO. 
But additionally the dynamics and integrator feel like core parts of the trajectory segment object, and so I think it makes sense to only treat them in an object-oriented way.

Let me know if something is going way over my head :)

kyle-cochran (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.cpp
```diff
@@ -0,0 +1,178 @@
...
+    const
+{
+    const Propagator propagator = {
+        numericalSolver_,
+        dynamics_,
+    };
+
+    const Instant startInstant = aState.getInstant();
+
+    // TBI: Handle the case where the condition is not met
```
Derollez, line 112: Do we need to handle this in this PR or make a separate one?

Derollez (approved):
LGTM!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/TrajectorySegment.cpp
vishwa2710: in another PR when we learn about how we want to handle this case.

PR #212, opened 2023-08-29
feat: Pass broker to conditions that require state knowledge

vishwa2710:


vishwa2710:
We're going to go with option 2.

PR #217, opened 2023-08-31
refactor: clean-up event condition tree

phc1990:
- Attempt to tidy up and simplify tree (see image below)
- `LogicalCondition` <> `Connective` follow the same design as `RealCondition` <> `Criterion`
- Need to rename `Criteria` to `Criterion`
- Need to fix bindings

image


vishwa2710 (approved):
Looks good man!

PR #220, opened 2023-09-06
fix: fix angle delta calculations in Event Conditions

vishwa2710:
Great suggestion Pau. It's slightly djanky but I think it works out quite well.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.hpp
@@ -1,0 +1,154 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_EventCondition_AngularCondition__
+#define __OpenSpaceToolkit_Astrodynamics_EventCondition_AngularCondition__
+
+#include <OpenSpaceToolkit/Core/Containers/Pair.hpp>
+#include <OpenSpaceToolkit/Core/Types/Real.hpp>
+#include <OpenSpaceToolkit/Core/Types/String.hpp>
+
+#include <OpenSpaceToolkit/Physics/Units/Derived/Angle.hpp>
+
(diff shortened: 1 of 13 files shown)


Derollez (approved):

phc1990 (commented):
Left suggestion to make it a bit "cleaner"

include/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+using ostk::core::types::Real;
+using ostk::core::types::String;
+using ostk::core::ctnr::Pair;
+
+using ostk::physics::units::Angle;
+
+using ostk::astro::trajectory::State;
+
+/// @brief                      An Event Condition that deals with angles.
```
kyle-cochran, line 29:
```suggestion
/// @brief                      An Event Condition that deals with angles. Automatically handles wrapping angle to [0,360).
```

Something like that? Just to explain why it's useful to have this separate. The problem that caused this isn't super obvious. We all missed it initially 😉

include/OpenSpaceToolkit/Astrodynamics/EventCondition/RealCondition.hpp
```diff
@@ -91,6 +91,14 @@ class RealCondition : public EventCondition
 
     Real evaluate(const State& state) const;
 
+    /// @brief                  Check if the Event Condition is satisfied based on current state and previous
```
kyle-cochran, line 94: If I'm understanding this re-org correctly, `RealCondition` is now meant for non-cyclic real numbers, and `AngularCondition` is meant for cyclic ones.  Wouldn't hurt to change this class description to something like:
```
/// @brief                      An Event Condition that can evaluate each state to a real number. Meant for non-cyclic continuous real numbers. For cyclic numbers (like angles), use AngularCondition.`
```

kyle-cochran (approved):

PR #222, opened 2023-09-14
feat: make atmospheric drag vary with mass, area, and Cd

kyle-cochran:
Atmospheric drag dynamics currently pull the Mass, Area, and coefficient of drag information from the Satellite System objects.

In order to vary those parameters in OD, they have been included in the State instead. This means the atmospheric drag dynamics no longer need to be instantiated with a Satellite System (which makes more sense IMO from a UX perspective).


bindings/python/test/flight/system/dynamics/test_atmospheric_drag.py
```diff
@@ -125,5 +127,5 @@ def test_compute_contribution(self, dynamics: AtmosphericDrag, state: State):
 
         assert len(contribution) == 3
         assert contribution == pytest.approx(
-            [0.0, -2.53370345e-05, 3.57473816e-12], abs=5e-11
+            [0.0, -0.0000278707803890, -0.0000000000197640], abs=5e-11
```
kyle-cochran, line 130: This is due to changing from wet mass to dry mass. I feel like the distinction is not necessary in the context of just drag tests with no thrusters.

bindings/python/test/trajectory/test_propagator.py
```diff
@@ -107,11 +128,41 @@ def state(
     return State(instant, coordinates, Frame.GCRF(), coordinates_broker)
 
 
+@pytest.fixture
```
kyle-cochran, line 131: Needed because the other one is too high for the exponential model

bindings/python/test/flight/system/dynamics/test_atmospheric_drag.py
vishwa2710: Also I don't think it's necessary to have tests here where the numbers match up. We can just keep is not None or == the correct size etc. something simple.

vishwa2710 (approved):

bindings/python/test/flight/system/dynamics/test_atmospheric_drag.py
Derollez: Noticed that too. Should go away with the rebase.

> keep is not None or == the correct size etc. something simple.

I also agree with this

bindings/python/test/trajectory/test_propagator.py
```diff
@@ -74,16 +83,26 @@ def satellite_system(propulsion_system: PropulsionSystem) -> SatelliteSystem:
...
 @pytest.fixture
 def environment() -> Environment:
-    return Environment.default()
+    sun = Sun.default()
+
+    earth: Earth = Earth.from_models(
+        EarthGravitationalModel(EarthGravitationalModel.Type.EGM96),
+        EarthMagneticModel(EarthMagneticModel.Type.Undefined),
+        EarthAtmosphericModel(EarthAtmosphericModel.Type.Exponential),
+    )
```
Derollez, line 92: Add as a fixture itself?

test/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.test.cpp
```diff
@@ -85,26 +86,12 @@ class OpenSpaceToolkit_Astrodynamics_Flight_System_Dynamics_AtmosphericDrag : pu
...
-        const Mass propellantMass = Mass::Kilograms(0.0);
-
-        satelliteSystem_ = {
-            satelliteDryMass,
-            satelliteGeometry,
-            Matrix3d::Identity(),
-            500.0,
-            2.1,
-        };
+        const Real mass = 100;
```
Derollez, line 89: Will be worth leveraging the `SatelliteSystem::Default()` that @vishwa2710 is adding in https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/223.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -102,6 +102,8 @@ using ostk::astro::trajectory::state::CoordinatesSubset;
 using ostk::astro::trajectory::state::CoordinatesBroker;
 using ostk::astro::trajectory::Propagator;
 using ostk::astro::trajectory::state::NumericalSolver;
+using ostk::astro::trajectory::state::CoordinatesSubset;
+using ostk::astro::trajectory::state::CoordinatesBroker;
```
Derollez, line 106: TBR, added above. Remove the ones above maybe

test/OpenSpaceToolkit/Astrodynamics/Flight/System/Dynamics/AtmosphericDrag.test.cpp
kyle-cochran: True, that will make things a little cleaner. Will add a todo for now to change over

PR #223, opened 2023-09-14
feat: add Sequence class

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -0,0 +1,151 @@
...
+    /// @param                  [in] aThruster A thruster dynamics.
+
+    void addManeuverSegment(const Shared<EventCondition>& anEventConditionSPtr, const Shared<Thruster>& aThruster);
+
+    /// @brief                  Solve the sequence given an initial state.
+    ///
+    /// @param                  [in] aState Initial state for the sequence.
+    /// @return                 A Solution that contains solutions for each segment.
+
+    Solution solve(const State& aState) const;
```
phc1990, line 130: Nit: Maybe `run`/`execute`/etc. Coming from the outside, it looked to me as it `solve` was actually finding something out for us. Maybe we can keep `solve` for when we get to differential correction, maneuver planning, etc.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -0,0 +1,151 @@
...
+    ///
+    /// @param                  [in] aTrajectorySegment A trajectory segment.
+
+    void addSegment(const TrajectorySegment& aTrajectorySegment);
+
+    /// @brief                  Add multiple segments.
+    ///
+    /// @param                  [in] aTrajectorySegmentArray An array of trajectory segments.
+
+    void addSegment(const Array<TrajectorySegment>& aTrajectorySegmentArray);
```
phc1990, line 110: maybe `addSegments` (plural)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -0,0 +1,151 @@
...
+    ///
+    /// @return                 Maximum propagation duration.
+
+    Duration getMaximumPropagationDuration() const;
+
+    /// @brief                  Add a trajectory segment.
+    ///
+    /// @param                  [in] aTrajectorySegment A trajectory segment.
+
+    void addSegment(const TrajectorySegment& aTrajectorySegment);
```
phc1990, line 104: Should they be Shared pointers? like the event condition below?

phc1990 (commented):
Left some minor comments and a "philosophical" question.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+using ostk::astro::trajectory::TrajectorySegment;
+using ostk::astro::trajectory::State;
+using ostk::astro::flight::system::SatelliteSystem;
+using ostk::astro::flight::system::dynamics::Thruster;
+
+/// @brief                      Represents a sequence of trajectory segments executed in order.
+
+class Sequence
+{
```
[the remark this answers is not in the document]
vishwa2710: Yeah you're 100% right, we want to support this behaviour. I believe we can do this in a follow up MR. Another idea I had to implement this without interfaces would be adding a method like:
```
void addSegmentsFromSequence(const Sequence& aSequence)
{
  for (const auto& segment: aSequence.accessSegments())
  {
     this->segments_.add(segment);
  }
}
```
If that makes sense. We would have to link the segment back to the sequence somehow maybe for bookkeeping purposes. But we can figure that out down the line. I like your idea as well, we can play with both.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
vishwa2710: Currently we only have a single Segment class, so I don't believe there's any value in having a shared pointer.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
vishwa2710: I'm leaving it as is so that it's overloaded, which makes it a bit simpler, we don't have to keep track of whether it's a single object or an array. It's easier in python as well.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -0,0 +1,151 @@
...
+using ostk::physics::Environment;
+
+using ostk::astro::trajectory::TrajectorySegment;
+using ostk::astro::trajectory::State;
+using ostk::astro::flight::system::SatelliteSystem;
+using ostk::astro::flight::system::dynamics::Thruster;
+
+/// @brief                      Represents a sequence of trajectory segments executed in order.
+
+class Sequence
```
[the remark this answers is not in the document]
vishwa2710: Hmmmmm fair point. I can rename.

src/OpenSpaceToolkit/Astrodynamics/Flight/System/SatelliteSystem.cpp
```diff
@@ -144,6 +144,22 @@ SatelliteSystem SatelliteSystem::Undefined()
...
 }
 
+SatelliteSystem SatelliteSystem::Default()
+{
+    const Composite satelliteGeometry(Cuboid(
+        {0.0, 0.0, 0.0}, {Vector3d {1.0, 0.0, 0.0}, Vector3d {0.0, 1.0, 0.0}, Vector3d {0.0, 0.0, 1.0}}, {1.0, 2.0, 3.0}
+    ));
+
+    return {
+        Mass(1000.0, Mass::Unit::Kilogram),
```
kyle-cochran, line 154: nit:
```suggestion
        Mass(100.0, Mass::Unit::Kilogram),
```

Might be closer to the average satellite.

src/OpenSpaceToolkit/Astrodynamics/Flight/System/SatelliteSystem.cpp
```diff
@@ -144,6 +144,22 @@ SatelliteSystem SatelliteSystem::Undefined()
...
+        satelliteGeometry,
+        Matrix3d::Identity(),
+        1.0,
+        2.1,
```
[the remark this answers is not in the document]
kyle-cochran: I'm nit-picking here because I've been thinking lately (and I think we spoke about this as well) about how defaulting the propagator with sensible number to help get a user off the ground before they start caring about building a satellite system.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Models/Kepler/COE.cpp
```diff
@@ -483,7 +483,7 @@ COE COE::Cartesian(const COE::CartesianState& aCartesianState, const Derived& aG
 
         aop_rad = 0.0;
 
-        nu_rad = std::acos((nodeVector.dot(positionVector)) / (node * position));
+        nu_rad = std::acos(std::clamp((double)(nodeVector.dot(positionVector) / (node * position)), -1.0, 1.0));
```
kyle-cochran, line 486: Worth adding a test for this case? Seems like an important functional part of the code.

phc1990 (approved):
LGTM!

PR #226, opened 2023-09-21
feat: add getCoordinatesSubsets method to States

kyle-cochran:
Exposing the list of subsets would be useful for the OD changes in FDS toolbox, so we don't have to pass around the original broker they were created with.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
@@ -49,6 +49,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State(pybind11::module& a
         .def("get_frame", &State::getFrame)
 
         .def("extract_coordinates", &State::extractCoordinates, arg("coordinates_subset"))
+        .def("get_coordinates_subsets", &State::getCoordinatesSubsets)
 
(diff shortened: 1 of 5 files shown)


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
```diff
@@ -49,6 +49,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State(pybind11::module& a
         .def("get_frame", &State::getFrame)
 
         .def("extract_coordinates", &State::extractCoordinates, arg("coordinates_subset"))
+        .def("get_coordinates_subsets", &State::getCoordinatesSubsets)
```
vishwa2710, line 52: Group this above with the rest of the getters?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
```diff
@@ -93,6 +95,8 @@ class State
 
     VectorXd extractCoordinates(const Shared<const CoordinatesSubset>& aSubetSPtr) const;
 
+    const Array<Shared<const CoordinatesSubset>> getCoordinatesSubsets() const;
```
vishwa2710, line 98: Move this with the rest of the getters?

vishwa2710 (approved):

PR #231, opened 2023-10-05
feat: update State printing to support extended state

kyle-cochran:
When printing the State, you currently see:
```
-- Trajectory :: State -----------------------------------------------------------------------------
    Instant:                                 2018-01-01 00:00:00 [UTC]                
    Position:                                [1.200000000000, 3.400000000000, 5.600000000000] [m] @ GCRF 
    Velocity:                                [7.800000000000, 9.000000000000, 1.200000000000] [m/s] @ GCRF 
----------------------------------------------------------------------------------------------------
```

This updates the print to support extending the state beyond position/velocity. So now it looks like:
```
-- Trajectory :: State -----------------------------------------------------------------------------
    Instant:                                 2018-01-01 00:00:00 [UTC]                
    Frame:                                   GCRF                                     
    CARTESIAN_POSITION                       [1.2000, 3.4000, 5.6000]                 
    CARTESIAN_VELOCITY                       [7.8000, 9.0000, 1.2000]                 
----------------------------------------------------------------------------------------------------
```

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
@@ -1,5 +1,7 @@
 /// Apache License 2.0
 
+// #include <Eigen/Core>
+
 #include <OpenSpaceToolkit/Core/Error.hpp>
 #include <OpenSpaceToolkit/Core/Utilities.hpp>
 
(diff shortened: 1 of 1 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -354,10 +356,25 @@ void State::print(std::ostream& anOutputStream, bool displayDecorator) const
...
 
+    if (!this->isDefined())
+    {
+        ostk::core::utils::Print::Line(anOutputStream) << "Coordinates: Undefined";
+    }
+    else
+    {
+        const Array<Shared<const CoordinatesSubset>> subsets = this->coordinatesBrokerSPtr_->getSubsets();
+
+        ostk::core::utils::Print::Line(anOutputStream) << "Coordinates: " << this->accessCoordinates().toString(4);
```
vishwa2710, line 369:
```suggestion
```
I think that we can skip printing the coordinates since we're printing it piece by piece below.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -354,10 +356,25 @@ void State::print(std::ostream& anOutputStream, bool displayDecorator) const
...
+
+        ostk::core::utils::Print::Line(anOutputStream) << "Where:";
+        for (auto subset : subsets)
```
vishwa2710, line 372:
```suggestion
        for (const auto& subset : subsets)
```

vishwa2710 (approved):
Just some nits.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
kyle-cochran: My thinking was that it could be nice in case we wanted to copy-pasta the coordinates, but looks like we didn't have that before...  Going to close this suggestion and implement this locally so I can also remove the "Where" line

kyle-cochran: (implemented)

PR #233, opened 2023-10-06
feat: return conditional solution for segment and sequences

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -35,6 +35,7 @@ class Sequence
     struct Solution
     {
         Array<Segment::Solution> segmentSolutions;
+        bool sequenceIsComplete;
```
kyle-cochran, line 38: nit: what do you think about something like `sequenceExecutionIsComplete`. It took me a second to understand that this didn't mean something like "the sequence is fully constructed and valid for execution"

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -168,14 +168,17 @@ class Propagator
...
+    ///                     anEventCondition);
     /// @endcode
     /// @param              [in] aState An initial state
     /// @param              [in] anInstant An instant
     /// @param              [in] anEventCondition An event condition
-    /// @return             State
+    /// @return             NumericalSolver::ConditionSolution
 
-    State calculateStateAt(const State& aState, const Instant& anInstant, const EventCondition& anEventCondition) const;
+    NumericalSolver::ConditionSolution calculateStateAt(
```
kyle-cochran, line 179: I think it could be confusing to have two functions `calculateStateAt`/`calculateStatesAt` that do indeed return States and one `calculateStateAt` that does not.

What do you think of either making a separate function or renaming the current function to something along the lines of `propagateStateToCondition` to use for this case?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.test.cpp
```diff
@@ -235,28 +235,49 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Sequence, AddManeuverSegment)
 
         EXPECT_TRUE(defaultSequence_.getSegments().getSize() == segmentsCount + 1);
     }
+
+    {
+    }
```
kyle-cochran, line 240:
```suggestion
```

typo?

vishwa2710: Yup! Nice catch.

vishwa2710 (commented):
Thanks for the review!

kyle-cochran (approved):

PR #235, opened 2023-10-07
refactor: Use the built-in State method to extract coordinates

kyle-cochran:
Use the built-in State method to extract multiple subsets at once, which makes a private function in Propagator obsolete.

Also changes the Numerical Solver State to _always_ be in the propagation Frame since we are _always_ converting the coordinates to that Frame anyway. Otherwise if the input State ever had a different Frame that the Propagation Frame, it would just silently give you crazy numbers.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
@@ -145,8 +145,8 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
 
     const State state = {
         aState.accessInstant(),
-        extractCoordinatesFromState(aState.inFrame(Propagator::IntegrationFrameSPtr)),
-        aState.accessFrame(),
+        aState.inFrame(Propagator::IntegrationFrameSPtr).extractCoordinates(this->coordinatesBrokerSPtr_->getSubsets()),
+        Propagator::IntegrationFrameSPtr,
         this->coordinatesBrokerSPtr_,
     };
 
@@ -170,8 +170,8 @@ NumericalSolver::ConditionSolution Propagator::calculateStateToCondition(
 
     const State state = {
         aState.accessInstant(),
-        extractCoordinatesFromState(aState.inFrame(Propagator::IntegrationFrameSPtr)),
-        aState.accessFrame(),
+        aState.inFrame(Propagator::IntegrationFrameSPtr).extractCoordinates(this->coordinatesBrokerSPtr_->getSubsets()),
+        Propagator::IntegrationFrameSPtr,
         this->coordinatesBrokerSPtr_,
     };
 
@@ -212,8 +212,8 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
 
     const State state = {
         aState.accessInstant(),
-        extractCoordinatesFromState(aState.inFrame(Propagator::IntegrationFrameSPtr)),
-        aState.accessFrame(),
+        aState.inFrame(Propagator::IntegrationFrameSPtr).extractCoordinates(this->coordinatesBrokerSPtr_->getSubsets()),
+        Propagator::IntegrationFrameSPtr,
         this->coordinatesBrokerSPtr_,
     };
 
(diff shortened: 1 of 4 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -145,8 +145,8 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
 
     const State state = {
         aState.accessInstant(),
-        extractCoordinatesFromState(aState.inFrame(Propagator::IntegrationFrameSPtr)),
-        aState.accessFrame(),
+        aState.inFrame(Propagator::IntegrationFrameSPtr).extractCoordinates(this->coordinatesBrokerSPtr_->getSubsets()),
+        Propagator::IntegrationFrameSPtr,
```
kyle-cochran, line 149: Someone double check me here. Since we're always converting the coordinates to the Propagator Frame, shouldn't we always be setting the State to that Frame as well?

vishwa2710: That's a good point. I think it was working out cause we convert everything to further up the chain. Great catch!

vishwa2710 (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
```diff
@@ -49,7 +49,16 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State(pybind11::module& a
         .def("get_coordinates_subsets", &State::getCoordinatesSubsets)
         .def("get_frame", &State::getFrame)
 
-        .def("extract_coordinates", &State::extractCoordinates, arg("coordinates_subset"))
+        .def(
+            "extract_coordinates",
+            overload_cast<const Shared<const CoordinatesSubset>&>(&State::extractCoordinates, const_),
```
apaletta3, line 54: What does this casting do? And the `const_` is a keyword that pybind recognizes?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesBroker.test.cpp
```diff
@@ -346,4 +347,55 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinatesBroker, Extrac
 
         EXPECT_ANY_THROW(broker.extractCoordinates(fullCoordinatesVector, subset_4));
     }
+
+    {
+        CoordinatesBroker broker = CoordinatesBroker();
+        broker.addSubset(subset_1);
```
apaletta3, line 353: Didn't know you could make these fixture like objects like subset_1 in cpp. Very cool. Should its name be more descriptive than _1 for test legibility perhaps?

apaletta3 (commented):
Looks nice! I gotta get back into cpp lol

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
kyle-cochran: The cast is needed because Python has no concept of function overloading like in strongly typed languages (C++/Java). This cast is just a PyBind construct to let it know that it needs to do some magic in the backend to handle that.

`const` is to indicate that the C++ function is `const`. Yeah it's just another PyBind construct as well.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinatesBroker.test.cpp
kyle-cochran: Yeah, they are effectively just C++ class variables. All of these test functions are considered part of the test class because they use `TEST_F` in GTest rather that `TEST`. It's pretty much just a class method accessing a class variable. `this->subset_1` would be equivalent, but you don't necessarily have to specify `this` in C++ as long as there's no naming collision.

Good looking out. In this case, the subsets don't represent anything physical (the names are like "S1", "S2", etc.), so I think the naming is okay.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
kyle-cochran: (`const` for a class function just means that it's not allowed to modify any of the class attributes)
https://www.geeksforgeeks.org/const-member-functions-c/

PR #237, opened 2023-10-10
feat: add StateBuilder

kyle-cochran:
Adds a StateBuilder class to enable creating batches of States that share the same `Coordinates Broker`, without a user needing to interact with the `CoordinateBroker` concept at all.

The main function is 
- `StateBuilder::buildState(Instant, VectorXd)` [Cpp]
- `StateBuilder.state(Instant, np.ndarray)` [Python]

and the flow in Python looks like:

```python
builder: StateBuilder = StateBuilder(
   Frame.gcrf(),
   [CartesianPosition.default(), CartesianVelocity.default()]
)

start_instant = Instant.date_time(datetime(2022, 1, 1))
start_coords = np.array([1,2,3,4,5,6])

states = []
for i in range(100):
    states.append(builder.state(
        start_instant+Duration.seconds(i),
        start_coords*i,
    ))
```

The MR is largely boilerplate copied from the `State` class. New functionality is mostly in the `buildState` method.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
@@ -1,0 +1,47 @@
+/// Apache License 2.0
+
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp>
+
+inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_StateBuilder(pybind11::module& aModule)
+{
+    using namespace pybind11;
+
+    using ostk::core::types::Shared;
+    using ostk::core::ctnr::Array;
+
+    using ostk::physics::coord::Frame;
+
+    using ostk::astro::trajectory::State;
+    using ostk::astro::trajectory::StateBuilder;
+    using ostk::astro::trajectory::state::CoordinatesBroker;
+
+    class_<StateBuilder>(aModule, "StateBuilder")
+
+        .def(
+            init<const Shared<const Frame>&, const Array<Shared<const CoordinatesSubset>>&>(),
+            arg("frame"),
+            arg("coordinates_subsets")
+        )
+        .def(
+            init<const Shared<const Frame>&, const Shared<const CoordinatesBroker>&>(),
+            arg("frame"),
+            arg("coordinates_broker")
+        )
+
+        .def(self == self)
+        .def(self != self)
+
+        .def("__str__", &(shiftToString<StateBuilder>))
+        .def("__repr__", &(shiftToString<StateBuilder>))
+
+        .def("is_defined", &StateBuilder::isDefined)
+
+        .def("state", &StateBuilder::buildState)
+
+        .def("get_coordinates_subsets", &StateBuilder::getCoordinatesSubsets)
+        .def("get_frame", &StateBuilder::getFrame)
+
+        .def_static("undefined", &StateBuilder::Undefined)
+
+        ;
+}
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
@@ -36,7 +36,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_StateBuilder(pybind11::mo
 
         .def("is_defined", &StateBuilder::isDefined)
 
-        .def("state", &StateBuilder::buildState)
+        .def("build_state", &StateBuilder::buildState)
 
         .def("get_coordinates_subsets", &StateBuilder::getCoordinatesSubsets)
         .def("get_frame", &StateBuilder::getFrame)
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp
@@ -1,7 +1,7 @@
 /// Apache License 2.0
 
-#ifndef __OpenSpaceToolkit_Astrodynamics_Trajectory_State__
-#define __OpenSpaceToolkit_Astrodynamics_Trajectory_State__
+#ifndef __OpenSpaceToolkit_Astrodynamics_Trajectory_StateBuilder__
+#define __OpenSpaceToolkit_Astrodynamics_Trajectory_StateBuilder__
 
 #include <OpenSpaceToolkit/Core/Containers/Array.hpp>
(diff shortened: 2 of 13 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp
```diff
@@ -0,0 +1,142 @@
...
+    ///
+    /// @return                 True if the StateBuilder is defined, false otherwise
+
+    bool isDefined() const;
+
+    /// @brief                  Produce a State linked to the Frame and Coordinates Broker of the StateBuilder.
+    ///
+    /// @return                 A State linked to the Frame and Coordinates Broker of the StateBuilder
+
+    const State buildState(const Instant& anInstant, const VectorXd& aCoordinates) const;
```
vishwa2710, line 94:
```suggestion
    const State build(const Instant& anInstant, const VectorXd& aCoordinates) const;
```
Nit: since it's called StateBuilder, it's maybe okay to leave out the `State` as it's known what it will be building.

vishwa2710 (approved):
Looks great. Just 1 nit.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp
kyle-cochran: Yeah I do agree here. That puts us at:
`StateBuilder::build` for C++ and
`StateBuilder.state` for Python

If we're going to be brief with the names, should probably be consistent.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+        return false;
+    }
+
+    if (this->frameSPtr_ != aStateBuilder.frameSPtr_)
+    {
+        return false;
+    }
+
+    for (const Shared<const CoordinatesSubset>& subset : this->coordinatesBrokerSPtr_->accessSubsets())
```
phc1990, line 42: It would not be enough to check that both brokers contain each other's subsets.
One would need to check that the subsets occupy the same index in each broker.

However, luckily for us, "somebody" went a bit overboard with the `CoordinatesBroker` `==` operator so that it hard checks all of this (dimension, subsets, and ordering), so all you need to do right after comparing the frame is:

```cpp
if (this -> coordinatesBrokerSPtr_ != aStateBuilder.coordinatesBrokerSPtr_)
{
    return false;
}
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.test.cpp
```diff
@@ -0,0 +1,367 @@
...
+    {
+        const StateBuilder aStateBuilder = {Frame::GCRF(), posVelBrokerSPtr};
+
+        const Shared<const CoordinatesBroker> velPosBrokerSPtr = std::make_shared<CoordinatesBroker>(
+            CoordinatesBroker({CartesianVelocity::Default(), CartesianPosition::Default()})
+        );
+
+        const StateBuilder anotherStateBuilder = {Frame::GCRF(), velPosBrokerSPtr};
+
+        EXPECT_TRUE(aStateBuilder == anotherStateBuilder);
```
phc1990, line 109: ...following my previous comment on `stateBuilder` comparison, this should now be `EXPECT_FALSE`, as coordinates subset do not take up the same position.

`aStateBuilder.build(1,1,1,2,2,2)` would yield pos{1,1,1} and vel{2,2,2}

BUT

`anotherStateBuilder.build(1,1,1,2,2,2)` would yield pos{2,2,2} and vel {1,1,1}

test/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.test.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+        EXPECT_FALSE(aStateBuilder == anotherStateBuilder);
+    }
+
+    {
+        const StateBuilder aStateBuilder = {Frame::GCRF(), posVelBrokerSPtr};
+        const StateBuilder anotherStateBuilder = {Frame::GCRF(), posVelMassBrokerSPtr};
+
+        EXPECT_FALSE(aStateBuilder == anotherStateBuilder);
+        EXPECT_FALSE(anotherStateBuilder == aStateBuilder);
```
phc1990, line 85: leftover line?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.test.cpp
```diff
@@ -0,0 +1,367 @@
...
+    {
+        const StateBuilder aStateBuilder = {Frame::GCRF(), posVelBrokerSPtr};
+
+        const Shared<const CoordinatesBroker> velPosBrokerSPtr = std::make_shared<CoordinatesBroker>(
+            CoordinatesBroker({CartesianVelocity::Default(), CartesianPosition::Default()})
+        );
+
+        const StateBuilder anotherStateBuilder = {Frame::GCRF(), velPosBrokerSPtr};
+
+        EXPECT_FALSE(aStateBuilder != anotherStateBuilder);
```
phc1990, line 186: ...and then following on the previous comment, this one would be `EXPECT_TRUE`

include/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp
phc1990: ➕ on just `.build` for both languages

phc1990 (commented):
Looks very good to me.
Left some comments.
Only major thing is on builder comparison.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.test.cpp
kyle-cochran: I the current comparison, it checks that each contain each other, so the reverse comparison is needed for full coverage

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+        .def(self == self)
+        .def(self != self)
+
+        .def("__str__", &(shiftToString<StateBuilder>))
+        .def("__repr__", &(shiftToString<StateBuilder>))
+
+        .def("is_defined", &StateBuilder::isDefined)
+
+        .def("build_state", &StateBuilder::buildState)
```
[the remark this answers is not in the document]
kyle-cochran: As mentioned on slack. Gonna go forward with build_state for _this MR_. We can continue the naming discussion in the next MR because I think that one has a bit better context.

PR #238, opened 2023-10-11
feat: add State constructor with Array of Subsets

kyle-cochran:
Adds a constructor to the State class to directly build a custom State from an Array of Subsets.

This constructor will presumably cause a larger memory footprint, but is more straightforward to use than the Broker constructor.

This downfall should be mitigated in Python with the introduction of the CustomState meta-class stuff.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
@@ -59,7 +59,8 @@ class State
     );
 
     /// @brief                  Constructor. This constructor makes a new Coordinates Broker under the hood for every
-    /// State.
+    /// State. When possible, users should prefer passing in an existing Coordinates Broker or using a StateBuilder to
+    /// reduce memory footprint when constructing many states.
     ///
     /// @param                  [in] anInstant An instant
     /// @param                  [in] aCoordinates The coordinates at the instant in International System of Units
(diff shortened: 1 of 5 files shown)


test/OpenSpaceToolkit/Astrodynamics/Trajectory/State.test.cpp
```diff
@@ -38,6 +38,17 @@ TEST(OpenSpaceToolkit_Astrodynamics_Trajectory_State, Constructor)
...
 
+    {
+        const Instant instant = Instant::DateTime(DateTime(2018, 1, 1, 0, 0, 0), Scale::UTC);
+        VectorXd coordinates(6);
+        coordinates << 1.0, 2.0, 3.0, 4.0, 5.0, 6.0;
+
+        const Array<Shared<const CoordinatesSubset>> subsets = {
+            CartesianPosition::Default(), CartesianVelocity::Default()
+        };
+        EXPECT_NO_THROW(State state(instant, coordinates, Frame::GCRF(), subsets));
```
phc1990, line 49: probably want to check that position is indeed `{1,2,3}` and that velocity is indeed `{4,5,6}`

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
```diff
@@ -59,7 +58,23 @@ class State
         const Shared<const CoordinatesBroker>& aCoordinatesBrokerSPtr
     );
 
-    /// @brief                  Constructor.
+    /// @brief                  Constructor. This constructor makes a new Coordinates Broker under the hood for every
+    /// State.
```
phc1990, line 62: Maybe add an extra sentence talking about the footprint it leaves and why using the other constructor is encouraged (without making this constructor look bad though 😂 )

phc1990 (approved):
LGTM, left minor comments.

PR #244, opened 2023-10-12
feat: Python class meta-programming for State construction

kyle-cochran:
This MR is some Python-only syntactic sugar on top of the State Builder concept which allows for some nice syntax:
```python
MySuperFunState: type = State.template(
    frame,
    [CartesianPosition.default(), CartesianVelocity.default()],
)

states: MySuperFunState = [
    MySuperFunState(instant,coords) for instant,coords in zip(instants, coordinates)
]
```

The output StateTypeX classes are polymorphic to State and so can be used anywhere a State can.
This example is 100% equivalent to using a builder and is just a syntax alternative.

(other meta-programming features like auto-assigning `state.mass` to get the mass value will be in a follow-up)

### DIFF
--- a/bindings/python/test/trajectory/test_state.py
+++ b/bindings/python/test/trajectory/test_state.py
@@ -104,7 +104,7 @@ class TestState:
         assert state is not None
         assert isinstance(state, State)
         assert state.is_defined()
-    
+
     def test_custom_class_generator(
         self,
         instant: Instant,
--- a/bindings/python/test/trajectory/test_state.py
+++ b/bindings/python/test/trajectory/test_state.py
@@ -105,7 +105,7 @@ class TestState:
         assert isinstance(state, State)
         assert state.is_defined()
 
-    def test_custom_class_generator(
+    def test_custom_state_class_generator(
         self,
         instant: Instant,
         position: Position,
--- a/bindings/python/test/trajectory/test_state.py
+++ b/bindings/python/test/trajectory/test_state.py
@@ -11,7 +11,7 @@ from ostk.physics.coordinate import Position
 from ostk.physics.coordinate import Velocity
 from ostk.physics.coordinate import Frame
 
-from ostk.astrodynamics.trajectory import State
+from ostk.astrodynamics.trajectory import State, StateBuilder
 from ostk.astrodynamics.trajectory.state import CoordinatesBroker
 from ostk.astrodynamics.trajectory.state.coordinates_subset import (
     CartesianPosition,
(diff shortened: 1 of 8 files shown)


bindings/python/test/trajectory/test_state.py
```diff
@@ -11,7 +11,7 @@
 from ostk.physics.coordinate import Velocity
 from ostk.physics.coordinate import Frame
 
-from ostk.astrodynamics.trajectory import State
+from ostk.astrodynamics.trajectory import State, StateBuilder
```
kyle-cochran, line 14:
```suggestion
from ostk.astrodynamics.trajectory import State
```

bindings/python/tools/python/ostk/astrodynamics/__init__.py
```diff
@@ -3,3 +3,9 @@
 from ostk.physics import *
 
 from .OpenSpaceToolkitAstrodynamicsPy import *
+
+from .pytrajectory.pystate import State as PyState
```
kyle-cochran, line 7: Curious if there are any better ideas here. This works, but it means State is imported whenever anything else in Astro is imported. State is already very pervasive, so it may not make much difference, but it's still messy.

bindings/python/test/trajectory/test_state.py
```diff
@@ -105,6 +105,38 @@ def test_subsets_constructor(
         assert isinstance(state, State)
         assert state.is_defined()
 
+    def test_custom_class_generator(
```
vishwa2710, line 108:
```suggestion
    def test_custom_state_class_generator(
```
Nit af:

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
```diff
@@ -83,6 +83,19 @@ class State
...
 
+    /// @brief                  Copy constructor.
+    ///
+    /// @param                  [in] aState An existing State
+
+    State(const State& aState);
+
+    /// @brief                  Copy-assignment operator
+    ///
+    /// @param                  [in] aState The State to compare to
```
vishwa2710, line 94:
```suggestion
    /// @param                  [in] aState The State to copy
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
```diff
@@ -83,6 +83,19 @@ class State
 
     State(const Instant& anInstant, const Position& aPosition, const Velocity& aVelocity);
 
+    /// @brief                  Copy constructor.
+    ///
+    /// @param                  [in] aState An existing State
+
+    State(const State& aState);
+
+    /// @brief                  Copy-assignment operator
```
vishwa2710, line 92: Why are we adding this? 🤔 Doesn't seem related to the Meta-class?

vishwa2710 (approved):
LGTM!
It's gonna be sick when we add the custom kwargs and getters 🥳

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp
kyle-cochran: The copy constructor is needed here:
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/244/files#diff-f995720c611c0d79a20e7b908e36d9360719bb3784955f510982c459db799006R32

When sub-classing a `State`, we need to call it's `__init__` during construction of the sub-class. In this case, I don't want the `State` object to actually be produced using the normal `State` constructor, but via a `StateBuilder`.

There could be a way around this by ignoring the `StateBuilder`, but I think it's a good idea to use it so we never have to re-code all of the extra stuff inside of it.


The copy assignment operator is because after creating the copy constructor, the compiler was giving warnings that it could not use the default copy operator anymore, so I added an explicit one.

phc1990 (commented):
I wonder whether we can achieve this by using bindings directly and not Python code 🤔 
I just dont' know

PR #246, opened 2023-10-13
feat: make the propagator maintain the frame and size of the input State

kyle-cochran:
Currently, the propagator converts any input state into the correct dimensions and Frame to propagate. 

This is fine, but get's confusing when you pass in a higher dimensional State in ITRF and get back a lower dimensional one in GCRF.

This MR introduces a layer that transforms the input/output State to/from the State that the numerical solver uses using the new StateBuilder class.


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -33,7 +33,8 @@ using ostk::astro::trajectory::state::CoordinatesSubset;
 const Shared<const Frame> Propagator::IntegrationFrameSPtr = Frame::GCRF();
 
 Propagator::Propagator(const NumericalSolver& aNumericalSolver, const Array<Shared<Dynamics>>& aDynamicsArray)
-    : dynamicsContexts_(),
+    : stateBuilder_(Propagator::IntegrationFrameSPtr, coordinatesBrokerSPtr_),
```
vishwa2710, line 36: Worth noting that the coordinatesBroker_ here will be nullptr.
It gets populated downstream but just thought it was worth mentioning.
I think it's fine cause it's a shared pointer.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -143,18 +144,18 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
...
+    const State solverOutputState = numericalSolver_.integrateTime(
+        solverInputState,
         anInstant,
-        Dynamics::GetSystemOfEquations(this->dynamicsContexts_, state.accessInstant(), Propagator::IntegrationFrameSPtr)
+        Dynamics::GetSystemOfEquations(
+            this->dynamicsContexts_, solverInputState.accessInstant(), Propagator::IntegrationFrameSPtr
+        )
     );
+
+    const StateBuilder outputStateBuilder(aState);
```
vishwa2710, line 157:
```suggestion
    const StateBuilder outputStateBuilder = {aState};
```
Nit: Just for consistency

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -143,18 +144,18 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
...
+    return outputStateBuilder.expand(solverOutputState.inFrame(aState.getFrame()), aState);
```
vishwa2710, line 158:
```suggestion

    return outputStateBuilder.expand(solverOutputState.inFrame(aState.accessFrame()), aState);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -168,20 +169,18 @@ NumericalSolver::ConditionSolution Propagator::calculateStateToCondition(
...
-    const NumericalSolver::ConditionSolution conditionSolution = numericalSolver_.integrateTime(
-        state,
+    NumericalSolver::ConditionSolution conditionSolution = numericalSolver_.integrateTime(
+        solverInputState,
         anInstant,
         Dynamics::GetSystemOfEquations(this->dynamicsContexts_, startInstant, Propagator::IntegrationFrameSPtr),
         anEventCondition
     );
 
+    const StateBuilder outputStateBuilder(aState);
```
vishwa2710, line 181:
```suggestion
    const StateBuilder outputStateBuilder = {aState};
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -168,20 +169,18 @@ NumericalSolver::ConditionSolution Propagator::calculateStateToCondition(
...
+
+    conditionSolution.state = outputStateBuilder.expand(conditionSolution.state.inFrame(aState.getFrame()), aState);
     return conditionSolution;
```
vishwa2710, line 184:
```suggestion
    conditionSolution.state = outputStateBuilder.expand(conditionSolution.state.inFrame(aState.accessFrame()), aState);

    return conditionSolution;
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -254,15 +250,25 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
...
         std::reverse(backwardPropagatedStates.begin(), backwardPropagatedStates.end());
     }
 
-    return backwardPropagatedStates + forwardPropagatedStates;
+    const Array<State> solverOutputStates = backwardPropagatedStates + forwardPropagatedStates;
+
+    Array<State> outputStates;
+    outputStates.reserve(solverOutputStates.getSize());
+
+    for (const State& solverOutputState : solverOutputStates)
```
vishwa2710, line 266:
```suggestion
    Array<State> outputStates;
    outputStates.reserve(solverOutputStates.getSize());

    for (const State& solverOutputState : backwardPropagatedStates + forwardPropagatedStates)
```
Nit:

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -254,15 +250,25 @@ Array<State> Propagator::calculateStatesAt(const State& aState, const Array<Inst
...
+    {
+        outputStates.add(outputStateBuilder.expand(solverOutputState.inFrame(aState.getFrame()), aState));
```
vishwa2710, line 268:
```suggestion
        outputStates.add(outputStateBuilder.expand(solverOutputState.inFrame(aState.accessFrame()), aState));
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -400,6 +400,66 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Models_Propagator, Calcul
...
+        const VectorXd coordinates1 = outputStateGCRF.getCoordinates();
+        const VectorXd coordinates2 = outputStateGCRF2.getCoordinates();
+
+        // Check that the two vectors have the same size
+        EXPECT_EQ(coordinates1.size(), coordinates2.size());
+
+        // Check that each element of the two vectors is close
+        for (int i = 0; i < coordinates1.size(); ++i)
+        {
+            EXPECT_NEAR(coordinates1(i), coordinates2(i), 1e-5);
```
vishwa2710, line 429: I feel like it should be a lot closer than 1e-5, was this the best tolerance you got?

vishwa2710 (approved):
Looks good to me! Mostly nits, 1 question on tolerance.
Can you run `make format` once I feel like some of the formatting changed and it didn't change for me locally so it feels kinda strange O_o

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
kyle-cochran: I believe when we call `std::make_shared` ([here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp#L230)) with no arguments, it calls the default constructor. 


Check this out:
https://cpp.sh/
```
#include <iostream>
#include <memory>
int main () {
  struct thing {
   int i = 1;
   int j = 2;
  };
  std::shared_ptr<thing> foo = std::make_shared<thing> ();
std::cout << "*foo: " << foo->i << '\n';
std::cout << "*foo: " << foo->j << '\n';
  return 0;
}
```

vishwa2710: Yeah good point it'll just be empty. So when we add subsets it'll be shared.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -228,6 +230,7 @@ class Propagator
 
    private:
     Shared<CoordinatesBroker> coordinatesBrokerSPtr_ = std::make_shared<CoordinatesBroker>();
+    StateBuilder stateBuilder_;
```
phc1990, line 233: I love the fact that we can use `StateBuilder` now.

My only comment is whether it's probably best to just create the integration `StateBuilder` whenever we need, rather than having it as a property of the `Propagator`.

So far it is used 3 times:
- in `calculateStateAt`
- in `calculateStateToCondition`
- in `caclulateStatesAt`

I'd say than rather than just creating it as a property. We just initialise at those 3 instances. 

```cpp
const StateBuilder integrationStateBuilder = StateBuilder(Propagator::IntegrationFrameSPTr, coordinatesBrokerSPtr_);
```

If we don't want to have this instantiation repeated 3 times, we can have a private fun that generates it and then we just call it whenever we need:

```cpp
const StateBuilder integrationStateBuilder = this->generateIntegrationStateBuilder();
```

**Why?**

Because at the moment there is some logic to keep the `Propagator` in "synch" with the dynamics and other external elements. For instance, whenever we clear dynamics:

```cpp
void Propagator::clearDynamics()
{
    this->dynamicsContexts_.clear();
    this->coordinatesBrokerSPtr_ = std::make_shared<CoordinatesBroker>();
}
```

We would probably need to take care here of updating the `stateBuilder_` so that we make sure it points to the proper coordinates broker:

```Cpp
void Propagator::clearDynamics()
{
    this->dynamicsContexts_.clear();
    this->coordinatesBrokerSPtr_ = std::make_shared<CoordinatesBroker>();

      // This line below;
    this->stateBuilder_ = {Propagator::IntegrationFrameSPtr, this->coordinatesBrokerSPtr_};}
```

Whilst **this is still ok for now**, as the "keeping-in-synch" logic is not very complex. **It could get cumbersome in the future** (for example, we now no longer have a hardcode integration frame, but a variable one, or another function that causes the coordinate broker to change it's pointer), so I think it's safer to just the integration stateBuilder on a needly basis:
- We ensure the `stateBuilder` is always "in-synch"
- One less thing to worry about in the functions that keep stuff "in-synch" (e.g. `clearDynamics`)
- Creating a `stateBuilder` is not expensive

...and you wouldn't have to modify/update your tests, as this is all happening behind the scenes : )

phc1990 (commented):
Looks great : )
Left a comment on `StateBuilder` creation safety.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
vishwa2710: Since the coordinates broker is a shared pointer. And the statebuilder has a shared pointer of the broker, if we update the broker anywhere, it gets updated in the state builder as well. However you're right in the fact that it's used in a pretty trivial way right now so there isn't a strong case for having it be a member variable. I'm fine with both approaches as they achieve the same outcome with about the same performance and readibility.

kyle-cochran: That's a great point Pau. I waffled between the two approaches and actually had it the other way at first. 

I think you're right that it's safer to just create it on-demand for now. It can easily become a property in the future if it makes sense to do that.

vishwa2710: This is a snippet that validates that updating the broker would update it within the StateBuilder as well as they are shared pointers:
```
// Online C++ compiler to run C++ program online
#include <iostream>
#include <memory>

struct Cat
{
    std::string name;
    
    Cat(std::string aName): name(aName) {};
};

struct Zoo
{
    std::shared_ptr<Cat> cat1;
    
    Zoo(const std::shared_ptr<Cat>& aCat): cat1(aCat) {};
};

int main() {
    // Write C++ code here
    std::shared_ptr<Cat> toto = std::make_shared<Cat>(Cat("toto"));
    
    std::cout << "toto name: " << toto->name << std::endl;
    Zoo zoo(toto);
    std::cout << "cat1 name: " << zoo.cat1->name << std::endl;
    
    toto->name = "not toto";
    std::cout << "toto name: " << toto->name << std::endl;
    std::cout << "cat1 name: " << zoo.cat1->name << std::endl;

    return 0;
}
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
kyle-cochran: Nah, this isn't really testing to any meaningful tolerance. The point here was just to ensure that the transformation is actually happening. If it weren't, I'd expect errors on the order of meters or kilometers. I can add a comment to that effect for clarity

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -143,18 +144,19 @@ State Propagator::calculateStateAt(const State& aState, const Instant& anInstant
         throw ostk::core::error::runtime::Undefined("Propagator");
     }
 
-    const State state = {
-        aState.accessInstant(),
-        aState.inFrame(Propagator::IntegrationFrameSPtr).extractCoordinates(this->coordinatesBrokerSPtr_->getSubsets()),
-        Propagator::IntegrationFrameSPtr,
-        this->coordinatesBrokerSPtr_,
-    };
+    const State solverInputState = this->stateBuilder_.reduce(aState.inFrame(Propagator::IntegrationFrameSPtr));
```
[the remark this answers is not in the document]
kyle-cochran: Yeah instant value 😁

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
kyle-cochran: Thanks for the snippet Vishwa 🙏 . It does confirm the intuition. I've still just made them local properties for now, since it seems that we all agree it won't be too impactful and it's easy to go the other way in the future if needed.

vishwa2710: For sure yeah, not advocating in either direction. I just wanted to confirm that's how shared pointers worked lol :P

phc1990 (approved):
LGTM, just blocked by pipeline atm.

PR #250, opened 2023-10-16
feat: add metadata for segment and sequence solutions

vishwa2710:


bindings/python/test/trajectory/test_segment.py
```diff
@@ -174,5 +192,18 @@ def test_solve(
...
-        assert solution.condition_is_satisfied
+        assert solution.condition_is_satisfied is True
+        assert solution.segment_type is not None
+
+        assert solution.access_start_instant() is not None
+        assert solution.access_end_instant() is not None
+
+        assert solution.get_initial_mass() is not None
+        assert solution.get_final_mass() is not None
+
```
kyle-cochran, line 207:
```suggestion

assert solution.get_propagation_duration() is not None
```
missing or nah?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -11,20 +14,86 @@ namespace trajectory
...
+      conditionIsSatisfied(aConditionIsSatisfied),
+      segmentType(aSegmentType)
+{
+}
+
+const Instant& Segment::Solution::accessStartInstant() const
+{
+    if (states.isEmpty())
+    {
+        throw ostk::core::error::RuntimeError("States are empty.");
```
kyle-cochran, line 41:
```suggestion
        throw ostk::core::error::RuntimeError("No solution available.");
```

Is this condition relevant when the Segment hasn't been solved yet?

If so, maybe an error that says something like "Segment not yet solved." would be a little more helpful.

Also, for this one specifically, is it possible for the startInstant to be available when there aren't states or nah?

kyle-cochran (approved):
Some nits. Sick dude!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: Yeah good point. Technically if we run solve it should atleast have 1 state even if the condition is met immediately.

vishwa2710 (commented):
Sick catches. Thanks man!

PR #253, opened 2023-10-17
fix: fix missing binding for Segment::Solution

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -64,6 +64,11 @@ Mass Segment::Solution::getInitialMass() const
     return Mass::Kilograms(states.accessFirst().extractCoordinates(CoordinatesSubset::Mass())[0]);
 }
 
+Duration Segment::Solution::getPropagationDuration() const
+{
+    return accessEndInstant() - accessStartInstant();
```
apaletta3, line 69:
```suggestion
    return this->accessEndInstant() - this->accessStartInstant();
```
To be more robust?

vishwa2710: I think we can do that across the board to be consistent later if we decide.

PR #254, opened 2023-10-17
fix: update README Eigen version and add {fmt} dependency

DhruvJ22:
Updated README to be consistent with the Eigen version stated in the development docker file and to add the explicit dependence on fmt.


README.md
```diff
@@ -225,7 +225,8 @@ Benchmarks are pushed to the [GitHub pages](https://open-space-collective.github
 | ----------- | --------- | ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------ |
 | Pybind11    | `2.10.1`  | BSD-3-Clause       | [github.com/pybind/pybind11](https://github.com/pybind/pybind11)                                                                           |
 | ordered-map | `0.6.0`   | MIT                | [github.com/Tessil/ordered-map](https://github.com/Tessil/ordered-map)                                                                     |
-| Eigen       | `3.3.7`   | MPL2               | [eigen.tuxfamily.org](http://eigen.tuxfamily.org/index.php)                                                                                |
+| {fmt}       | `5.2.0`   | BSD-2-Clause       | [github.com/fmtlib/fmt](https://github.com/fmtlib/fmt)                                                                                     |
```
lucas-bremond, line 228: Thanks for updating! Actually, `fmt` is already listed as an [OSTk Core dependency](https://github.com/open-space-collective/open-space-toolkit-core#dependencies) this is why it isn't duplicated here.

However, the Eigen version bump is valid, thanks for spotting that!

DhruvJ22: Thank you for your comment! 

(OTSK-physics)[https://github.com/open-space-collective/open-space-toolkit-physics#dependencies] and (OTSK-mathematics)[https://github.com/open-space-collective/open-space-toolkit-mathematics#dependencies] list core and {fmt} as a dependency, hence I suggested the change to be consistent with other repos.

lucas-bremond: Whoops, that's likely an oversight! Let's summon @Derollez and @robinpdm to see what they think.

robinpdm: Good point. If we want these lists to match the [Dockerfile](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/docker/development/Dockerfile#L9)s, then indeed we may want to add `{fmt}` here? Or should they rather match `[CMakeLists](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/CMakeLists.txt#L213)`?

PR #262, opened 2023-10-25
feat: Event Condition termination for Sequences

vishwa2710:


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Sequence.cpp
```diff
@@ -267,44 +267,52 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Sequence(pybind11::module
...
 
             .def(
                 "solve",
                 &Sequence::solve,
-                arg("state"),
                 R"doc(
                     Solve the sequence.
 
+                    This method will terminate early if the maximum propagation duration is exceeded.
+                    If an event condition is provided, it will terminate if the event condition is met.
```
kyle-cochran, line 301:
```suggestion
                    If an event condition is provided, it will terminate when the event condition is met.
```

mega super ultra nit

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -222,17 +222,30 @@ void Sequence::addManeuverSegment(const Shared<EventCondition>& anEventCondition
...
+
     for (Size i = 0; i < repetitionCount_; ++i)
     {
         for (const Segment& segment : segments_)
         {
+            // Maximum propagation check
+            if ((initialState.accessInstant() - aState.accessInstant()) > aMaximumPropagationDuration)
+            {
+                BOOST_LOG_TRIVIAL(warning)
+                    << "Terminating Sequence because maximum propagation duration is reached." << std::endl;
```
kyle-cochran, line 244:
```suggestion
                    << "Terminating Sequence because maximum propagation duration was reached in the preceding segment." << std::endl;
```

nit: just thinking about that when this happens, the last segment will likely have gone _past_ the maximum duration. Might be good to indicate somehow that this check is discretized by segments to avoid confusion.

In general, I could see this condition being a little confusing. For example, if someone is testing a low-thrust maneuver to get to some altitude that they may never reach, the intuitive thing to do is probably to set this `maximumPropagationDuration`, so that it doesn't go forever. However, I think this condition alone would not be sufficient. They would need a maximum duration condition on the maneuver segment as well (or a maximum fuel used or something). In that case, this condition would appear to do nothing.

Let me know if I'm misunderstanding here. Perhaps we could name the condition something slightly different so it's clear that it's only applied at discrete instants between segments.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -244,18 +257,33 @@ Sequence::Solution Sequence::solve(const State& aState) const
...
+
+                return {segmentSolutions, true};
+            }
+
             initialState = segmentSolution.states.accessLast();
         }
     }
 
-    return {segmentSolutions, true};
+    if (eventConditionIsSatisfied)
```
kyle-cochran, line 281: Are both this and line 270 required?

I might be reading the flow wrong, but this looks like:
- if you get to the end of the sequence and a sequence Event Condition _existed_ but was **not** satisfied, return true.

It looks like nothing ever sets a value to `eventConditionIsSatisfied`. I.e. maybe a name like `sequenceEventConditionExists` is more appropriate. (maybe I'm blind and missing it somewhere)

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Sequence.cpp
vishwa2710: So technically the condition could never be met. Therefore I put an `if`.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
vishwa2710: yeah I think this is the confusion. We have 2 sets of `maximumPropagationDurations`. There's one for a Segment, so that any individual Segment doesn't blast off into infinity.
And then there's the one for a Sequence. So that if you have like 20 segments, each with a max of 7 days, you might still not want to run the sequence for more than 30 days or something.

I can update the naming so that the Sequence constructor has the `maximumPropagationDuration` changed to `segmentPropagationDurationLimit` and then the one in this function can remain. Is that better? @kyle-cochran

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
vishwa2710: So in line 234 we set
`    bool eventConditionIsSatisfied = anEventConditionSPtr == nullptr;`
So basically, if the eventCondition is a nullptr, then automatically it's "satisfied" as there is no condition.
If there isn't one, then it's set as false, and appropriately updated.
I thought this was clever but it might make the code less readable, what do you think?

vishwa2710 (commented):
Sick comments, 🏓

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
kyle-cochran: I don't think I follow... I believe this bool is never updated based on whether the event condition is actually satisfied. It's only set based on whether the event condition exists. So like 281 is effectively:
`if (sequenceEventConditionExists){return {segmentSolutions, true};}`

_But_ if we've gotten to this point in the code, that means we're at the end of all of the segments. I.e. the sequence event condition was _not_ hit. So this condition is effectively "if there was a sequence event condition but we didn't hit it, return success".

I might totally be missing something...

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
kyle-cochran: Hmmm... I think for the the `maximumPropagationDuration` naming implies a condition on "continuous" duration rather than duration discretized by segments. What about calling the sequence one something like: `maximumSegmentsDuration`. Trying to imply something like: "the sum of all of the segments cannot exceed X"

?

kyle-cochran (approved):

PR #263, opened 2023-10-26
feat: event condition relative target

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -28,16 +29,20 @@ class EventCondition
...
+    /// @param                  [in] anEvaluator A function evaluating a state
     /// @param                  [in] aTarget A target value associated with the Real Event Condition
+    /// @param                  [in] targetIsRelative A boolean indicating whether or not the target is relative
 
-    EventCondition(const String& aName);
+    EventCondition(
+        const String& aName,
+        std::function<Real(const State&)> anEvaluator,
+        const Real& aTarget,
+        const bool& targetIsRelative = false
```
kyle-cochran, line 44:
```suggestion
        const TargetSource& targetFrom = TargetSource::AbsoluteValue
```

Re: conversation about making this a bool vs. enum. If there will likely only be two values, the bool an enum approaches are similar and the bool approach is easier to implement.

Looking at the interface, it's a little unclear IMO what relative means here or how to use it. E.g. what is it relative to? I think if I approached this with fresh eyes, I would assume that the way to implement an event condition that relative to a value of 600km altitude is to use a relative event condition. 

This can probably be clearer with some naming like `targetRelativeToSegmentStart`. Maybe something better than that... just something to make it clear that it's relative to the start of the segment rather that a value known a-priori.

I like the semantics of an enum here for that reason, even if we don't technically need it for 2 values. Seems a little easier to indicate that the condition will depend on the initial Segment state at runtime. Something like:
```
    enum TargetSource
    {
        AbsoluteValue,
        SegmentStart,
        SequenceStart
    };
```

The `SequenceStart` value is not _really_ needed because as you pointed out, we always know the Sequence start State at construction time. However, it could be nice UX to be able to target the Sequence start automatically without having to copy the initial State around. This is a minor thing IMO though

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -28,16 +29,20 @@ class EventCondition
...
     /// @param                  [in] aName A string representing the name of the Real Event Condition
-    /// @param                  [in] aCriterion An enum indicating the criterion used to determine if the Real Event
-    /// Condition is met
-    /// @param                  [in] anEvaluator A function evaluating a state and a time
+    /// @param                  [in] anEvaluator A function evaluating a state
     /// @param                  [in] aTarget A target value associated with the Real Event Condition
+    /// @param                  [in] targetIsRelative A boolean indicating whether or not the target is relative
 
-    EventCondition(const String& aName);
+    EventCondition(
```
kyle-cochran, line 40: Is this an interface breaking change? FWIW I still like the idea semantically of just making Target a monad (i.e. either real valued or relative valued) but agree not worth a major breaking change just for that 😛 

One (perhaps big) plus, though, is that we wouldn't have to propagate down an extra argument to all of the event condition implementation children.

include/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp
```diff
@@ -43,11 +43,19 @@ using ostk::astro::trajectory::State;
 class COECondition
 {
    public:
+    /// @brief                  Semi-Major Axis based constructor
+    ///
+    /// @param                  [in] aCriterion An enum indicating the criterion used to determine the Event Condition
+    /// @param                  [in] aFrameSPtr A frame in which the Element is to be computed
+    /// @param                  [in] aSemiMajorAxis A Semi-Major Axis
+    /// @param                  [in] aGravitationalParameter A gravitational parameter
+
```
kyle-cochran, line 52: nit: missing docstrings for these for the new parameter

include/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp
```diff
@@ -43,11 +43,19 @@ using ostk::astro::trajectory::State;
...
     static RealCondition SemiMajorAxis(
```
kyle-cochran, line 53: nit: I know it was a lot of work to propagate these, but do you think it's worth it?

To me, it seems like fine UX to leave all of these specific COE condition generators as absolute valued, and if someone wants a relative condition, they can make their own. 

This is really a nitpick though. To me, it seems a little cleaner if factory functions do exactly 1 thing rather than trying to make them fully featured. Mostly a personal preference thing here though, so curious if others chime in.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -233,6 +233,11 @@ Sequence::Solution Sequence::solve(const State& aState) const
     {
         for (const Segment& segment : segments_)
         {
+            if (segment.accessEventCondition()->targetIsRelative())
+            {
+                segment.accessEventCondition()->updateTarget(initialState);
```
kyle-cochran, line 238:
```suggestion
                segment.accessEventCondition()->setTargetState(initialState);
```

nit: Seeing it in context, maybe this naming? It seems like this is something that each segment will have to set once at the beginning, so this naming feels a little cleaner to me.

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
```diff
@@ -28,16 +29,20 @@ class EventCondition
...
+    ///                         EventCondition eventCondition = {aName, anEvaluator, aTarget, targetIsRelative};
     /// @endcode
     ///
     /// @param                  [in] aName A string representing the name of the Real Event Condition
-    /// @param                  [in] aCriterion An enum indicating the criterion used to determine if the Real Event
-    /// Condition is met
-    /// @param                  [in] anEvaluator A function evaluating a state and a time
+    /// @param                  [in] anEvaluator A function evaluating a state
     /// @param                  [in] aTarget A target value associated with the Real Event Condition
+    /// @param                  [in] targetIsRelative A boolean indicating whether or not the target is relative
```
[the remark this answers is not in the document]
vishwa2710: I don't want to specify the term Segment here, as it's not required to be used only in conjunction with an OSTk Segment, you could use it anywhere technically. For example in a Window Generator (Yet to be created) class.

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
vishwa2710: This is an interface breaking change for the EventCondition class, but none of the children, so technically things should still work, since the EventCondition is just an abstract class.
I do like the Monad idea I might play with that in this MR on another branch.

include/OpenSpaceToolkit/Astrodynamics/EventCondition.hpp
vishwa2710: Yeah +1 on Enum. I don't think it's worth sublcassing EventCondition into a RelativeEventCondition style. I'll play with the Enum and see what I get.

include/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp
vishwa2710: Yeah that's fair. I think it's a balance between readibility and maintainability.
Adding N RelativeSemiMajorAxis etc. is cumbersome and just more Tests, Bindings, etc. etc.
Therefore more to maintain. But it might be more readable/ better UX

Let me play with the target Enum idea and see how far I get.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
```diff
@@ -37,6 +40,113 @@ class PyEventCondition : public EventCondition
...
+            R"doc(
+                Event Condition Target type.
+            )doc"
+        )
+
+            .value("Absolute", EventCondition::Target::Type::Absolute, "Absolute")
+            .value(
+                "RelativeSegmentStart",
+                EventCondition::Target::Type::RelativeSegmentStart,
+                "Relative to the State at the end of the previous Segment."
```
kyle-cochran, line 66:
```suggestion
                "Relative to the State at the start of the current Segment."
```
nit: just because the enum is `RelativeSegmentStart` and (I think?) it's equivalent.

include/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp
```diff
@@ -43,122 +44,129 @@ using ostk::astro::trajectory::State;
...
 {
    public:
+    /// @brief                  Semi-Major Axis based constructor
+    ///
+    /// @param                  [in] aCriterion An enum indicating the criterion used to determine the Event Condition
+    /// @param                  [in] aFrameSPtr A frame in which the Element is to be computed
+    /// @param                  [in] aTarget A Target
+    /// @param                  [in] aGravitationalParameter A gravitational parameter
+
     static RealCondition SemiMajorAxis(
```
kyle-cochran, line 54: nit: Think it's worth keeping the original factory function that takes a Length as well to avoid breaking backwards compatibility here? Not sure where all these are used, but could do a similar thing as `AngularCondition` above. Your call

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -118,6 +118,11 @@ class Segment
 
     const Shared<EventCondition>& accessEventCondition() const;
```
kyle-cochran, line 119: Sorry, this is not part of the change. Noticed it when comparing to the other you added, but should this be:

```suggestion
    const Shared<const EventCondition>& accessEventCondition() const;
```
?

if we don't want the event conditions associated with the segment to be modified

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -10,8 +10,54 @@ namespace ostk
...
 
-EventCondition::EventCondition(const String& aName)
-    : name_(aName)
+EventCondition::Target::Target(const Real& aValue, const Type& aType)
+    : value_(aValue),
+      type_(aType)
+{
+}
+
+EventCondition::Target::Target(const Angle& anAngle, const Type& aType)
```
kyle-cochran, line 19: Man this is a kind-of annoying downside to using the OSTk native units

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -29,11 +75,42 @@ String EventCondition::getName() const
...
+EventCondition::Target EventCondition::getTarget() const
+{
+    return target_;
+}
+
+void EventCondition::updateTarget(const State& aState)
+{
+    if (target_.type_ == EventCondition::Target::Type::Absolute)
+    {
+        throw ostk::core::error::RuntimeError("Can only set state for 'Relative' conditions.");
```
kyle-cochran, line 92:
```suggestion
        throw ostk::core::error::RuntimeError("Can only set state for 'Relative' Event Conditions.");
```

nit

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -29,11 +75,42 @@ String EventCondition::getName() const
...
+    }
+
+    if (!evaluator_)
+    {
+        throw ostk::core::error::runtime::Undefined("Evaluator.");
```
kyle-cochran, line 97:
```suggestion
        throw ostk::core::error::runtime::Undefined("Evaluator");
```

nit, because this will print as `Evaluator. is undefined`

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -29,11 +75,42 @@ String EventCondition::getName() const
...
+}
+
 void EventCondition::print(std::ostream& anOutputStream, bool displayDecorator) const
 {
     displayDecorator ? ostk::core::utils::Print::Header(anOutputStream, "Event Condition") : void();
 
     ostk::core::utils::Print::Line(anOutputStream) << "Name:" << getName();
+    // TBM: Use enum to string from Target::Type
+    ostk::core::utils::Print::Line(anOutputStream
+    ) << String::Format("Target {}:", ((target_.type_ == EventCondition::Target::Type::Absolute) ? " (Relative)" : ""))
```
kyle-cochran, line 110: Quick double check, hard to visualize what this will look like. Is this line necessary with line 113 as well? Also is this condition backwards? (i.e. won't it print `(Relative)` when target type is absolute?)

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
```diff
@@ -29,11 +75,42 @@ String EventCondition::getName() const
...
+      << getTarget().value_;
+    ostk::core::utils::Print::Line(anOutputStream)
+        << "Target Type:" << (target_.type_ == EventCondition::Target::Type::Absolute ? "Relative" : "Absolute");
```
kyle-cochran, line 113:
```suggestion
        << "Target Type: " << (target_.type_ == EventCondition::Target::Type::Absolute ? "Absolute" : "Relative");
```

Maybe I'm missing something. Hard to visualize.

kyle-cochran (commented):
1/2 of review. All nits

src/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.cpp
```diff
@@ -36,19 +47,14 @@ AngularCondition::Criterion AngularCondition::getCriterion() const
...
 
-std::function<Real(const State&)> AngularCondition::getEvaluator() const
-{
-    return evaluator_;
-}
-
 Angle AngularCondition::getTargetAngle() const
 {
-    if (!target_.isDefined())
+    if (!target_.value_.isDefined())
```
kyle-cochran, line 52: nit: if `value_` has an underscore, I would usually expect it to be private to the Target class and not accessed from outside sources. Worth making a `getValue()` function?

src/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.cpp
```diff
@@ -67,7 +73,7 @@ void AngularCondition::print(std::ostream& anOutputStream, bool displayDecorator
 
     EventCondition::print(anOutputStream, false);
     ostk::core::utils::Print::Line(anOutputStream) << "Criterion: " << StringFromCriterion(getCriterion());
-    if (target_.isDefined())
+    if (target_.value_.isDefined())
```
kyle-cochran, line 76: nit: could possibly also be worth it to make a `Target::isDefined()` that checks the value and relative state properly based on the target type.

src/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.cpp
```diff
@@ -170,6 +171,7 @@ std::function<Real(const State&)> COECondition::GenerateEvaluator(
             case COE::Element::Raan:
                 return coe.getRaan().inRadians(0.0, Real::TwoPi());
             case COE::Element::TrueAnomaly:
+                // std::cout << "TrueAnomaly: " << coe.getTrueAnomaly().inDegrees() << std::endl;
```
kyle-cochran, line 174:
```suggestion
```

nit

src/OpenSpaceToolkit/Astrodynamics/EventCondition/RealCondition.cpp
```diff
@@ -49,14 +53,13 @@ void RealCondition::print(std::ostream& anOutputStream, bool displayDecorator) c
...
     ostk::core::utils::Print::Line(anOutputStream) << "Criterion: " << StringFromCriterion(getCriterion());
-    ostk::core::utils::Print::Line(anOutputStream) << "Target: " << getTarget();
 
     displayDecorator ? ostk::core::utils::Print::Footer(anOutputStream) : void();
 }
 
 Real RealCondition::evaluate(const State& aState) const
 {
-    return this->evaluator_(aState) - target_;
+    return this->evaluator_(aState) - (target_.value_ + target_.valueOffset_);
```
kyle-cochran, line 62:
```suggestion
    return this->evaluator_(aState) - (target_.getValue() + target_.getValueOffset());
```

nit: similar comment here to the one in `AngularCondition.cpp`. I think our convention is to have getters/accessors for underscored class variables.

src/OpenSpaceToolkit/Astrodynamics/EventCondition/RealCondition.cpp
```diff
@@ -90,6 +93,19 @@ String RealCondition::StringFromCriterion(const Criterion& aCriterion)
...
+RealCondition RealCondition::DurationCondition(const Criterion& aCriterion, const Duration& aDuration)
+{
+    return {
+        "Duration",
+        aCriterion,
+        [](const State& aState) -> Real
+        {
+            return (aState.accessInstant() - Instant::J2000()).inSeconds();
+        },
+        {aDuration.inSeconds(), RealCondition::Target::Type::RelativeSegmentStart}
```
kyle-cochran, line 105:
```suggestion
        {aDuration.inSeconds(), EventCondition::Target::Type::RelativeSegmentStart}
```

nit: seems a little clearer to me to use the original base class in which it's defined. Don't feel strongly about this one though

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -243,6 +243,15 @@ Sequence::Solution Sequence::solve(
     {
         for (const Segment& segment : segments_)
         {
+            if (segment.accessEventCondition()->getTarget().type_ == EventCondition::Target::Type::RelativeSegmentStart)
```
kyle-cochran, line 246:
```suggestion
            if (segment.accessEventCondition()->getTarget().getType() == EventCondition::Target::Type::RelativeSegmentStart)
```

nit: ditto here. Consider a getter/accessor for retrieving underscored variables

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -243,6 +243,15 @@ Sequence::Solution Sequence::solve(
...
+            {
+                segment.accessEventCondition()->updateTarget(initialState);
+            }
+            else if (segment.accessEventCondition()->getTarget().type_ == EventCondition::Target::Type::RelativeSequenceStart)
+            {
+                segment.accessEventCondition()->updateTarget(aState);
```
kyle-cochran, line 252: Since this seems like a pretty core part of the flow, can you add a test that covers it 🙏

kyle-cochran (commented):
2/2 of review. Looks awesome dude! More nits

include/OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp
vishwa2710: They're marginally used in some notebooks, I think it's easier to update it than support it for backwards compatability IMO

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
vishwa2710: Ah shit you're right, I don't think I need to add this method as the EventCondition is not const. I'll just delete my method.

src/OpenSpaceToolkit/Astrodynamics/EventCondition.cpp
vishwa2710: Yeah it's unfortunate, but atleast it's overloaded so there's that...

src/OpenSpaceToolkit/Astrodynamics/EventCondition/AngularCondition.cpp
vishwa2710: Excellent point.

vishwa2710 (commented):
Thanks for the sick feedback!

kyle-cochran (approved):

PR #264, opened 2023-11-01
feat: add finite difference solver

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Solvers/FiniteDifferenceSolver.hpp
```diff
@@ -0,0 +1,137 @@
...
+    {
+        Forward,
+        Backward,
+        Central
+    };
+
+    /// @brief                  Constructor
+    ///
+    /// @code
+    ///                         const Real stepPercentage = 1e-3;
```
kyle-cochran, line 49: Is this code example correct for this one? Looks like it takes a type instead

include/OpenSpaceToolkit/Astrodynamics/Solvers/FiniteDifferenceSolver.hpp
```diff
@@ -0,0 +1,137 @@
...
+    /// @param                  [in] anInstant An instant.
+    /// @param                  [in] getState Callable to generate a State at the requested Instant.
+    /// @param                  [in] aStepPercentage The step percentage to use for the perturbation. Defaults to 1e-3.
+    ///
+    /// @return                 The State Transition Matrix
+
+    MatrixXd computeStateTransitionMatrix(
+        const State& aState,
+        const Instant& anInstant,
+        std::function<State(const State&, const Instant&)> getState,
```
kyle-cochran, line 95:
```suggestion
        std::function<State(const State&, const Instant&)> propagateState,
```

nit: a bit pedantic, but something like `propagateState` (or `getStateAtInstant` or `propagateStateToInstant`) is a little clearer IMO and describes what the input lambda should do.

kyle-cochran (approved):
Couple nits. lgtm

include/OpenSpaceToolkit/Astrodynamics/Solvers/FiniteDifferenceSolver.hpp
vishwa2710: I've updated this to generateState, agreed that getState is a bit generic and used as getter.

PR #269, opened 2023-11-08
fix: fix relative logical conditions usage with Sequences

vishwa2710:


test/OpenSpaceToolkit/Astrodynamics/EventCondition.test.cpp
```diff
@@ -188,11 +179,12 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_EventCondition, UpdateTarget)
     };
 
     {
-        EXPECT_THROW(defaultCondition_.updateTarget(state), ostk::core::error::RuntimeError);
+        defaultCondition_.updateTarget(state);
+        EXPECT_FALSE(defaultCondition_.getTarget().valueOffset == 0.0);
```
kyle-cochran, line 183: I think this is one of the test failing. Should this be:
```suggestion
        EXPECT_TRUE(defaultCondition_.getTarget().valueOffset == 0.0);
```

since it's an absolute target?

kyle-cochran (approved):
LGTM after pipeline is good

test/OpenSpaceToolkit/Astrodynamics/EventCondition.test.cpp
vishwa2710: Yeah that's it. I changed it from `!= 0.0` but forgot to change the EXPECT_FALSE to true. Great catch buddy.

PR #273, opened 2023-11-14
feat: scale mass flow rate by thrust percentage

vishwa2710:
In the scenario that we return an acceleration that is < the thrust magnitude maximum we should scale the mass flow rate based on the effective thrust.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Thruster.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Thruster.cpp
@@ -95,11 +95,13 @@ VectorXd Thruster::computeContribution(
         anInstant, positionCoordinates, velocityCoordinates, maximumThrustAccelerationMagnitude, aFrameSPtr
     );
 
+    const Real effectiveThrustPercent = acceleration.norm() / maximumThrustAccelerationMagnitude;
+
     // Compute contribution
     VectorXd contribution(4);
     // TBI: Can be optimized to cache the SI value of mass flow rate as a Real
     contribution << acceleration[0], acceleration[1], acceleration[2],
-        -satelliteSystem_.accessPropulsionSystem().getMassFlowRate().getValue();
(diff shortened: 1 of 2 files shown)


Derollez:
When is this the case with the current guidance laws we support? 🤔

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Thruster.cpp
```diff
@@ -95,11 +95,13 @@ VectorXd Thruster::computeContribution(
         anInstant, positionCoordinates, velocityCoordinates, maximumThrustAccelerationMagnitude, aFrameSPtr
     );
 
+    const Real effectiveThrustPercent = acceleration.norm() / maximumThrustAccelerationMagnitude;
```
phc1990, line 98: is `calculateThrustAccelerationAt` not always returning the max thrust acceleration?

Derollez: So that's basically what I was wondering because:
- `ConstantThrust` Law does `const Vector3d acceleration_LOF = aThrustAcceleration * this->localOrbitalFrameDirection_.getValue();` and simply rotates the vector
- `QLaw` does `aThrustAcceleration * R_thetaRH_GCRF * thrustDirection` 
So I would expect both current laws to have effectiveThrustPercent to 1 🤔 But the change in tests seems to indicate there are currently cases where that efficiency is not 1 I guess

Useful in future cases maybe

Derollez: hum, unless this is for the current case where `QLaw` recommends not to burn but max thrust is still the default value, in which case, currently `effectiveThrustPercent` can be 1 or 0 basically (matches the test example)

Derollez: Checked, That's it I think!

Derollez (approved):

PR #275, opened 2023-11-18
fix: addSegment method on Sequence

Derollez:


vishwa2710 (approved):
Looks good to me! Merge depending on the decision made on slack.

PR #277, opened 2023-11-20
chore: cache mass flow rate

vishwa2710:


apaletta3 (changes requested):
The test that doesn't pass can be fixed on line 99 of `Thurster.test.cpp` by changing it to this
```
EXPECT_THROW(Thruster(SatelliteSystem::Undefined(), defaultGuidanceLaw_, defaultName_), ostk::core::error::runtime::Undefined);
```
However, I'm not sure that's exactly the behavior you want here since it fails during construction, before you even get to use the `.getSatelliteSystem()`. So you might have to change something in the `SatelliteSystem` class for it to behave correctly.

apaletta3 (approved):
solid fix!

PR #278, opened 2023-11-21
chore: align repo with plural/singular function args

vishwa2710:


Derollez (approved):
Thanks for aligning 🙏

PR #279, opened 2023-11-21
feat: add tests for QLaw targeting from SSO

vishwa2710:
I removed the Case E test (from the paper) and instead added these tests for various Element targeting combos. I've bumped up the thrust so that it doesn't take long to propagate to get there.


apaletta3 (approved):
ok I see, so this about checking that far from the convergence point, the Qlaw targeting sends you in the correct direction for all 4 of the five elements. What about AOP? Does that targeting not work yet? Or does it need to be combined with Ecc targeting?

In general looks good. Can't wait to start structuring these properly with our validation framework :D

Derollez (approved):

vishwa2710:
@apaletta3 I've left out Aop targeting for now as we won't really use that for upcoming missions, but it should work. I'll play with it in the future and add some appropriate tests when possible.

PR #282, opened 2023-11-22
chore: improve-makefile-and-dev-experience

apaletta3:
This MR aims to make the ostk dev experience better by streamlining the build/make process by improving the makefile:
- allow the `make debug-jupyter-notebook` target to build and cache the binaries for the locally modified source code everytime you run that command, instead of needing to spin up a development container and build/re-build before re-launching jupyter
- make clean now removes all build/lib/package/etc folders to actually clean things properly

The part that hasn't been done yet is to fix container users so that we aren't root inside the container, which causes issues when working inside and outside the container since the hosts' files are now changed owners. This might become a little difficult because CMake doesn't seem to like being run from a non-root user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ build-packages-python-standalone: ## Build Python packages (standalone)
 		--workdir=/app/build \
 		$(docker_development_image_repository):$(docker_image_version) \
 		/bin/bash -c "cmake -DBUILD_UNIT_TESTS=OFF -DBUILD_PYTHON_BINDINGS=ON .. \
-		&& $(MAKE) -j 4 \
+		&& $(MAKE) -j $(shell nproc) \
 		&& mkdir -p /app/packages/python \
 		&& mv /app/build/bindings/python/dist/*.whl /app/packages/python"
 
@@ -237,7 +237,7 @@ start-development-no-link: build-development-image ## Start development environm
 		-it \
 		--rm \
 		--privileged \
-		--volume="$(CURDIR):/app:delegated" \
+		--volume="$(CURDIR):/app" \
 		--workdir=/app/build \
 		$(docker_development_image_repository):$(docker_image_version) \
 		/bin/bash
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,7 +223,7 @@ build-packages-python-standalone: ## Build Python packages (standalone)
 		--workdir=/app/build \
 		$(docker_development_image_repository):$(docker_image_version) \
 		/bin/bash -c "cmake -DBUILD_UNIT_TESTS=OFF -DBUILD_PYTHON_BINDINGS=ON .. \
-		&& $(MAKE) -j 4 \
+		&& $(MAKE) -j $(shell nproc) \
```
vishwa2710, line 226: This was hardcoded to 4 as it was causing issues in the pipeline by causing it crash if all processors were busy I believe.

Makefile
```diff
@@ -289,7 +289,21 @@ start-jupyter-notebook: build-release-image-jupyter ## Start Jupyter Notebook en
...
+
+	docker run \
+		-it \
+		--rm \
+		--volume="$(CURDIR):/app:delegated" \
+		--workdir=/app/build \
+		$(docker_development_image_repository):$(docker_image_version) \
+		/bin/bash -c "cmake -DBUILD_UNIT_TESTS=OFF -DBUILD_BENCHMARK=OFF -DBUILD_PYTHON_BINDINGS=ON -DPYTHON_SEARCH_VERSIONS="$(jupyter_python_version)" .. \
+		&& $(MAKE) -j $(shell nproc) \
+		&& cp /app/build/bindings/python/dist/*.whl /app/packages/python"
```
vishwa2710, line 306: Honestly this is a bit "worse" for the dev experience because even if I have built the binary this will rebuild it. SO I have to wait for it to build which takes a while.
If you want to include this I would make it a flag as this also prevents you from using a custom binary. For example if I wanted to build it in debug mode or compile it with very specific flags this will not allow me to do so.

Makefile
apaletta3: Ah I see, I'll revert

Makefile
apaletta3: Yup good point, changed it, let me know what you think.

Makefile
Derollez: > This was hardcoded to 4 as it was causing issues in the pipeline by causing it crash if all processors were busy I believe

Yes, let's keep the current numbers on nproc for now until we understand better the behavior 🙏 (I made that mistake before)

Makefile
```diff
@@ -289,7 +287,27 @@ start-jupyter-notebook: build-release-image-jupyter ## Start Jupyter Notebook en
...
+		&& $(MAKE) -j $(shell nproc) \
+		&& mkdir -p /app/packages/python \
+		&& rm -rf /app/packages/python/* \
+		&& cp /app/build/bindings/python/dist/*$(extract_python_package_version)*.whl /app/packages/python"
+
+	@ $(MAKE) debug-jupyter-standalone
+
+.PHONY: debug-jupyter
+
+debug-jupyter-standalone: build-release-image-jupyter ## Debug jupyter notebook using the ostk-astro package built from current source code
```
vishwa2710, line 310:
```suggestion
debug-jupyter-rebuild: build-release-image-jupyter ## Debug jupyter notebook using the ostk-astro package built from current source code
```
I would reserve the `standalone` naming for that which is to be used specifically by the CI.

Makefile
```diff
@@ -298,17 +316,16 @@ debug-jupyter-notebook: build-release-image-jupyter ## Debug jupyter notebook us
...
 		$(docker_release_image_jupyter_repository):$(docker_image_version) \
-		bash -c "chown -R jovyan:users /home/jovyan ; python$(jupyter_python_version) -m pip install /opt/conda/lib/python$(jupyter_python_version)/site-packages/ostk/$(project_name)/ --force-reinstall ; start-notebook.sh --ServerApp.token=''"
+		bash -c "chown -R jovyan:users /home/jovyan ; python$(jupyter_python_version) -m pip install /home/jovyan/.packages/*.whl --force-reinstall ; start-notebook.sh --ServerApp.token=''"
 
-	bash -c "sudo chown -R $(shell id -u):$(shell id -g) $(CURDIR)/bindings/python/docs"
-	bash -c "sudo chown -R $(shell id -u):$(shell id -g) $(CURDIR)/tutorials/python/notebooks"
+	@ sudo chown -R $(shell id -u):$(shell id -g) $(CURDIR)
 
-.PHONY: debug-jupyter-notebook
+.PHONY: debug-jupyter-standalone
```
vishwa2710, line 328:
```suggestion
.PHONY: debug-jupyter-rebuild
```

vishwa2710 (approved):

PR #283, opened 2023-11-26
build: use explicit namespacing

kyle-cochran:
Math: obj->object, geom->geometry, trf::rot->transformations::rotations
Core: fs->filesystem
Physics: env::object::celest->environment::object::celestial

Draft pending release of OSTk Math with extra namespacing changes.


apaletta3 (approved):
LGTM

PR #284, opened 2023-11-27
refactor: split sequence solve methods

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -226,42 +219,22 @@ void Sequence::addManeuverSegment(const Shared<EventCondition>& anEventCondition
...
 }
 
-Sequence::Solution Sequence::solve(
-    const State& aState,
-    const Duration& sequencePropagationDurationLimit,
-    const Shared<EventCondition>& anEventConditionSPtr
-) const
+Sequence::Solution Sequence::solve(const State& aState, const Size& aRepetitionCount) const
 {
     Array<Segment::Solution> segmentSolutions;
```
phc1990, line 224: maybe check that `aRepetitionCount` > 0?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -147,19 +146,21 @@ class Sequence
 
     void addManeuverSegment(const Shared<EventCondition>& anEventConditionSPtr, const Shared<Thruster>& aThruster);
 
+    /// @brief                  Solve the sequence given an initial state, for a number of reptitions.
+    ///
+    /// @param                  [in] aState Initial state for the sequence.
+    /// @param                  [in] aRepetitionCount Number of repetitions. Defaults to 1.
```
phc1990, line 152: Maybe add `Default to 1 (1 meaning that the sequence is run once)` in case somebody wonders "does 1 mean that it repeats itself once and therefore the total number of runs id 2?"

PR #286, opened 2023-11-28
build: fix jupyter notebook make commands

vishwa2710:


Makefile
```diff
@@ -298,16 +300,13 @@ debug-jupyter: build-development-image ## Debug jupyter notebook using the ostk-
 		--workdir=/app/build \
 		$(docker_development_image_repository):$(docker_image_version) \
 		/bin/bash -c "cmake -DBUILD_UNIT_TESTS=OFF -DBUILD_BENCHMARK=OFF -DBUILD_PYTHON_BINDINGS=ON -DPYTHON_SEARCH_VERSIONS="$(jupyter_python_version)" .. \
-		&& $(MAKE) -j $(shell nproc) \
-		&& mkdir -p /app/packages/python \
-		&& rm -rf /app/packages/python/* \
-		&& cp /app/build/bindings/python/dist/*$(extract_python_package_version)*.whl /app/packages/python"
+		&& $(MAKE) -j $(shell nproc)
```
apaletta3, line 303:
```suggestion
		&& $(MAKE) -j $(shell nproc)"
```

apaletta3 (approved):
Thanks for catching these, I just accepted the suggestions on my last MR and merged without checking them locally again.

What you did is definitely better, do you know what the difference between installing a python package from the `astrodynamics` folder is vs directly from the .whl as I was doing it before?

vishwa2710:
The issue with the way you were doing it before is that it mandated having run the `debug-jupyter-rebuild` (as it was creating directories) before running the `debug-jupyter` command. This sort of defeated the point of separating the two commands.

PR #291, opened 2023-12-04
feat: add dynamics contribution getters to segment solution

Derollez:
- `getDynamicsContribution` method allowing to compute the contribution of a `Dynamics` attached to the Segment Solution for a `Frame`, `Dynamics` and optional subset of `CoordinatesSubset` attached to Dynamics write coordinates Subset. 
- `getAllDynamicsContributions` method leveraging the first method and returning on Map of contributions Matrices for all dynamics of the Segment Solution. 
- `getAccelerationDynamicsContribution` method leveraging the first method and returning the acceleration contribution only of a Dynamics.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
@@ -203,13 +203,13 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
                     frame (Frame): The frame.
 
                 Returns:
-                    MatrixXd: The matrix of dynamics contributions to acceleration.
+                    np.ndarray: The matrix of dynamics contributions to acceleration.
 
             )doc"
         )
         .def(
-            "get_dynamics_contributions",
-            &Segment::Solution::getDynamicsContributions,
+            "get_all_dynamics_contributions",
+            &Segment::Solution::getAllDynamicsContributions,
             arg("frame"),
             R"doc(
                 Compute the contributions of all segment's dynamics in the provided frame for all states assocated with the segment.
@@ -218,7 +218,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
                     frame (Frame): The frame.
 
                 Returns:
-                    list[MatrixXd]: The list of matrices with individual dynamics contributions.
+                    dict[Dynamics, np.ndarray]: The list of matrices with individual dynamics contributions.
 
             )doc"
         )
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
@@ -171,13 +171,34 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
             )doc"
         )
         .def(
-            "compute_accelerations",
-            &Segment::Solution::computeAccelerations,
+            "get_dynamics_contribution",
+            &Segment::Solution::getDynamicsContribution,
+            arg("dynamics"),
+            arg("frame"),
             R"doc(
-                Compute the accelerations contributions of the involved dynamics.
+                Compute the contribution of the provided dynamics in the provided frame for all states associated with the segment.
+
+                Args:
+                    dynamics (Dynamics): The dynamics.
+                    frame (Frame): The frame.
 
                 Returns:
-                    list[MatrixXd]: The list of acceleration matrices.
+                    MatrixXd: The matrix of dynamics contributions.
+
+            )doc"
+        )
+        .def(
+            "get_dynamics_contributions",
+            &Segment::Solution::getDynamicsContributions,
+            arg("frame"),
+            R"doc(
+                Compute the contributions of all segment's dynamics in the provided frame for all states assocated with the segment.
+
+                Args:
+                    frame (Frame): The frame.
+
+                Returns:
+                    list[MatrixXd]: The list of matrices with individual dynamics contributions.
 
             )doc"
         )
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
@@ -170,6 +170,17 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
 
             )doc"
         )
+        .def(
+            "compute_accelerations",
+            &Segment::Solution::computeAccelerations,
+            R"doc(
+                Compute the accelerations contributions of the involved dynamics.
+
+                Returns:
+                    list[MatrixXd]: The list of acceleration matrices.
+
+            )doc"
+        )
 
         ;
 
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
@@ -13,6 +13,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
     using ostk::physics::time::Duration;
 
     using ostk::astro::trajectory::state::NumericalSolver;
+    using ostk::astro::trajectory::state::CoordinatesSubset;
     using ostk::astro::trajectory::Segment;
     using ostk::astro::Dynamics;
 
@@ -175,15 +176,34 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
             &Segment::Solution::getDynamicsContribution,
             arg("dynamics"),
             arg("frame"),
+            arg("coordinates_subsets") = Array<Shared<const CoordinatesSubset>>::Empty(),
             R"doc(
                 Compute the contribution of the provided dynamics in the provided frame for all states associated with the segment.
 
                 Args:
                     dynamics (Dynamics): The dynamics.
                     frame (Frame): The frame.
+                    coordinates_subsets (list[CoordinatesSubset], optional): A subset of the dynamics writing coordinates subsets to consider.
 
                 Returns:
-                    MatrixXd: The matrix of dynamics contributions.
+                    MatrixXd: The matrix of dynamics contributions for the selected coordinates subsets of the dynamics.
+
+            )doc"
+        )
+        .def(
+            "get_dynamics_acceleration_contribution",
+            &Segment::Solution::getDynamicsAccelerationContribution,
+            arg("dynamics"),
+            arg("frame"),
+            R"doc(
+                Compute the contribution of the provided dynamics to the acceleration in the provided frame for all states associated with the segment.
+
+                Args:
+                    dynamics (Dynamics): The dynamics.
+                    frame (Frame): The frame.
+
+                Returns:
+                    MatrixXd: The matrix of dynamics contributions to acceleration.
 
             )doc"
         )
--- a/bindings/python/test/trajectory/test_segment.py
+++ b/bindings/python/test/trajectory/test_segment.py
@@ -243,7 +243,7 @@ class TestSegment:
         assert second_dynamics_contribution.shape == (len(solution.states), 3)
         assert third_dynamics_contribution.shape == (len(solution.states), 4)
 
-        all_contributions = solution.get_dynamics_contributions(state_frame)
+        all_contributions = solution.get_all_dynamics_contributions(state_frame)
         assert len(all_contributions) == len(solution.dynamics)
 
         assert all_contributions is not None
--- a/bindings/python/test/trajectory/test_segment.py
+++ b/bindings/python/test/trajectory/test_segment.py
@@ -1,6 +1,7 @@
 # Apache License 2.0
 
 import pytest
+import numpy as np
 
 from ostk.physics import Environment
 from ostk.physics.time import Instant
@@ -221,4 +222,50 @@ class TestSegment:
         assert solution.compute_delta_mass() is not None
         assert solution.compute_delta_v(1500.0) is not None
 
-        assert solution.compute_accelerations() is not None
+        state_frame: Frame = state.get_frame()
+
+        first_dynamics_contribution = solution.get_dynamics_contribution(
+            solution.dynamics[0], state_frame
+        )
+
+        second_dynamics_contribution = solution.get_dynamics_contribution(
+            solution.dynamics[1], state_frame
+        )
+
+        third_dynamics_contribution = solution.get_dynamics_contribution(
+            solution.dynamics[2], state_frame
+        )
+
+        assert first_dynamics_contribution is not None
+        assert second_dynamics_contribution is not None
+        assert third_dynamics_contribution is not None
+        assert first_dynamics_contribution.shape == (len(solution.states), 3)
+        assert second_dynamics_contribution.shape == (len(solution.states), 3)
+        assert third_dynamics_contribution.shape == (len(solution.states), 4)
+
+        all_contributions = solution.get_dynamics_contributions(state_frame)
+        assert len(all_contributions) == len(solution.dynamics)
+
+        assert all_contributions is not None
+        assert isinstance(all_contributions, dict)
+        assert np.array_equal(
+            all_contributions[solution.dynamics[0]], first_dynamics_contribution
+        )
+        assert np.array_equal(
+            all_contributions[solution.dynamics[1]], second_dynamics_contribution
+        )
+        assert np.array_equal(
+            all_contributions[solution.dynamics[2]], third_dynamics_contribution
+        )
+
+        acceleration_contribution = solution.get_dynamics_acceleration_contribution(
+            solution.dynamics[1], state_frame
+        )
+        assert acceleration_contribution is not None
+        assert acceleration_contribution.shape == (len(solution.states), 3)
+        assert isinstance(acceleration_contribution, np.ndarray)
+
+        with pytest.raises(Exception):
+            solution.get_dynamics_acceleration_contribution(
+                third_dynamics_contribution, state_frame
+            )
--- a/bindings/python/test/trajectory/test_segment.py
+++ b/bindings/python/test/trajectory/test_segment.py
@@ -221,9 +221,4 @@ class TestSegment:
         assert solution.compute_delta_mass() is not None
         assert solution.compute_delta_v(1500.0) is not None
 
-        accelerations = solution.compute_accelerations()
-
-        print(accelerations[0])
-        print(accelerations[1])
-        print(accelerations[2])
-        # assert solution.compute_accelerations() is not None
+        assert solution.compute_accelerations() is not None
--- a/bindings/python/test/trajectory/test_segment.py
+++ b/bindings/python/test/trajectory/test_segment.py
@@ -220,3 +220,10 @@ class TestSegment:
 
         assert solution.compute_delta_mass() is not None
         assert solution.compute_delta_v(1500.0) is not None
+
+        accelerations = solution.compute_accelerations()
+
+        print(accelerations[0])
+        print(accelerations[1])
+        print(accelerations[2])
+        # assert solution.compute_accelerations() is not None
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
@@ -7,6 +7,8 @@
 #include <OpenSpaceToolkit/Core/Types/Shared.hpp>
 #include <OpenSpaceToolkit/Core/Types/String.hpp>
 
+#include <OpenSpaceToolkit/Mathematics/Objects/Matrix.hpp>
+
 #include <OpenSpaceToolkit/Physics/Time/Duration.hpp>
 #include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
 
@@ -26,6 +28,8 @@ namespace trajectory
 using ostk::core::ctnr::Array;
 using ostk::core::types::Shared;
 using ostk::core::types::String;
+
+using ostk::math::object::MatrixXd;
 
 using ostk::physics::time::Instant;
 using ostk::physics::time::Duration;
@@ -74,6 +78,8 @@ class Segment
 
         Real computeDeltaV(const Real& aSpecificImpulse) const;
         Mass computeDeltaMass() const;
+
+        Array<MatrixXd> computeAccelerations() const;
 
         void print(std::ostream& anOutputStream, bool displayDecorator = true) const;
 
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
@@ -41,7 +41,7 @@ using ostk::astro::Dynamics;
 using ostk::astro::dynamics::Thruster;
 using ostk::astro::EventCondition;
 
-/// @brief                                  Represent a propagation segment for astrodynamics purposes
+/// @brief                      Represent a propagation segment for astrodynamics purposes
 
 class Segment
 {
@@ -79,7 +79,10 @@ class Segment
         Real computeDeltaV(const Real& aSpecificImpulse) const;
         Mass computeDeltaMass() const;
 
-        Array<MatrixXd> computeAccelerations() const;
+        MatrixXd getDynamicsContribution(
+            const Shared<const Dynamics>& aDynamicsSPtr, const Shared<const Frame>& aFrameSPtr
+        ) const;
+        Array<MatrixXd> getDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const;
 
         void print(std::ostream& anOutputStream, bool displayDecorator = true) const;
 
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
@@ -57,6 +57,15 @@ class Segment
     struct Solution
     {
        public:
+        /// @brief              Constructor
+        ///
+        /// @param              [in] aName Name of the segment
+        /// @param              [in] aDynamicsArray Array of dynamics
+        /// @param              [in] aStates Array of states for the segment
+        /// @param              [in] aConditionIsSatisfied True if the event condition is satisfied
+        /// @param              [in] aSegmentType Type of segment
+        /// @return             An instance of Solution
+
         Solution(
             const String& aName,
             const Array<Shared<Dynamics>>& aDynamicsArray,
@@ -65,36 +74,94 @@ class Segment
             const Segment::Type& aSegmentType
         );
 
-        String name;                       /// Name of the segment.
-        Array<Shared<Dynamics>> dynamics;  /// List of dynamics used.
-        Array<State> states;               /// Array of states for the segment.
-        bool conditionIsSatisfied;         /// True if the event condition is satisfied.
-        Segment::Type segmentType;         /// Type of segment.
+        /// @brief              Access Start Instant
+        /// @return             Start Instant
 
         const Instant& accessStartInstant() const;
+
+        /// @brief              Access end instant
+        /// @return             End Instant
+
         const Instant& accessEndInstant() const;
 
+        /// @brief              Get initial mass
+        /// @return             Initial mass
+
         Mass getInitialMass() const;
+
+        /// @brief              Get final mass
+        /// @return             Final mass
+
         Mass getFinalMass() const;
+
+        /// @brief              Get propagation duration
+        /// @return             Propagation duration
+
         Duration getPropagationDuration() const;
 
+        /// @brief              Compute delta V
+        ///
+        /// @param              [in] aSpecificImpulse Specific impulse
+        /// @return             Delta V
+
         Real computeDeltaV(const Real& aSpecificImpulse) const;
+
+        /// @brief              Compute delta mass
+        /// @return             Delta mass
+
         Mass computeDeltaMass() const;
 
-        Map<Shared<Dynamics>, MatrixXd> getDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const;
+        /// @brief              Get dynamics contribution
+        ///
+        /// @param              [in] aDynamicsSPtr Dynamics
+        /// @param              [in] aFrameSPtr Frame
+        /// @param              [in] aCoordinatesSubsetSPtrArray Array of coordinates subsets
+        /// @return             Dynamics contribution
+
         MatrixXd getDynamicsContribution(
             const Shared<Dynamics>& aDynamicsSPtr,
             const Shared<const Frame>& aFrameSPtr,
             const Array<Shared<const CoordinatesSubset>>& aCoordinatesSubsetSPtrArray =
                 Array<Shared<const CoordinatesSubset>>::Empty()
         ) const;
+
+        /// @brief              Get dynamics acceleration contribution
+        ///
+        /// @param              [in] aDynamicsSPtr Dynamics
+        /// @param              [in] aFrameSPtr Frame
+        /// @return             Dynamics acceleration contribution
+
         MatrixXd getDynamicsAccelerationContribution(
             const Shared<Dynamics>& aDynamicsSPtr, const Shared<const Frame>& aFrameSPtr
         ) const;
 
+        /// @brief              Get all segment dynamics contributions
+        ///
+        /// @param              [in] aFrameSPtr Frame
+        /// @return             All segment dynamics contributions
+
+        Map<Shared<Dynamics>, MatrixXd> getAllDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const;
+
+        /// @brief              Print the segment solution
+        ///
+        /// @param              [in] anOutputStream An output stream
+        /// @param              [in] (optional) displayDecorators If true, display decorators
+
         void print(std::ostream& anOutputStream, bool displayDecorator = true) const;
 
+        /// @brief              Output stream operator
+        ///
+        /// @param              [in] anOutputStream An output stream
+        /// @param              [in] aSolution A Solution
+        /// @return             An output stream
+
         friend std::ostream& operator<<(std::ostream& anOutputStream, const Solution& aSolution);
+
+        String name;                       /// Name of the segment.
+        Array<Shared<Dynamics>> dynamics;  /// List of dynamics used.
+        Array<State> states;               /// Array of states for the segment.
+        bool conditionIsSatisfied;         /// True if the event condition is satisfied.
+        Segment::Type segmentType;         /// Type of segment.
     };
 
     /// @brief                  Output stream operator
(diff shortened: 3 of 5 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+            VectorXd readStateCoordinates =
+                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsReadCoordinatesSubsets);
+            VectorXd writeStateCoordinates =
+                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsWriteCoordinatesSubsets);
+
+            VectorXd accelerationContributionAtState = singleDynamics->computeContribution(
+                state.getInstant(),
+                readStateCoordinates,
+                Frame::GCRF()  // TBI: Make more general here
```
phc1990, line 151: Careful, this has to be the Frame in which `readStateCoordinates` are resolved, so, in this code it should be `state.getFrame()`.

I know there is a `TBI`, but just in case.

In any case see my other comment about signature inputs.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
     return Mass::Kilograms(getInitialMass().inKilograms() - getFinalMass().inKilograms());
 }
 
+Array<MatrixXd> Segment::Solution::computeAccelerations() const
```
phc1990, line 106: Nice! just few comments:

I would update the signatrue to:

```
MatrixXd Segment::Solution::getContributionsconst Shared<Dynamics> aDynamcis, const Shared<Frame> aFrame)
```

- `aDynamics` allows us to only re-compute the dynamic we are interested in.
- `aFrame` allows us to specify the output frame in which we would like the contribution to be expressed in. This is not a big rework the only thing you'd need to add is a `state.inFrame(aFrame)` conversion before you to the rest.

On the naming:
- I would go from `Acceleration` to `Contribution` to keep it generic and flexible.
- I know that at the moment this is re-computing, but I would still call this `get` rather than `compute`. The reason being that hopefully, at some point, we don't need to recompute* this here (as it happens during segment solving), then we can start using `get` already and not break the API later on.

*Additional comment on re-computing (not needed or anything, just a suggestion):

If I were to call this twice for the same dynamic, and the same frame, we will be computing it twice, if I then call it again, then three times. Maybe you can have an internal `Map<Dynamics,MatrixXd>` that starts empty. And whenever you're asked the contribution of a Dynamics, you first check if you already computed it by checking the map, otherwise you recompute it and add it to the Map for potential future calls.

phc1990 (commented):
Left some comments

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: Yup agreed

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: > as it happens during segment solving

Yeah for sure, we might be able to cache it now actually, will check

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+{
+    // Each MatrixXd contains the acceleration for a single dynamics for all the segment states
```
apaletta3, line 108:
```suggestion
    // Each MatrixXd contains the acceleration contribution from each Dynamics for each state belonging to the Segment.
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+    // TBI: Return contributions in the same order as the dynamics, might want to standardize later OR use a map with
+    // dynamics name (but need to make sure they are unique and existing)
+    Array<MatrixXd> accelerations = Array<MatrixXd>::Empty();
+    Size stateSize = states.getSize();
+
+    for (const auto& singleDynamics : dynamics)
```
apaletta3, line 114: Can you not make this a Dynamics SPtr type and rely on polymorphism?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+        // TBI: Decide if all contributions should be homogenized already
+        Size dynamicsWriteSize = std::accumulate(
+            dynamicsWriteCoordinatesSubsets.begin(),
+            dynamicsWriteCoordinatesSubsets.end(),
+            0,
+            [](int sum, const Shared<const CoordinatesSubset>& subset)
+            {
+                return sum + subset->getSize();
+            }
+        );
```
apaletta3, line 133: We could build this calculation lines 125-133 into a method of the `Propagator` or `Dynamics` classes maybe? That way it can be leveraged elsewhere, like here.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -202,6 +203,27 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, SegmentSolution_Comput
...
+
+    {
+        const Segment::Solution segmentSolution =
+            Segment::Solution(defaultName_, defaultDynamics_, {defaultState_}, true, Segment::Type::Coast);
+
+        Array<MatrixXd> accelerations = segmentSolution.computeAccelerations();
+
+        std::cout << accelerations << std::endl;
+    }
+}
```
apaletta3, line 225: This test looks like the same exact one as the previous one? Also don't you want to actually check that the values match the default ones and not print stuff to console? (maybe this is still a draft in which case I imagine you were planning on doing that)

apaletta3 (dismissed):
Nice!

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
Derollez: > maybe this is still a draft in which case I imagine you were planning on doing that

Precisely ;)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: I'll check if it's worth it yup

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: so not sure what you mean here, singleDynamics is of type `Shared<Dynamics>` already and I am using the function (common to all `Dynamics` classes) to `computeContribution`, is that what you meant?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: +1 on calling it `get` to future proof. And contribution to keep it generic.
I would punt on the map, that is added complexity for some small (unquantified) performance improvement. If the user is doing multiple calls of `getContributions` let them do the caching.
+1 on having the user provide the dynamic.

We can thus have 2 methods:
- getContributions(frame) (just loops over all dynamics and calls the method below)
- getContribution(dynamic, frame) (which does it for specific dynamics, uses 95% of the code you have here)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
     return Mass::Kilograms(getInitialMass().inKilograms() - getFinalMass().inKilograms());
 }
 
+Array<MatrixXd> Segment::Solution::computeAccelerations() const
+{
+    // Each MatrixXd contains the acceleration for a single dynamics for all the segment states
+    // TBI: Return contributions in the same order as the dynamics, might want to standardize later OR use a map with
+    // dynamics name (but need to make sure they are unique and existing)
+    Array<MatrixXd> accelerations = Array<MatrixXd>::Empty();
```
vishwa2710, line 111:
```suggestion
    Array<MatrixXd> contributions = Array<MatrixXd>::Empty();
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: I think using `auto` when looping through an Array is fine for readibility. but yeah, can always be explicit.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+    Size stateSize = states.getSize();
```
vishwa2710, line 112:
```suggestion
    const Size stateSize = states.getSize();
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+    for (const auto& singleDynamics : dynamics)
+    {
+        // Extract dynamics context and behavior relative to state
+        Array<Shared<const CoordinatesSubset>> dynamicsReadCoordinatesSubsets =
+            singleDynamics->getReadCoordinatesSubsets();
+        Array<Shared<const CoordinatesSubset>> dynamicsWriteCoordinatesSubsets =
+            singleDynamics->getWriteCoordinatesSubsets();
```
vishwa2710, line 120:
```suggestion
        const Array<Shared<const CoordinatesSubset>> dynamicsReadCoordinatesSubsets =
            singleDynamics->getReadCoordinatesSubsets();
        const Array<Shared<const CoordinatesSubset>> dynamicsWriteCoordinatesSubsets =
            singleDynamics->getWriteCoordinatesSubsets();
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+        // Initialize the singleDynamicsAcceleration matrix
+        // Each row corresponds to a state, each column to a coordinate of the (full or reduced?) State
+        // TBI: Decide if all contributions should be homogenized already
+        Size dynamicsWriteSize = std::accumulate(
```
vishwa2710, line 125:
```suggestion
        const Size dynamicsWriteSize = std::accumulate(
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+            {
+                return sum + subset->getSize();
+            }
+        );
+        MatrixXd singleDynamicsAcceleration = MatrixXd::Zero(stateSize, dynamicsWriteSize);
+
+        // Construct the singleDynamicsAcceleration matrix, state by state
+        for (Index stateIndex = 0; stateIndex < stateSize; ++stateIndex)
+        {
+            State state = states[stateIndex];
```
vishwa2710, line 139:
```suggestion
            const State state = states[stateIndex];
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+            Shared<const CoordinatesBroker> coordinatesBroker = state.accessCoordinatesBroker();
```
vishwa2710, line 141:
```suggestion
            const Shared<const CoordinatesBroker>& coordinatesBroker = state.accessCoordinatesBroker();
```
Since we are accessing, we can access as a const reference so as to not copy memory.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+            VectorXd readStateCoordinates =
+                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsReadCoordinatesSubsets);
+            VectorXd writeStateCoordinates =
+                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsWriteCoordinatesSubsets);
```
vishwa2710, line 146:
```suggestion
            const VectorXd readStateCoordinates =
                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsReadCoordinatesSubsets);
            const VectorXd writeStateCoordinates =
                coordinatesBroker->extractCoordinates(state.getCoordinates(), dynamicsWriteCoordinatesSubsets);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+            VectorXd accelerationContributionAtState = singleDynamics->computeContribution(
```
vishwa2710, line 148:
```suggestion
            const VectorXd accelerationContributionAtState = singleDynamics->computeContribution(
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+                state.getInstant(),
```
vishwa2710, line 149:
```suggestion
                state.accessInstant(),
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +103,65 @@ Mass Segment::Solution::computeDeltaMass() const
...
+                Frame::GCRF()  // TBI: Make more general here
+            );
+
+            singleDynamicsAcceleration.row(stateIndex) = accelerationContributionAtState;
+        }
+
+        accelerations.add(singleDynamicsAcceleration);
+    }
+
+    // TBI: Cache at solution determination time?
```
vishwa2710, line 160:
```suggestion
```
Not necessary IMO.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -202,6 +203,27 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, SegmentSolution_Comput
...
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, SegmentSolution_ComputeAccelerations)
+{
+    {
+        const Segment::Solution segmentSolution =
+            Segment::Solution(defaultName_, defaultDynamics_, {defaultState_}, true, Segment::Type::Coast);
+
+        Array<MatrixXd> accelerations = segmentSolution.computeAccelerations();
+
+        std::cout << accelerations << std::endl;
```
vishwa2710, line 214:
```suggestion
        EXPECT_EQ(contributions.getSize(), segmentSolution.dynamics.getSize());
        for (const auto& contribution : contributions)
        {
            EXPECT_EQ(contribution.getSize(), segmentSolution.states.getSize());
        }
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -202,6 +203,27 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, SegmentSolution_Comput
     }
 }
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, SegmentSolution_ComputeAccelerations)
+{
+    {
+        const Segment::Solution segmentSolution =
+            Segment::Solution(defaultName_, defaultDynamics_, {defaultState_}, true, Segment::Type::Coast);
+
+        Array<MatrixXd> accelerations = segmentSolution.computeAccelerations();
```
vishwa2710, line 212:
```suggestion
        const Array<MatrixXd> accelerations = segmentSolution.computeAccelerations();
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
```diff
@@ -170,6 +170,17 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
 
             )doc"
         )
+        .def(
+            "compute_accelerations",
```
vishwa2710, line 174:
```suggestion
            "get_contributions",
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
```diff
@@ -170,6 +170,17 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
...
+            &Segment::Solution::computeAccelerations,
```
vishwa2710, line 175:
```suggestion
            &Segment::Solution::getContributions,
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
```diff
@@ -170,6 +170,17 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
...
+            R"doc(
+                Compute the accelerations contributions of the involved dynamics.
```
vishwa2710, line 177:
```suggestion
                Get the contributions of the involved dynamics.
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
```diff
@@ -170,6 +170,17 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
...
+
+                Returns:
+                    list[MatrixXd]: The list of acceleration matrices.
```
vishwa2710, line 180:
```suggestion
                    list[MatrixXd]: The list of contributions matrices.
```

bindings/python/test/trajectory/test_segment.py
```diff
@@ -220,3 +220,10 @@ def test_solve(
 
         assert solution.compute_delta_mass() is not None
         assert solution.compute_delta_v(1500.0) is not None
+
+        accelerations = solution.compute_accelerations()
+
+        print(accelerations[0])
+        print(accelerations[1])
+        print(accelerations[2])
+        # assert solution.compute_accelerations() is not None
```
vishwa2710, line 229:
```suggestion
        assert solution.get_contributions() is not None
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -75,6 +79,8 @@ class Segment
         Real computeDeltaV(const Real& aSpecificImpulse) const;
         Mass computeDeltaMass() const;
 
+        Array<MatrixXd> computeAccelerations() const;
```
vishwa2710, line 82:
```suggestion
        Array<MatrixXd> getAccelerations() const;
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -1,5 +1,7 @@
 /// Apache License 2.0
 
+#include <numeric>
```
vishwa2710, line 3:
```suggestion
```
Is this required?

vishwa2710 (commented):
Great stuff!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: Yup, needed for the std::accumulate

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+    const Array<Shared<const CoordinatesSubset>> dynamicsWriteCoordinatesSubsets =
+        aDynamicsSPtr->getWriteCoordinatesSubsets();
+
+    // Check that the provided coordinates subsets are part of the dynamics write coordinates subsets
+    for (auto aCoordinatesSubsetSPtr : aCoordinatesSubsetSPtrArray)
+    {
+        if (!dynamicsWriteCoordinatesSubsets.contains(aCoordinatesSubsetSPtr))
+        {
+            throw ostk::core::error::RuntimeError(
+                "Provided coordinates subset is not part of the dynamics write coordinates subsets."
```
vishwa2710, line 130:
```suggestion
                String::Format("Provided coordinates subset [{}] is not part of the dynamics write coordinates subsets.", aCoordinatesSubsetSPtr->getName())
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+        }
+    );
+
+    // Initialize the dynamicsContributionMatrix
+    MatrixXd dynamicsContributionMatrix = MatrixXd::Zero(stateSize, dynamicsWriteSize);
+
+    // Construct the dynamicsContributionMatrix, state by state (a.k.a row by row)
+    for (Index stateIndex = 0; stateIndex < stateSize; ++stateIndex)
+    {
+        const State state = states[stateIndex];
```
vishwa2710, line 167:
```suggestion
        const State& state = states[stateIndex];
```
Nit: so that we're not copying the values unnecessarily.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+        Shared<const CoordinatesBroker> coordinatesBrokerSPtr = state.accessCoordinatesBroker();
```
vishwa2710, line 169:
```suggestion
        const Shared<const CoordinatesBroker>& coordinatesBrokerSPtr = state.accessCoordinatesBroker();
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+        VectorXd readStateCoordinates =
+            coordinatesBrokerSPtr->extractCoordinates(state.getCoordinates(), dynamicsReadCoordinatesSubsets);
+        VectorXd writeStateCoordinates =
+            coordinatesBrokerSPtr->extractCoordinates(state.getCoordinates(), definitiveCoordinateSubsetArray);
+
+        VectorXd dynamicsContributionAtState =
+            aDynamicsSPtr->computeContribution(state.getInstant(), readStateCoordinates, aFrameSPtr);
```
vishwa2710, line 177:
```suggestion
        const VectorXd readStateCoordinates =
            coordinatesBrokerSPtr->extractCoordinates(state.getCoordinates(), dynamicsReadCoordinatesSubsets);
        const VectorXd writeStateCoordinates =
            coordinatesBrokerSPtr->extractCoordinates(state.getCoordinates(), definitiveCoordinateSubsetArray);

        const VectorXd dynamicsContributionAtState =
            aDynamicsSPtr->computeContribution(state.getInstant(), readStateCoordinates, aFrameSPtr);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+Map<Shared<Dynamics>, MatrixXd> Segment::Solution::getDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const
+{
+    // TBI: Use smart caching for multiple calls in the future
+
+    // Each MatrixXd contains the contribution of a single dynamics for all the segment states
+    Map<Shared<Dynamics>, MatrixXd> dynamicsContributionsMap = Map<Shared<Dynamics>, MatrixXd>();
+
+    for (const Shared<Dynamics>& aDynamicsSPtr : this->dynamics)
+    {
+        MatrixXd dynamicsContribution = this->getDynamicsContribution(aDynamicsSPtr, aFrameSPtr);
```
vishwa2710, line 203:
```suggestion
        const MatrixXd dynamicsContribution = this->getDynamicsContribution(aDynamicsSPtr, aFrameSPtr);
```

vishwa2710 (approved):
Looks great! Just some nits.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -75,6 +81,17 @@ class Segment
         Real computeDeltaV(const Real& aSpecificImpulse) const;
         Mass computeDeltaMass() const;
 
+        Map<Shared<Dynamics>, MatrixXd> getDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const;
```
phc1990, line 84: Since the return type is quite different from the one below, maybe `getAllDynamicsContributions` ?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+    }
+
+    return dynamicsContributionMatrix;
+}
+
+MatrixXd Segment::Solution::getDynamicsAccelerationContribution(
+    const Shared<Dynamics>& aDynamicsSPtr, const Shared<const Frame>& aFrameSPtr
+) const
+{
+    const MatrixXd accelerationContribution =
```
phc1990, line 189: Maybe no need to create the variable

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -37,7 +43,7 @@ using ostk::astro::Dynamics;
 using ostk::astro::dynamics::Thruster;
 using ostk::astro::EventCondition;
 
-/// @brief                                  Represent a propagation segment for astrodynamics purposes
+/// @brief                      Represent a propagation segment for astrodynamics purposes
 
 class Segment
```
phc1990, line 48: should all of these return functions, `getDynamics...` but also `computedeltaV` etc. return `const`?

e.g. `const MatrixXd getDynamics...`

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -75,6 +81,17 @@ class Segment
         Real computeDeltaV(const Real& aSpecificImpulse) const;
         Mass computeDeltaMass() const;
 
+        Map<Shared<Dynamics>, MatrixXd> getDynamicsContributions(const Shared<const Frame>& aFrameSPtr) const;
+        MatrixXd getDynamicsContribution(
```
phc1990, line 85: I think we're missing docs for these functions. Don't know how important it is.
But if we add them we could state that "at the moment, the dynamics are recomputed"

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -75,6 +81,17 @@ class Segment
...
+            const Shared<Dynamics>& aDynamicsSPtr,
+            const Shared<const Frame>& aFrameSPtr,
+            const Array<Shared<const CoordinatesSubset>>& aCoordinatesSubsetSPtrArray =
```
phc1990, line 88: would be great if this was a variable argument rather than an array

```
    (
            const Shared<Dynamics>& aDynamicsSPtr,
            const Shared<const Frame>& aFrameSPtr,
            Shared<const CoordinatesSubset> coordinatesSubsetSPtrs, ...
    )
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+    Array<Shared<const CoordinatesSubset>> definitiveCoordinateSubsetArray = aCoordinatesSubsetSPtrArray;
+
+    // Check value for aCoordinatesSubsetSPtrArray
+    if (aCoordinatesSubsetSPtrArray.isEmpty())
+    {
+        definitiveCoordinateSubsetArray = aDynamicsSPtr->getWriteCoordinatesSubsets();
+    }
+
+    // Extract states size
+    const Size stateSize = this->states.getSize();
```
phc1990, line 145: This is the number of rows right? (i.e. number of states?)

Just asking as `stateSize` sound like the size of the state (i.e. number of columns)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+    // Extract dynamics context and behavior relative to state
+    Array<Shared<const CoordinatesSubset>> dynamicsReadCoordinatesSubsets = aDynamicsSPtr->getReadCoordinatesSubsets();
+
+    // Compute the size of dynamicsContributionMatrix
+    Size dynamicsWriteSize = std::accumulate(
```
phc1990, line 151: Since you're talking about "rows" in the comments below, maybe we could call them `numberOfRows and numberOfColumns` ?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
phc1990: Two things:

1. What is `writeStateCoordinates` used for?
2. ⚠️ `computeContribution(state.getInstant(), readStateCoordinates, aFrameSPtr` has not made sure that `readStateCoordinates` are actually resolved in `aFrameSPtr` (at least I could not find it).

phc1990: ⚠️ Actually, all-in-all I think it is very much worth it here to just use the `StateBuilder`:

```cpp
const StateBuilder builder = StateBuilder(aFrameSPTr, dynamicsReadCoordinatesSubsets);
```

Then you just use it in the for loop to get the read state coordinates:
```cpp
const VectorXd dynamicsContribution = aDynamicsSPTr->computeContribution(
    state.getInstant(),
    builder.reduce(state.inFrame(aFrameSPTr)), // Consider frame consistency
    aFrameSPTr);
```

I think this is a good use case for the StateBuilder. It will sort thing sort things out for you, warn you if coordinates are not found, if frames are not consistent, etc.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,108 @@ Mass Segment::Solution::computeDeltaMass() const
     return Mass::Kilograms(getInitialMass().inKilograms() - getFinalMass().inKilograms());
 }
 
+MatrixXd Segment::Solution::getDynamicsContribution(
```
phc1990, line 108: Nit: I think the order of the new 3 functions in the cpp is not the same as in the hpp.

phc1990 (approved):
Nice, left some nit comments and a "major" one (marked with ⚠️ )

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: Thanks 🙏 Sorted

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: Good point, that's confusing adapting

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: Adapted the naming of `stateSize`, I think `MatrixXd::Zero(numberOfstates, dynamicsWriteSize);` conveys the meaning we are using each of these for (rows, columns), lmk

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
Derollez: > Shared<const CoordinatesSubset> coordinatesSubsetSPtrs

Would allow a single coordinatesSubsetPtr though, is that what you would want? We can always use an Array of a single Pointer, nah?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
Derollez: 1. Agreed, should be removed, not used
2. Great catch, using the StateBuilder suggestion 💯

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
Derollez: So usually the philosophy for OSTk is to have getters and compute functions not return `const` and if wanted, implement a `access*` function which would return `const` type that then can't be modified. We can add accessors in the future on top if wanted, see
```

        const Instant& accessStartInstant() const;

        /// @brief              Access end instant
        /// @return             End Instant

        const Instant& accessEndInstant() const;

        /// @brief              Get initial mass
        /// @return             Initial mass

        Mass getInitialMass() const;

        /// @brief              Get final mass
        /// @return             Final mass

        Mass getFinalMass() const;

        /// @brief              Get propagation duration
        /// @return             Propagation duration
```
for instance

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -101,6 +105,104 @@ Mass Segment::Solution::computeDeltaMass() const
...
+
+    // Initialize the dynamicsContributionMatrix
+    MatrixXd dynamicsContributionMatrix = MatrixXd::Zero(numberOfstates, dynamicsWriteSize);
+
+    // Construct the dynamicsContributionMatrix, state by state (a.k.a row by row)
+    for (Index stateIndex = 0; stateIndex < numberOfstates; ++stateIndex)
+    {
+        const State& state = states[stateIndex];
+
+        const StateBuilder builder = StateBuilder(aFrameSPtr, dynamicsReadCoordinatesSubsets);
```
phc1990, line 170: move out of for loop

phc1990 (approved):
Great stuff!

PR #293, opened 2023-12-04
Draft: feat: add equinoctial orbital elements

Derollez:
- [ ] Clean definitions and format
- [ ] Add tests
- [ ] Add bindings
- [ ] Check definitions and conversions on edge cases


vishwa2710:
Siiiiick. I think it's worth maybe re-organizing under an Orbital Elements class under orbits, where we can house COE, Mean Orbital Elements, Equinoctial, modified equinoctial etc.

PR #295, opened 2023-12-06
feat: add fixed step size factory method for numerical solver

apaletta3:
This adds a static constructor to enable the creation of a fixed step size numerical solver with the RK4 stepper type like so:
```
solver = NumericalSolver.fixed_step_size(RK4, 30.0)
```

### DIFF
--- a/bindings/python/test/trajectory/state/test_numerical_solver.py
+++ b/bindings/python/test/trajectory/state/test_numerical_solver.py
@@ -282,7 +282,7 @@ class TestNumericalSolver:
         assert NumericalSolver.undefined() is not None
         assert NumericalSolver.undefined().is_defined() is False
 
-    def test_fixed_step_size_factory(
+    def test_fixed_step_size(
         self,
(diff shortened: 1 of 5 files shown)


bindings/python/test/trajectory/state/test_numerical_solver.py
```diff
@@ -271,14 +278,27 @@ def test_integrate_conditional_with_logger(
...
 
-    def test_default_conditional(self, state_logger):
-        assert NumericalSolver.default_conditional() is not None
-        assert NumericalSolver.default_conditional(state_logger) is not None
-
     def test_undefined(self):
         assert NumericalSolver.undefined() is not None
         assert NumericalSolver.undefined().is_defined() is False
 
+    def test_fixed_step_size_factory(
```
vishwa2710, line 285:
```suggestion
    def test_fixed_step_size(
```

vishwa2710 (approved):
Looks good!

PR #298, opened 2023-12-07
feat: add validation app and parser

apaletta3:
Validation framework ready for review. I'm planning on adding unit tests for the `CrossValidator`, `Parser`, and `MissionSequence` in parallel while you guys are reviewing (this review should focus on the code functionality, cause I also have docstrings to add)

The `Framework.validation.cpp` file has all the validation mission sequences and the main parametrized loop that performs the setup, running, and comparison of the `MissionSequence` to one other tool. 
- `Parser` is supposed to take a yaml scenario and convert it into an ostk version that defines that scenario
- `MissionSequence` is supposed to be that ostk version, where the scenario can be `.run`
- `CrossValidator` is supposed to contain the tools and variables to comparing the states from `MissionSequence` to the states read in from the csv file corresponding to a specific tool and specific scenario.

Questions for the reviewer:
- I think I may have made my life more difficult by manually handling coordinate subsets in the `IngestOutputQuantities` and `CompareOutputQuantities` functions, let me know if it could be refactored to be more concise and clearer by actually using `CoordinateSubset`. Also I think that would be more robust since right now the order of the `OutputQuantities` needs to stay the same otherwise the comparison will break...
- Also, given the coming refactor of the dynamics and addition of dynamics contribution "observing" that we will likely do, I chose to not include the capability to compare the acceleration of the ostk propagation to GMAT/Orekit accelerations in this MR, since it will be easier to add a method to `Sequence` and `Segment` to handle this once the dynamics contribution "observing" is in place. You can see this in the comments variables in `MissionSequence` and `CrossValidator`


validation/Validation.cpp
```diff
@@ -0,0 +1,69 @@
...
+        std::vector<std::string> compareWith_;
+        std::vector<double> tolerance_;
+    };
+
+    void SetUp() override
+    {
+        // List of validation test scenarios go here (replacement for a "super yaml file", althought this could be added
+        // in the future)
+        const std::vector<TestCase> scenarios_ = {
+            {"SCENARIO_1: Spherical gravity only", {"GMAT", "OREKIT"}, {1.0e-6, 1.0e-6}},
```
phc1990, line 20: Should the friendly scenario name be part of the yaml?

Also, how does this link to the yaml/csv file name?

phc1990 (commented):
Lef one small comment and a "big one"

validation/data/gmat_astrodynamics/gmat_scripts/scenario001-mission-sequence.script
```diff
@@ -0,0 +1,141 @@
+% Copyright Loft Orbital Solutions Inc.
```
Derollez, line 1: Storage of these files here is temporary (except for testing of the app) agreed?

validation/MissionSequence.hpp
```diff
@@ -0,0 +1,25 @@
+
+#include <any>
+#include <iostream>
+#include <map>
+
+class MissionSequence
+{
+   public:
+    struct ComparisonResults
```
Derollez, line 9:
```suggestion
    struct Metrics
```

?

Derollez: or simply `Results`

validation/MissionSequence.hpp
```diff
@@ -0,0 +1,25 @@
...
+    {
+        Position OSTkPosition_Minus_OtherPosition_RMS_;
+        Position OSTkVelocity_Minus_OtherVelocity_RMS_;
+    };
+
+    MissionSequence(const std::map<std::string, std::any>& data_tree);
```
Derollez, line 15:
```suggestion
    MissionSequence(const std::map<std::string, std::any>& dataTree);
```

and adapt casing elswhere

validation/Validation.cpp
```diff
@@ -0,0 +1,69 @@
+#include <iostream>
+
```
Derollez, line 2: This should be in `/tests/` at least for starter, no?

validation/Validation.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <Global.test.hpp>
+
+class ValidationTestDefinitions : public ::testing::Test
+{
+   protected:
+    struct TestCase
+    {
+        std::string testName_;
+        std::vector<std::string> compareWith_;
+        std::vector<double> tolerance_;
```
Derollez, line 12: break down into explicit names for tolerances for now, otherwise kinda hard to know what tol we are talking about

validation/Validation.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <Global.test.hpp>
+
+class ValidationTestDefinitions : public ::testing::Test
+{
+   protected:
+    struct TestCase
+    {
+        std::string testName_;
+        std::vector<std::string> compareWith_;
```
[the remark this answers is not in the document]
Derollez: - I like the enum for the tools, it makes sense yeah
- Same for parameters
- "Note: if we think "per-tool" tolerance might be overkill": I think we should support this straight away: our current tests are pretty "simple" and match pretty well, that will likely not be the case in the future
- Yeah same, that's a good idea

validation/Validation.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+            // Assert on comparison results
+            EXPECT_GT(scenario_.tolerance_[index], comparisonResults[index]);
+        }
+    }
+}
+
+// Alternative approach, validation test definitions in a separate yaml file and parsed in:
+// std::ifstream configFile("validation/ValidationConfig.yaml");
+// YAML::Node config = YAML::Load(configFile);
```
Derollez, line 56: This is the future alternative with `yaml` yeah?

validation/data/gmat_astrodynamics/gmat_scripts/scenario001-mission-sequence.script
apaletta3: You mean the gmat script files? I am keeping them for now because they are handy to use when debugging is needed but yes eventually there's no need for them.

validation/Validation.cpp
apaletta3: Correct.

validation/data/gmat_astrodynamics/gmat_scripts/scenario001-mission-sequence.script
Derollez: > but yes eventually there's no need for them.

✅

validation/Validation.cpp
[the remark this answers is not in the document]
apaletta3: Yup I agree, made all of these changes. And yes, we will need per tool tolerances from the get go so I'm implementing that.

apaletta3:
Made my first refinement of the apps. I plan to make the validation google tests into a `TEST_P` parametrized fixture in the future, but i'll get around to that once i've finished the `MissionSequence` and `Parser` apps.

validation/Validation.cpp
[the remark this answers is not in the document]
Derollez: > And yes, we will need per tool tolerances from the get go so I'm implementing that.

✅ 🙏 (That was basically the comment I had on this @apaletta3)

test/ValidationFramework/Validation.hpp
```diff
@@ -0,0 +1,125 @@
+/// Apache License 2.0
+
+#include <iostream>
+
+#include <Global.test.hpp>
+
+class Validation : public ::testing::Test
+{
+   public:
+    enum class OutputQuantity
```
Derollez, line 10: Observable?

test/ValidationFramework/Validation.hpp
```diff
@@ -0,0 +1,125 @@
...
+    {
+        GMAT,
+        OREKIT,
+    };
+
+    struct Comparison
+    {
+        Tool tool_;
+        std::map<OutputQuantity, double> toleranceMap_;
+    };
```
Derollez, line 29: In the future, we can imagine defining different types of comparisons: one that is active for all states saved, one that needs to be checked only at the end of the propagation etc

test/ValidationFramework/Validation.hpp
```diff
@@ -0,0 +1,125 @@
+/// Apache License 2.0
+
+#include <iostream>
+
+#include <Global.test.hpp>
+
+class Validation : public ::testing::Test
```
phc1990, line 7: LGTM, will resole the big comment I left 👍

test/ValidationFramework/Validation.hpp
phc1990, line 10: Will eventually need some static function that clusters a `String` into an `OutputQuantity`

test/ValidationFramework/Validation.hpp
```diff
@@ -0,0 +1,125 @@
...
+                        Tool::GMAT,
+                        {
+                            {OutputQuantity::CARTESIAN_POSITION, 1.0e-3},
+                            {OutputQuantity::CARTESIAN_VELOCITY, 1.0e-6},
+                        },
+                    },
+                    {
+                        Tool::OREKIT,
+                        {
+                            {OutputQuantity::CARTESIAN_POSITION, 1.0e-3},
```
Derollez, line 116: Do we have standardized names for the output columns in csv in the different tools, yeah?

validation/Validation.cpp
phc1990: I agree with having a human-friendly scenario name 👍 

What about this? 

> Also, how does this link to the yaml/csv file name?

Where do you specify the actual file name, e.g. `scenario-001.` ?

Maybe I'm getting ahead/missing something.

phc1990 (commented):
Nice rework 👍

test/ValidationFramework/Parser/Parser.hpp
```diff
@@ -0,0 +1,149 @@
+/// Apache License 2.0
```
Derollez, line 1: So, the `parser` is a class holding a bunch of methods that allow to convert a yaml (and maybe in the future other input formats if needed, like `xml` or else as this is usually used in CDM, OEM and others as per CCSDS recommendations) into OSTk objects (`SatelliteSystem`, `NumericalSolver`, `Propagator`), IMO it can be used outside of the testing and validation framework, it is another way to configure OSTk, do you agree?

Derollez: The apps are different as they need to match other apps from other tools etc, but still, there might be some benefits in having them outside of `tests` or `validation` IMO

Derollez: Then we should use those concepts in `tests` and `validation` ofc to standardize the way we set them up and perform comparison but it's one use case for `Parser` etc, do you agree?

test/ValidationFramework/Validation.test.cpp
```diff
@@ -0,0 +1,51 @@
...
+// {
+//     TestCase testCase;
+//     testCase.testName = scenario["test"].as<std::string>();
+//     for (const auto& compareWith : scenario["compare-with"])
+//     {
+//         testCase.compareWith.push_back(compareWith.as<std::string>());
+//     }
+//     testCase.tolerance = scenario["tolerance"].as<double>();
+//     validationConfig_.scenarios.push_back(testCase);
+// }
```
Derollez, line 51: Let's remove this for now as we go for approach 1. Let's use the proposal or a separate MR later to show how we would go from the intermediate step we are taking right now to this next step

test/ValidationFramework/Validation.hpp
Derollez: Yep, same comment on my part below ✅

test/ValidationFramework/MissionSequence/MissionSequence.hpp
```diff
@@ -0,0 +1,153 @@
...
+using ostk::astro::dynamics::AtmosphericDrag;
+using ostk::astro::dynamics::Thruster;
+using ostk::astro::guidancelaw::ConstantThrust;
+using ostk::astro::guidancelaw::QLaw;
+using ostk::astro::solvers::FiniteDifferenceSolver;
+
+static const Derived::Unit GravitationalParameterSIUnit =
+    Derived::Unit::GravitationalParameter(Length::Unit::Meter, ostk::physics::units::Time::Unit::Second);
+
+class MissionSequence
```
Derollez, line 141: Might want to add all the file output writing and common func into an `Application` abstract class in the future (probably when we start adding a second `App`), what do you think?

validation/Validation.cpp
apaletta3: I'll just parse it out of the string or something like that to indicate the filename

test/ValidationFramework/Validation.hpp
apaletta3: Yup

apaletta3 (commented):
@phc1990 @Derollez Responsed to your Q's

apaletta3 (commented):
@phc1990 @Derollez Responsed to your Q's

test/ValidationFramework/Validation.hpp
Derollez: It depends, in trajectory optimization or solvers, the intermediate steps might be different for instance and diverge a bit for while depending on the heuristics, anyway that's for later, no need to change now

test/ValidationFramework/Validation.hpp
Derollez: Just a suggestion on naming

CMakeLists.txt
```diff
@@ -448,6 +449,76 @@ IF (BUILD_UNIT_TESTS)
...
+
+    #     SET (COVERAGE_EXCLUDES '*.test.cpp' '/usr/*' '/app/bindings/*' '*.benchmark.cpp')
+
+    #     SETUP_TARGET_FOR_COVERAGE (
+    #         NAME ${COVERAGE_TARGET}
+    #         EXECUTABLE ${VALIDATION_TESTS_TARGET}
+    #         DEPENDENCIES "coverage"
+    #     )
+
+    # ENDIF ()
```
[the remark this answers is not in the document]
Derollez: > For coverage I can delete all these comments cause we don't need "validation coverage", 

That's what I mean

Agreed with the rest

test/ValidationFramework/Validation.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+   protected:
+    void SetUp() override
+    {
+        // List of validation test scenarios go here (replacement for a "super yaml file", althought this could be added
+        // in the future)
+
+        testCases_ = {
+            {
+                "001: Spherical gravity only",
```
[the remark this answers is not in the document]
Derollez: let me think about it

test/ValidationFramework/Parser/Parser.hpp
Derollez: Okay perfect aligned. Let's keep them there for now I agree ✅

test/ValidationFramework/MissionSequence/MissionSequence.hpp
Derollez: > (probably when we start adding a second App)

Yeah that was my point with this, we can refactor when we have more later maybe. Let's see

test/ValidationFramework/Validation.hpp
apaletta3: Yeah, but I think the scope of this specific app should stay as a cross validation platform, and we can see what our needs in terms of self-validation are in the future for the solvers.

apaletta3:
Made a little more progress on the validation app this week, but its still in early form. For this MR, I would like to have:
- a working `Parser`
- a working `MissionSequence`
- unit tests for both of those

The parametrization of the validation test case and refactoring of all the existing test cases for `Propagator` into this validation format can come in a following MR.

test/ValidationFramework/Framework.validation.cpp
```diff
@@ -0,0 +1,291 @@
...
+    for (const ToolComparison& toolComparison : toolComparisons)
+    {
+        const Tool tool = toolComparison.tool;
+        const Array<QuantityComparison> quantityComparisons = toolComparison.quantityComparisons;
+
+        // Compare output quantities with reference tool
+        const Array<VectorXd> errors =
+            missionSequence.compareResults(Parser::ParseCSV(pathToData, scenarioName, tool), toolComparison);
+
+        for (Size coordinateSubsetIndex = 0; coordinateSubsetIndex < quantityComparisons.getSize();
```
phc1990, line 191: Just a small detail:

Here your iterating first through `QuantityComparison` (outer loop), then through `States` (inner loop). E.g.:
```
- Test position[0]
- Test position[1]
- ...
- Test position[N]
- Test velocity[0]
- Test velocity[1]
- ...
```

Would it be more intuitive to test by state then by coordinate?
```
- Test position[0], test velocity[0], test mass[0]  // Tested whole State [0]
- Test position[1], test velocity[1], test mass[1]  // Tested whole State [1]
- ...
- 
- Test position[N], test velocity[N], test mass[N]  // Tested whole State [N]
```

It's no different than what you've done. It would just be swapping the loops. Just thought rows then cols might feel more _natural._

phc1990 (approved):
LGTM left minor comments.

test/ValidationFramework/Parser.cpp
```diff
@@ -0,0 +1,365 @@
...
+        CartesianVelocity::Default(),
+        CoordinatesSubset::Mass(),
+        CoordinatesSubset::SurfaceArea(),
+        CoordinatesSubset::DragCoefficient()
+    };
+
+    return {StateBuilder(frame, subsets).build(initialInstant, coordinates), satelliteSystem};
+}
+
+Array<Shared<Dynamics>> Parser::CreateEnvironmentAndDynamics(const Dictionary& aDictionary)
```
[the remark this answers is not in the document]
apaletta3: Oh you think it would be? What use case do you have in mind?

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,75 @@
...
+namespace validation
+{
+
+MissionSequence::MissionSequence(const Dictionary& aDataTree)
+    : dataTree_(aDataTree),
+      initialStateAndSatelliteSystem_(Parser::CreateInitialStateAndSatelliteSystem(dataTree_)),
+      environmentAndDynamicsSPtr_(Parser::CreateEnvironmentAndDynamics(dataTree_)),
+      sequence_(Parser::CreateSequence(dataTree_, initialStateAndSatelliteSystem_, environmentAndDynamicsSPtr_)),
+      solvedStates_(Array<State>::Empty())
+//   solvedDynamicsContributions_(Array<MatrixXd>::Empty())
```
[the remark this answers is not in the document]
apaletta3: Yeah I just kept them in for the review to show what I was intending to store in terms of dynamics contributions, but yes i'll remove them.

test/ValidationFramework/Parser.cpp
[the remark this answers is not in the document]
apaletta3: Actually I agree with you, I'm thinking why not have both the environment and dynamics as memvars, and then it'll be simpler in the future when we do the refactor of the yaml and those things

test/ValidationFramework/Framework.validation.cpp
apaletta3: Yeah I thought about that, but I chose the current layout to follow the structure of the `parameters` variable that holds all the test cases. The only way to have the test cases listed at the bottom of the file in a visually easy to understand way is to have the output quantity and tolerance in the same structure `QuantityComparison` (otherwise the quantity and the tolerance do not appear next to each other in each test case and it because harder to track which quantity should be compared with which tolerance). The looping structure I have reflects this structure by looping through all the states corresponding to one `QuantityComparison`.

test/ValidationFramework/Parser.cpp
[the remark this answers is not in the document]
phc1990: There might not be a usecase for it yet.
My suggestion was that, if your function returns only one thing then:
- its name should be `CreateX` (where `X` is the type of that thing)
- if we have to choose between returning `Dynamics` and `Environment` then I think `Environment` might be more future-proof as one can get `Dynamics` from `Environment` (with only 1 line) but not the other way around

phc1990:
@apaletta3 

`Also @phc1990 let me know what you think about this question I left at the top

I think I may have made my life more difficult by manually handling coordinate subsets in the IngestOutputQuantities and CompareOutputQuantities functions, let me know if it could be refactored to be more concise and clearer by actually using CoordinateSubset. Also I think that would be more robust since right now the order of the OutputQuantities needs to stay the same otherwise the comparison will break...`

So I think that there is room for refactor to simplify and generalise. However, I think what you did is great for now so I wouldn't touch it 👍

test/ValidationFramework/CrossValidator.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <ValidationFramework/CrossValidator.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace validation
+{
+
+Array<Array<VectorXd>> CrossValidator::IngestOutputQuantities(
```
Derollez, line 12: Worth using other containers for those for perf in the future (not saying we should do this now?) but might need some improvements for longer scenarios in the future?

test/ValidationFramework/CrossValidator.cpp
```diff
@@ -0,0 +1,125 @@
...
+    return groupedCoordinatesArray;
+}
+
+VectorXd CrossValidator::CompareOutputQuantities(
+    const State& anOSTkState,
+    const Array<VectorXd>& aReferenceCoordinates,
+    const Array<QuantityComparison>& aQuantityComparisons
+)
+{
+    VectorXd errorsPerQuantities(aQuantityComparisons.getSize());
```
Derollez, line 61: I would switch `error` to `delta` because uncertainties are present everywhere including in other providers externally (nit)

test/ValidationFramework/data/gmat_astrodynamics/gmat_scripts/scenario004-mission-sequence.script
```diff
@@ -0,0 +1,230 @@
+% Copyright Loft Orbital Solutions Inc.
```
Derollez, line 1: Just to be 100% sure: This is what is basically built by the gmat tooling to then run and obtain the output csv: I don't think we need them here (totally agree we should keep them in the artifacts though 💯 !)

CMakeLists.txt
```diff
@@ -16,7 +16,8 @@ SET (PROJECT_VENDOR_URL "open-space-collective.org")
```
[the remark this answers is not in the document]
Derollez: Similarly I'd add really quick in the doc/README what modification to be made when regenerating output csv from external tooling/modifying a scenario yml or adding a new one so people can use this easily

Derollez (approved):
HUUUGE! 💪  One note for the future: we might need to think about adding python bindings (and maybe change the structure of the files a bit?) such that users can leverage the Parser/Mission Sequence app to seamlessly run mission analysis/astrodynamics scenarios with very few lines in python (and with the standardized yml input put forward in here). Most comments are nits and style

test/ValidationFramework/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+    else
+    {
+        throw ostk::core::error::runtime::Wrong("Tool not recognized.");
+    }
+
+    return Table::Load(File::Path(Path::Parse(csvPath)), Table::Format::CSV, true);
+}
+
+Pair<State, SatelliteSystem> Parser::CreateInitialStateAndSatelliteSystem(const Dictionary& aDictionary)
```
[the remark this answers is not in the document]
apaletta3: Yes, but also cross sectional area and drag coeff

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,100 @@
...
+MissionSequence::~MissionSequence() {}
+
+void MissionSequence::run()
+{
+    Sequence::Solution sequenceSolution = sequence_.solve(initialStateAndSatelliteSystem_.first);
+    const Array<Instant> comparisonInstants = Parser::CreateComparisonInstants(
+        dataTree_, sequenceSolution.accessStartInstant(), sequenceSolution.accessEndInstant()
+    );
+
+    solvedStates_ = sequenceSolution.reComputeStatesAt(comparisonInstants, sequence_.getNumericalSolver());
```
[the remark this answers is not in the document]
apaletta3: What do you mean the purpose? There's no other way to obtain states at an arbitrary set of instants from a sequence solution currently.

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+    return dynamics_;
+}
+
+const Sequence& MissionSequence::accessSequence() const
+{
+    // if (sequence_.isDefined())
+    // {
+    //     throw ostk::core::error::RuntimeError("No sequence available.");
+    // }
```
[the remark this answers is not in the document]
apaletta3: Yeah, `sequences` don't have an `isDefined()` method, and it has to do with the fact that you can create one without any arguments and add segments and other internals after constructions, so whether or not a sequence is defined is a little nebulous. I'll just remove the check here.

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,100 @@
...
+    );
+
+    solvedStates_ = sequenceSolution.reComputeStatesAt(comparisonInstants, sequence_.getNumericalSolver());
+}
+
+const Pair<State, SatelliteSystem>& MissionSequence::accessInitialStateAndSatelliteSystem() const
+{
+    if (initialStateAndSatelliteSystem_.first.isDefined() || initialStateAndSatelliteSystem_.second.isDefined())
+    {
+        throw ostk::core::error::RuntimeError("No initial state or satellite system available.");
```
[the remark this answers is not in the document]
apaletta3:
```suggestion
        throw ostk::core::error::RuntimeError("No initial state or satellite system defined.");
```

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return initialStateAndSatelliteSystem_;
+}
+
+const Environment& MissionSequence::accessEnvironment() const
+{
+    if (environment_.isDefined())
+    {
+        throw ostk::core::error::RuntimeError("No environment available.");
```
apaletta3, line 47:
```suggestion
        throw ostk::core::error::RuntimeError("No environment defined.");
```

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    return environment_;
+}
+
+const Array<Shared<Dynamics>>& MissionSequence::accessDynamics() const
+{
+    if (dynamics_.isEmpty())
+    {
+        throw ostk::core::error::RuntimeError("No dynamics available.");
```
apaletta3, line 56:
```suggestion
        throw ostk::core::error::RuntimeError("No dynamics defined.");
```

test/ValidationFramework/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+const Sequence& MissionSequence::accessSequence() const
+{
+    return sequence_;
+}
+
+const Array<State>& MissionSequence::accessSolvedStates() const
+{
+    if (solvedStates_.isEmpty())
+    {
+        throw ostk::core::error::RuntimeError("No states available.");
```
apaletta3, line 70:
```suggestion
        throw ostk::core::error::RuntimeError("No states defined.");
```

test/ValidationFramework/Parser.hpp
```diff
@@ -0,0 +1,230 @@
...
+        const Pair<State, SatelliteSystem>& anInitialStateAndSatelliteSystem,
+        const Array<Shared<Dynamics>>& aDynamicsArray
+    );
+
+    /// @brief Create an array of evenly spaced instants at which to compare the output of two tools.
+    /// @param aDictionary A Dictionary.
+    /// @param anInitialInstant The initial instant.
+    /// @param aFinalInstant The final instant.
+    /// @return An array of instants.
+    static Array<Instant> CreateComparisonInstants(
```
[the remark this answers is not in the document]
apaletta3: This function is independent, only relies on the initial condition and reporting step (both of which are specificed in the yaml)

test/ValidationFramework/CrossValidator.cpp
apaletta3: hmmm what kinds of containers? you mean `CoordinateSubsets`? Cause 1) I don't think other containers are going to be different in terms of performance and 2) I think its quite clear to have an array of timesteps and an array of coordinate subsets within that array instead of putting stuff into a MatrixXd where you don't know which dimension means what. 

What do you think?

Derollez: Yeah i was thinking MatrixXd but that might be premature optimization

test/ValidationFramework/Validation.hpp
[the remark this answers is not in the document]
Derollez: resolving

test/ValidationFramework/Parser.hpp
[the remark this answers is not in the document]
Derollez: Yup, sorry saw that later in the review, should be closed earlier, resolving

test/ValidationFramework/CrossValidator.hpp
```diff
@@ -0,0 +1,228 @@
...
+using ostk::astro::dynamics::AtmosphericDrag;
+using ostk::astro::dynamics::Thruster;
+using ostk::astro::guidancelaw::ConstantThrust;
+using ostk::astro::guidancelaw::QLaw;
+using ostk::astro::solvers::FiniteDifferenceSolver;
+
+enum Quantity  // Enum values here correspond to index of each output in the csv file
+{
+    CARTESIAN_POSITION_GCRF,
+    CARTESIAN_VELOCITY_GCRF,
```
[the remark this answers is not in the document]
Derollez: Understood, thanks for clarifying

test/ValidationFramework/MissionSequence.cpp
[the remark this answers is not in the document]
Derollez: I just mean: let's add the explanation of that step: why we are recomputing at comparionInstants such that it's clear for other people later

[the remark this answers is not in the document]
Derollez: > There's no other way to obtain states at an arbitrary set of instants from a sequence solution currently.

That's exactly what I think should be explained with a tiny comment there :)

test/Validation/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+namespace ostk
+{
+namespace astro
+{
+namespace validation
+{
+
+MissionSequence::MissionSequence(const Dictionary& aDataTree)
+    : dataTree_(aDataTree),
+      initialStateAndSatelliteSystem_(Parser::CreateInitialStateAndSatelliteSystem(dataTree_)),
```
vishwa2710, line 14: I would separate these into two separate methods:
- Parser::ParseInitialState
- Parser::ParseSatelliteSystem

test/Validation/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (solvedStates_.isEmpty())
+    {
+        throw ostk::core::error::RuntimeError("No states defined.");
+    }
+    return solvedStates_;
+}
+
+Array<VectorXd> MissionSequence::compareResults(const Table& referenceData, const ToolComparison& aToolComparison) const
+{
+    const Array<QuantityComparison> quantityComparisons = aToolComparison.quantityComparisons;
```
vishwa2710, line 77:
```suggestion
    const Array<QuantityComparison>& quantityComparisons = aToolComparison.quantityComparisons;
```

test/Validation/MissionSequence.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    Array<Array<VectorXd>> referenceOutputs = CrossValidator::IngestOutputQuantities(referenceData, aToolComparison);
+
+    Array<VectorXd> allDeltasWithTool = Array<VectorXd>::Empty();
```
vishwa2710, line 81:
```suggestion
    Array<VectorXd> allDeltasWithTool = Array<VectorXd>::Empty();
    allDeltasWithTool.reserve(solvedStates_.getSize());
```

test/Validation/MissionSequence.hpp
```diff
@@ -0,0 +1,209 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Validation_MissionSequence__
+#define __OpenSpaceToolkit_Astrodynamics_Validation_MissionSequence__
+
+#include <Validation/CrossValidator.hpp>
+#include <Validation/Parser.hpp>
+
+#include <OpenSpaceToolkit/Core/Containers/Array.hpp>
```
vishwa2710, line 9: Same, all of these imports are not necessary.

test/Validation/MissionSequence.hpp
```diff
@@ -0,0 +1,209 @@
...
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace validation
+{
+
+using ostk::core::ctnr::Array;
```
vishwa2710, line 78: idem. All of these are not necessary.

test/Validation/MissionSequence.hpp
```diff
@@ -0,0 +1,209 @@
...
+using ostk::astro::guidancelaw::QLaw;
+using ostk::astro::solvers::FiniteDifferenceSolver;
+
+using ostk::astro::validation::Parser;
+using ostk::astro::validation::Quantity;
+using ostk::astro::validation::Tool;
+using ostk::astro::validation::ToolComparison;
+using ostk::astro::validation::QuantityComparison;
+
+/// @brief Hold all OSTk objects and data needed to define and run a "Mission Sequence".
```
vishwa2710, line 161:
```suggestion
/// @brief Holds the OSTk objects and data required to define and run a "Mission Sequence".
```

test/Validation/MissionSequence.hpp
```diff
@@ -0,0 +1,209 @@
...
+class MissionSequence
+{
+   public:
+    /// @brief Constructor
+    /// @param aDataTree A dictionary containing all the data needed to define and run a "Mission Sequence".
```
vishwa2710, line 166:
```suggestion
    /// @param aDataTree A dictionary containing all the data needed to define and initialize a "Mission Sequence".
```

test/Validation/MissionSequence.hpp
```diff
@@ -0,0 +1,209 @@
...
+    MissionSequence(const Dictionary& aDataTree);
+
+    ~MissionSequence();
```
vishwa2710, line 169:
```suggestion

    @brief Destructor
    ~MissionSequence();
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    Instant initialInstant = Instant::Undefined();
+    if (orbit["data"]["date"]["time-scale"].accessString() == "UTC")
+    {
+        initialInstant = Instant::DateTime(
+            DateTime::Parse(orbit["data"]["date"]["value"].accessString(), DateTime::Format::ISO8601), Scale::UTC
+        );
+    }
+    else
+    {
+        throw ostk::core::error::runtime::Wrong("Time scale not yet supported.");
```
vishwa2710, line 80:
```suggestion
        throw ostk::core::error::runtime::Wrong("Time scale");
```
OR
throw a ToBeImplemented, otherwise the error will look like:
`{Time scale not yet supported} is wrong.`

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    Shared<const Frame> frame = orbit["data"]["frame"].accessString() == "GCRF" ? Frame::GCRF() : Frame::Undefined();
```
vishwa2710, line 83:
```suggestion
    const Shared<const Frame> frame = orbit["data"]["frame"].accessString() == "GCRF" ? Frame::GCRF() : Frame::Undefined();
```
Do we want to support undefined frames? O_o Should we raise an error instead?

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            {
+                if (force["data"]["model"].accessString() == "EGM96")
+                {
+                    earthGravitationalModelType = EarthGravitationalModel::Type::EGM96;
+                    earthGravitationalModelDegree = force["data"]["degree"].accessInteger();
+                    earthGravitationalModelOrder = force["data"]["order"].accessInteger();
+                }
+                else
+                {
+                    throw ostk::core::error::runtime::Wrong("Earth gravitational model not recognized.");
```
vishwa2710, line 144:
```suggestion
                    throw ostk::core::error::runtime::Wrong("Earth gravitational");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            {
+                celestials.add(std::make_shared<Sun>(Sun::Default()));
+            }
+            else if (force["data"]["body"].accessString() == "MOON")
+            {
+                celestials.add(std::make_shared<Moon>(Moon::Default()));
+            }
+            else
+            {
+                throw ostk::core::error::runtime::Wrong("Planet not recognized.");
```
vishwa2710, line 157:
```suggestion
                throw ostk::core::error::runtime::Wrong("Planet");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+                {
+                    earthAtmosphericModelType = EarthAtmosphericModel::Type::NRLMSISE00;
+                }
+                else if (force["data"]["model"].accessString() == "EXPONENTIAL")
+                {
+                    earthAtmosphericModelType = EarthAtmosphericModel::Type::Exponential;
+                }
+                else
+                {
+                    throw ostk::core::error::runtime::Wrong("Earth atmospheric model not recognized.");
```
vishwa2710, line 174:
```suggestion
                    throw ostk::core::error::runtime::Wrong("Earth atmospheric model");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+                }
+            }
+            else
+            {
+                throw ostk::core::error::runtime::Wrong("Celestial Body not recognized.");
```
vishwa2710, line 179:
```suggestion
                throw ostk::core::error::runtime::Wrong("Celestial Body");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            }
+        }
+        else
+        {
+            throw ostk::core::error::runtime::Wrong("Force type not recognized.");
```
vishwa2710, line 184:
```suggestion
            throw ostk::core::error::runtime::Wrong("Force type");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    else if (aDictionary["data"]["sequence"]["propagator"]["type"].accessString() == "RUNGE_KUTTA_4")
+    {
+        numericalSolver = NumericalSolver::FixedStepSize(
+            NumericalSolver::StepperType::RungeKutta4,
+            aDictionary["data"]["sequence"]["propagator"]["data"]["step"].accessReal()
+        );
+    }
+    else
+    {
+        throw ostk::core::error::runtime::Wrong("Propagator type not recognized.");
```
vishwa2710, line 229:
```suggestion
        throw ostk::core::error::runtime::Wrong("Propagator type");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            segment.accessDictionary(), anInitialStateAndSatelliteSystem, aDynamicsArray, numericalSolver
+        ));
+    }
+
+    return {
+        segments,
+        numericalSolver,
+        Array<Shared<Dynamics>>::Empty(),
+        Duration::Seconds(aDictionary["data"]["sequence"]["max-duration"].accessReal()),
+        0
```
vishwa2710, line 245:
```suggestion
        0,
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+                        segmentDictionary["data"]["stop-condition"]["data"]["value"].accessString(),
+                        DateTime::Format::ISO8601
+                    ),
+                    Scale::UTC
+                )
+            ));
+        }
+        else
+        {
+            throw ostk::core::error::runtime::Wrong("Time scale not recognized.");
```
vishwa2710, line 315:
```suggestion
            throw ostk::core::error::runtime::Wrong("Time scale");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+        }
+    }
+    else
+    {
+        throw ostk::core::error::runtime::Wrong("Segment stop condition type not recognized.");
```
vishwa2710, line 320:
```suggestion
        throw ostk::core::error::runtime::Wrong("Segment event condition type.");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+    if (segmentDictionary["type"].accessString() == "COAST")
+    {
+        return Segment::Coast("coast", conditionSPtr, aDynamicsArray, aNumericalSolver);
+    }
+    else if (segmentDictionary["type"].accessString() == "CONSTANT_DIRECTION_MANEUVER")
+    {
+        if ((segmentDictionary["data"]["attitude"]["type"].accessString() != "LOF") ||
+            (segmentDictionary["data"]["attitude"]["data"]["lof"].accessString() != "VNC"))
+        {
+            throw ostk::core::error::runtime::Wrong("Attitude type or local orbital type not supported.");
```
vishwa2710, line 333:
```suggestion
            throw ostk::core::error::runtime::Wrong("Attitude type or local orbital type");
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        Shared<const Frame> frame = segmentDictionary["data"]["attitude"]["data"]["parent"].accessString() == "GCRF"
```
vishwa2710, line 336:
```suggestion
        const Shared<const Frame> frame = segmentDictionary["data"]["attitude"]["data"]["parent"].accessString() == "GCRF"
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+                                      : Frame::Undefined();
+
+        // Create thruster dynamics
+        const Vector3d maneuverVector = {
+            segmentDictionary["data"]["attitude"]["data"]["direction"].accessArray()[0].accessReal(),
+            segmentDictionary["data"]["attitude"]["data"]["direction"].accessArray()[1].accessReal(),
+            segmentDictionary["data"]["attitude"]["data"]["direction"].accessArray()[2].accessReal()
+        };
+
+        Shared<ConstantThrust> constantThrustSPtr = std::make_shared<ConstantThrust>(
```
vishwa2710, line 347:
```suggestion
        const Shared<ConstantThrust> constantThrustSPtr = std::make_shared<ConstantThrust>(
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            ConstantThrust(LocalOrbitalFrameDirection(maneuverVector, LocalOrbitalFrameFactory::VNC(frame)))
+        );
+        Shared<Thruster> aThrusterDynamicSPtr =
```
vishwa2710, line 350:
```suggestion
        const Shared<Thruster> aThrusterDynamicSPtr =
```

test/Validation/Parser.cpp
```diff
@@ -0,0 +1,363 @@
...
+            std::make_shared<Thruster>(Thruster(anInitialStateAndSatelliteSystem.second, constantThrustSPtr));
+
+        return Segment::Maneuver("maneuver", conditionSPtr, aThrusterDynamicSPtr, aDynamicsArray, aNumericalSolver);
+    }
+    else
+    {
+        throw ostk::core::error::runtime::Wrong("Segment type not recognized.");
```
vishwa2710, line 357:
```suggestion
        throw ostk::core::error::runtime::Wrong("Segment type");
```

test/Validation/Parser.hpp
```diff
@@ -0,0 +1,230 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Validation_Parser__
+#define __OpenSpaceToolkit_Astrodynamics_Validation_Parser__
+
+#include <Validation/CrossValidator.hpp>
+
+#include <OpenSpaceToolkit/Core/Containers/Array.hpp>
```
vishwa2710, line 8: idem, not all imports are necessary.

test/Validation/Parser.hpp
```diff
@@ -0,0 +1,230 @@
...
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp>
+
+namespace ostk
+{
+namespace astro
+{
+namespace validation
+{
+
+using ostk::core::ctnr::Array;
```
vishwa2710, line 82: idem, not all namespaces are necessary

vishwa2710 (commented):
First pass. Looking great!
Is there a reason we don't have any unit tests for these?
IMO it would be worth adding tests, it would be kind of sad to find a bug in a validation framework as it is untested.

CMakeLists.txt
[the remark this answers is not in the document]
apaletta3: done

CMakeLists.txt
```diff
@@ -16,7 +16,8 @@ SET (PROJECT_VENDOR_URL "open-space-collective.org")
 
 OPTION (BUILD_SHARED_LIBRARY "Build shared library." ON)
 OPTION (BUILD_STATIC_LIBRARY "Build static library." OFF)
-OPTION (BUILD_UNIT_TESTS "Build tests" ON)
+OPTION (BUILD_UNIT_TESTS "Build unit tests" ON)
+OPTION (BUILD_VALIDATION_TESTS "Build validation tests" ON)
```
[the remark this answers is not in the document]
apaletta3: yeah fair

test/Validation/CrossValidator.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include <OpenSpaceToolkit/Physics/Environment/Ephemerides/Analytical.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Earth.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Magnetic/Earth.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Object.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Earth.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Moon.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Objects/CelestialBodies/Sun.hpp>
+#include <OpenSpaceToolkit/Physics/Time/DateTime.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Duration.hpp>
+#include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
```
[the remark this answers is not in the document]
apaletta3: Sure, but it shouldn't increase compile time should it? cause of the `idnef` header guard in each .hpp file?

test/Validation/MissionSequence.cpp
apaletta3: It would be cleaner that way, but the reason I combined them is that they constructing those objects is quite related and is more easily done together

test/Validation/Parser.cpp
apaletta3: I figured it would error if the frame was undefined and you tried to use it later on. But maybe I should be more explicit in throwing an error here.

README.md
```diff
@@ -213,12 +213,24 @@ Or to run them manually:
...
+
+OSTk has a cross-validation framework built into it, which allows us to compare its end-to-end propagation accuracy of a full "mission sequence" to other flight qualified tools like GMAT and Orekit.
+
+A standardized input format (a yaml file) to define the particular mission sequence scenario is used, and is tooling agnostic.
+
+The validation framework takes this yaml scenario definition (inside the /test/Validation/data/scenarios folder) and runs the scenario using OSTk. Next, it compares the outputs (things like the position, velocity, mass, acceleration of the spacecraft) at each timestep during the scenario and checks if they are within a certain tolerance of those same quantities generated by GMAT or Orekit in another repo and read in as a .csv file.
+
+Each scenario, the tool it is being compared against, what output quantities are being compared and to what tolerance are defined in the `parameters` variable inside the `Framework.validation.cpp` file, and can be modified or have new scenarios and tolerances appended to them.
+
+The validation tests can be run with `ostk-validate` from within the dev container, or `make validation` as a standalone.
```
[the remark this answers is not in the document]
Derollez: I agree. Users should be capable of adding a scenario and understanding the standard input definition easily

README.md
```diff
@@ -213,12 +213,24 @@ Or to run them manually:
...
 
 Benchmarks are pushed to the [GitHub pages](https://open-space-collective.github.io/open-space-toolkit-astrodynamics/dev/bench/).
 
+### Validation
+
+OSTk has a cross-validation framework built into it, which allows us to compare its end-to-end propagation accuracy of a full "mission sequence" to other flight qualified tools like GMAT and Orekit.
+
+A standardized input format (a yaml file) to define the particular mission sequence scenario is used, and is tooling agnostic.
+
+The validation framework takes this yaml scenario definition (inside the /test/Validation/data/scenarios folder) and runs the scenario using OSTk. Next, it compares the outputs (things like the position, velocity, mass, acceleration of the spacecraft) at each timestep during the scenario and checks if they are within a certain tolerance of those same quantities generated by GMAT or Orekit in another repo and read in as a .csv file.
```
Derollez, line 228:
```suggestion
The validation framework takes this yaml scenario definition (inside the `/test/Validation/data/scenarios` folder) and runs the scenario using OSTk. Next, it compares the outputs (quantities like position, velocity, mass, acceleration of the spacecraft) at each timestep during the scenario and checks if they are within a certain tolerance of those same quantities generated by GMAT or Orekit in another repo and read in as a `.csv` file.
```

README.md
```diff
@@ -213,12 +213,24 @@ Or to run them manually:
...
+
+Each scenario, the tool it is being compared against, what output quantities are being compared and to what tolerance are defined in the `parameters` variable inside the `Framework.validation.cpp` file, and can be modified or have new scenarios and tolerances appended to them.
```
Derollez, line 230: Missing words? 🤔

Derollez (commented):
@apaletta3 A few comments on the README to address and improve + imports comment from @vishwa2710 👍

README.md
apaletta3: lol yes, wrote it while I was half asleep, can't english

test/Validation/MissionSequence.cpp
apaletta3: actually it is cleaner that way I made the change

test/Validation/CrossValidator.hpp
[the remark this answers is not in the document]
apaletta3: @vishwa2710 ?

test/ValidationFramework/Parser.cpp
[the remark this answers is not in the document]
apaletta3: Split them up again

validation/OpenSpaceToolkit/Astrodynamics/data/test/test_parser_create_satellite_system.yaml
```diff
@@ -0,0 +1,19 @@
+# Test create satellite system function
+success-no-prop-system:
+  data:
+    spacecraft:
+      mass: 100.0
+      drag-cross-section: 1.0
+      drag-coefficient: 2.2
+
```
Derollez, line 8: (very nit) Some yaml have extra lines in between test cases when others don't. No issues when parsing?

validation/OpenSpaceToolkit/Astrodynamics/CrossValidator.test.cpp
```diff
@@ -0,0 +1,207 @@
...
+    void SetUp() override
+    {
+        VectorXd initialCoordinates(7);
+        initialCoordinates << 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 10.0;
+        defaultState_ = {Instant::J2000(), initialCoordinates, Frame::GCRF(), coordinatesBrokerSPtr_};
+
+        referenceCoordinates_ = {
+            initialCoordinates.segment(0, 3), initialCoordinates.segment(3, 3), initialCoordinates.segment(6, 1)
+        };
+    }
```
Derollez, line 95: (nit) To align with other test classes and classes in OSTk in general, the methods are usually added above the fields of the classes. Apply elsewhere

validation/OpenSpaceToolkit/Astrodynamics/Parser.test.cpp
```diff
@@ -0,0 +1,451 @@
...
+            ostk::core::error::runtime::Wrong
+        );
+    }
+
+    {
+        EXPECT_THROW(
+            Parser::CreateInitialState(dataTree["failure-frame"].accessDictionary(), defaultSatelliteSystem_),
+            ostk::core::error::runtime::Wrong
+        );
+    }
```
Derollez, line 166: What about performing the steps:
- accessDictionary()
- CreateInitialState
separately, i.e. the accessDictionary() should be done before and not be included in the EXPECT_THROW to reduce the error to be found in that step only

Also,if wanted, we can try/catch the exact error expected for the THROW

Derollez (approved):
Some small (nits), LGTM otherwise, let's ship it!

validation/OpenSpaceToolkit/Astrodynamics/data/test/test_parser_create_satellite_system.yaml
apaletta3: Nope, no issues when parsing because its just searching for the keys when parsing. But I will adjust this for consistency

validation/OpenSpaceToolkit/Astrodynamics/Parser.test.cpp
apaletta3: Good point, addressed this and applied everywhere

PR #300, opened 2023-12-11
chore: improve dev container with non-root access 

apaletta3:
This PR adds a nicer interface for our local dev container as non-root user and eliminates volume sharing permissions issues associated with build folders having root permissions when you exit the dev container.

This MR depends on this one https://github.com/open-space-collective/open-space-toolkit/pull/81 to be merged first and a new version of the ostk-base image to be pushed to docker hub.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,44 +1,45 @@
 {
     "name": "Open Space Toolkit Astrodynamics",
-    "mounts": [
-        "source=ostk_astrodynamics_vscode_extensions,target=/home/vscode/.vscode-server/extensions,type=volume"
-    ],
+    "image": "openspacecollective/open-space-toolkit-astrodynamics-development:latest",
+    "initializeCommand": "make build-development-image",
+    "workspaceMount": "source=${localWorkspaceFolder},target=/app,type=bind",
+    "workspaceFolder": "/app",
+    "containerUser": "vscode",
+    "updateRemoteUserUID": true,
     "runArgs": [
         "--name",
         "open-space-toolkit-astrodynamics-dev",
         "--privileged"
     ],
-    "containerEnv": {
-        "PACKAGE_DEPLOY_TOKEN": "${localEnv:PACKAGE_DEPLOY_TOKEN}"
-    },
-    "workspaceMount": "source=${localWorkspaceFolder},target=/app,type=bind",
-    "workspaceFolder": "/app",
-    "updateRemoteUserUID": true,
-    "image": "openspacecollective/open-space-toolkit-astrodynamics-development:latest",
-    "initializeCommand": "make build-development-image",
+    "mounts": [
+        "source=ostk_astrodynamics_vscode_extensions,target=/home/vscode/.vscode-server/extensions,type=volume",
+        "source=${localEnv:HOME}/.ssh,target=/home/vscode/.ssh,type=bind,readonly"
+    ],
     "shutdownAction": "stopContainer",
     "customizations": {
         "vscode": {
-            "settings": {
-                "C_Cpp.clang_format_style": "file:/app/thirdparty/clang/.clang-format",
-                "terminal.integrated.cwd": "/app/build",
-                "terminal.integrated.profiles.linux": {
-                    "bash": {
-                        "path": "/bin/bash",
-                        "icon": "terminal-bash"
-                    }
-                },
-                "python.analysis.diagnosticSeverityOverrides": {
-                    "reportMissingModuleSource": "none"
-                }
-            },
             "extensions": [
                 "ms-vscode.cpptools",
                 "ms-python.python",
                 "ms-python.black-formatter",
                 "mhutchie.git-graph",
-                "ms-vscode.cpptools-extension-pack"
-            ]
+                "ms-vscode.cpptools-extension-pack",
+                "mechatroner.rainbow-csv"
+            ],
+            "settings": {
+                "C_Cpp.clang_format_style": "file:/app/thirdparty/clang/.clang-format",
+                "terminal.integrated.cwd": "/app/build",
+                "terminal.integrated.profile.linux": {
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,44 +1,47 @@
...
     "shutdownAction": "stopContainer",
     "customizations": {
         "vscode": {
+            "extensions": [
+                "ms-vscode.cpptools",
+                "ms-python.python",
+                "ms-python.black-formatter",
+                "mhutchie.git-graph",
+                "ms-vscode.cpptools-extension-pack",
+                "mechatroner.rainbow-csv",
```
kyle-cochran, line 28: Did you mean to add this csv extension? Was this something that slipped in from your local config?

.github/workflows/build-test.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     name: Build & Test
     needs:
       - prepare-environment
-    uses: open-space-collective/open-space-toolkit/.github/workflows/build-test.yml@main
+    uses: open-space-collective/open-space-toolkit/.github/workflows/build-test.yml@chore/improve-dev-container
```
kyle-cochran, line 47:
```suggestion
    uses: open-space-collective/open-space-toolkit/.github/workflows/build-test.yml@main
```
TODO: remove before release

Makefile
```diff
@@ -19,6 +19,8 @@ jupyter_python_version_without_dot := $(shell echo $(jupyter_python_version) | s
 jupyter_notebook_image_repository := jupyter/scipy-notebook:x86_64-python-$(jupyter_python_version).3
 extract_python_package_version := $(shell echo $(project_version) | sed 's/-/./' | sed 's/-.*//')
 
+dev_username := vscode
```
kyle-cochran, line 22: I know we all use vscode, but I would say that we want to keep the dev tools independent of an editor since this is an open source project.

docker/development/Dockerfile
```diff
@@ -172,6 +174,37 @@ ARG VERSION
...
+
+RUN sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.sh)" \
+ && git clone https://github.com/bhilburn/powerlevel9k.git /home/${USERNAME}/.oh-my-zsh/custom/themes/powerlevel9k \
+ && git clone https://github.com/zsh-users/zsh-autosuggestions /home/${USERNAME}/.oh-my-zsh/custom/plugins/zsh-autosuggestions \
+ && git clone https://github.com/zsh-users/zsh-syntax-highlighting.git /home/${USERNAME}/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting \
+ && mkdir -p /home/${USERNAME}/.vscode-server/extensions /home/${USERNAME}/.vscode-server-insiders/extensions
+
+## Configure environment
 
-CMD [ "/bin/bash" ]
```
kyle-cochran, line 177: Why remove the default hook?

.devcontainer/devcontainer.json
apaletta3: Yes I did, its quite useful for visualizing CSVs because it colors each column differently, and since there are plenty of validation files in the .csv format I figure its a nice tool to have.

Makefile
apaletta3: This actually doesn't have anything to do with VSCode itself, its just the name of the non-root user that is configured in the docker image for the `development-image-non-root` at build time. It just needs to match what is specified in line-189 of the development Dockerfile.

But you bring up a good point, I'll change the name to something more generic like `developer`.

docker/development/Dockerfile
apaletta3: I was thinking just because all our make commands specify the shell to be used at runtime. But maybe there is a good reason to keep it.

kyle-cochran: It's nice to have when testing the container build sometimes. You can do `docker run open-space-collective/open-space-toolkit-astrodynamics:latest` without any extra args and it will throw you into a shell. Not required, but I think it's a nice-to-have

Makefile
```diff
@@ -229,18 +248,21 @@ build-packages-python-standalone: ## Build Python packages (standalone)
...
 		--rm \
-		--privileged \
 		--volume="$(CURDIR):/app:delegated" \
+		--volume="$(HOME)/.ssh:/home/$(dev_username)/.ssh:ro" \
+		--volume="$(HOME)/.gitconfig:/home/$(dev_username)/.gitconfig:ro" \
 		--workdir=/app/build \
-		$(docker_development_image_repository):$(docker_image_version) \
-		/bin/bash
+		$(docker_development_image_repository)-non-root:$(docker_image_version) \
+		/bin/zsh
```
kyle-cochran, line 265: We should probably be consistent with the shell used across the dockerfile. This just opens an interactive shell, but it's possible to install something with `bash` and it not be visible to `zsh` because they have different instantiations of the `PATH` variable.

I'm fine with using zsh, but if we go that route, maybe can put a `SHELL_PATH` variable at the top of the file and reference that everywhere

Makefile
```diff
@@ -250,7 +272,9 @@ start-development-link: build-development-image ## Start linked development envi
 
 	@ echo "Starting development environment (linked)..."
 
+	@ mkdir -p $(CURDIR)/build
```
kyle-cochran, line 275: Does the build folder add get removed somewhere? Also, does this fail when the folder already exists (mkdir by default does)?

Makefile
```diff
@@ -493,7 +515,6 @@ test-unit-python-standalone: ## Run Python unit tests (standalone)
 		--volume="$(CURDIR):/app:delegated" \
 		--volume="/app/build" \
 		--workdir=/app/build \
-		--entrypoint="" \
```
kyle-cochran, line 496: ?

kyle-cochran (approved):

PR #302, opened 2023-12-14
fix: Satellite System Builder test import

phc1990:


apaletta3 (approved):
Yup all of this makes sense! Looks good to me

PR #308, opened 2024-01-02
test: update past data expectation with wiggle room for CSSI rollover

kyle-cochran:
The OSTk test image takes a bit to include the CSSI space weather rollover. Extend the failure look point a couple months to give it a little wiggle room.


apaletta3 (approved):
I'm confused, could you explain a bit more about this fix works in the MR description?
Is this a permanent fix or a temporary fix?

Derollez (approved):
Awesome, thanks 🙏

PR #309, opened 2024-01-02
chore: add sequence solution default constructor bindings

Derollez:


bindings/python/test/trajectory/test_sequence.py
```diff
@@ -318,6 +318,22 @@ def sequence(
     )
 
 
+@pytest.fixture
+def segment_solution(
+    dynamics: list,
+    state: State,
+):
+    return Segment.Solution(
+        name="A Segment",
```
kyle-cochran, line 327:
```suggestion
        name="A Segment Solution",
```

nit: Just checking, is this missing?

kyle-cochran (approved):

PR #318, opened 2024-01-14
feat: add orbit pass computation logic

vishwa2710:
With this we can do some precise Orbit pass time computations and then leverage that in the Orbit service to alleviate some of the edge case bugs we've noticed.


Derollez (approved):
LGTM, really nice refactor

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Pass.cpp
```diff
@@ -14,11 +14,25 @@ namespace trajectory
...
+)
     : type_(aType),
       revolutionNumber_(aRevolutionNumber),
-      interval_(anInterval)
+      interval_(anInterval),
+      instantAtDescendingNode_(anInstantAtDescendingNode),
+      instantAtNorthPoint_(anInstantAtNorthPoint),
+      instantAtSouthPoint_(anInstantAtSouthPoint)
 {
+    if ((aType == Pass::Type::Complete) && (!anInstantAtNorthPoint.isDefined() || !anInstantAtSouthPoint.isDefined()))
```
robinpdm, line 32: Should we check that `anInterval` and `anInstantAtDescendingNode` are defined too?

PR #320, opened 2024-01-16
fix: set instant from inputs in NumericalSolver::integrateTime

vishwa2710:


kyle-cochran (changes requested):
There are a few more places where we could substitute the input instant instead of the calculated one:
- https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/NumericalSolver.cpp#L120
- https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/NumericalSolver.cpp#L258

I think it makes sense to do this because the expectation for these overloads is the same as the expectation for the other version of this function: If the Solver is provided with an Instant to propagate to, then the output state will be at an instant that exactly matches the input instant.

vishwa2710:
@kyle-cochran I believe L120 has already been addressed?
And L258 cannot be done via setting the endInstant as it is a condition based solve, so we do not know when the end Instant is. The provided end Instant is a maximum propagation time that is used to prevent infinite loops.

kyle-cochran (approved):

PR #321, opened 2024-01-17
style: change dynamics contributions observing to an array of vectorxds

apaletta3:
I've refactored the way that dynamics contributions for the duration of the Segment are returned to the user, as an Array<VectorXd> instead of a MatrixXd since I think its more clear for someone who wants to index through them later on. 

This is intended to go along with the dynamics refactoring and observing dynamics contributions discussion.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
@@ -105,7 +105,7 @@ Mass Segment::Solution::computeDeltaMass() const
     return Mass::Kilograms(getInitialMass().inKilograms() - getFinalMass().inKilograms());
 }
 
-MatrixXd Segment::Solution::getDynamicsContribution(
+Array<VectorXd> Segment::Solution::getDynamicsContributions(
     const Shared<Dynamics>& aDynamicsSPtr,
     const Shared<const Frame>& aFrameSPtr,
     const Array<Shared<const CoordinatesSubset>>& aCoordinatesSubsetSPtrArray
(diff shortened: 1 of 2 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -151,57 +151,68 @@ MatrixXd Segment::Solution::getDynamicsContribution(
...
+    const Size numberOfstates = this->states.getSize();
+    for (Index stateIndex = 0; stateIndex < numberOfstates; ++stateIndex)
+    {
+        // Create map for each state and fill it with the dynamics contributions
+        Map<Shared<Dynamics>, VectorXd> dynamicsContributionMapAtState = Map<Shared<Dynamics>, VectorXd>();
+        for (const Shared<Dynamics>& aDynamicsSPtr : this->dynamics)
+        {
+            dynamicsContributionMapAtState.emplace(aDynamicsSPtr, dynamicsContributionsMap[aDynamicsSPtr][stateIndex]);
+        }
+        dynamicsContributionsMaps.add(dynamicsContributionMapAtState);
```
vishwa2710, line 212: I feel like this will duplicate so much data in terms of repeated headers.
Can you provide an example that this data format is cleaner?
Currently:
```
{
    CartesianVelocity : [[0,1,2], [2,3,4], [4,5,6]...]
    Acceleration: [[5,6,7], [6,7,8], [7,8,9]]
}
```
vs.
- {CartesianVelocity: [0,1,2], Acceleration: [4,5,6]}
- {CartesianVelocity: [0,1,2], Acceleration: [4,5,6]}
- {CartesianVelocity: [0,1,2], Acceleration: [4,5,6]}
...

vishwa2710 (commented):
A bit confused by the change you want to drive here, if you can provide some context that would be great.

vishwa2710:
I would change the heading from `style` to `refactor`

apaletta3:
Not relevant anymore.

PR #324, opened 2024-01-20
chore: add typing, update inputs for utilities

vishwa2710:


kyle-cochran (approved):

apaletta3 (approved):
LGTM! Maybe also add the types contained within each return list for clarity like so `list[float]`

PR #325, opened 2024-01-22
feat: return 0.0 when outside model bounds in TabulatedDynamics::computeContribution

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -130,6 +130,11 @@ VectorXd Tabulated::computeContribution(
         throw ostk::core::error::runtime::Wrong("Frame");
     }
 
+    if (anInstant < instants_.accessFirst() || anInstant > instants_.accessLast())
+    {
+        return VectorXd::Zero(interpolators_.getSize());
+    }
+
     const double epoch = (anInstant - instants_.accessFirst()).inSeconds();
```
apaletta3, line 138:
```suggestion
    const Real epoch = (anInstant - instants_.accessFirst()).inSeconds();
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -214,15 +215,26 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Getters)
...
+            );
+        }
+
+        // Outside bounds
+        {
+            EXPECT_EQ(
+                tabulated.computeContribution(defaultInstants_[0] - Duration::Seconds(1.0), x, Frame::GCRF()),
+                VectorXd::Zero(3)
+            );
+        }
```
apaletta3, line 237:
```suggestion
        {
            EXPECT_EQ(
                tabulated.computeContribution(defaultInstants_[0] - Duration::Seconds(1.0), x, Frame::GCRF()),
                VectorXd::Zero(3)
            );
        }
        
        {
            EXPECT_EQ(
                tabulated.computeContribution(defaultInstants_[defaultInstants_.getSize() -1] + Duration::Seconds(1.0), x, Frame::GCRF()),
                VectorXd::Zero(3)
            );
        }
```

apaletta3 (approved):
Looks great!

Couple things:
- Can you remove the `[[maybe_unused]]` for the `aFrameSPtr` argument on line 124 since it is now actually used?
- Is there not already an out of data set guard in the interpolator? It might be worth adding at the interpolator level one so that our interpolation doesn't unwittingly turn into extrapolation

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
vishwa2710: The evaluator accepts a `double` https://github.com/open-space-collective/open-space-toolkit-mathematics/blob/93d7dc56e59d906c9c5527e828ce36f2f3a2225f/src/OpenSpaceToolkit/Mathematics/CurveFitting/Interpolator.cpp#L18 so it think it's better to be explicit.

vishwa2710:
> Can you remove the [[maybe_unused]] for the aFrameSPtr argument on line 124 since it is now actually used?
Agreed!

> Is there not already an out of data set guard in the interpolator? It might be worth adding at the interpolator level one so that our interpolation doesn't unwittingly turn into extrapolation

There is! However it will throw an error. In this case we want to just return 0.0 instead of raising an error.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: Ok to keep it like this for now, but in the future perhaps better to align OSTk with using `Real` whenever possible?

apaletta3 (approved):
Dope. Good to resolve all changes and merge!

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
vishwa2710: Fair, I think we can update the interface there to accept a Real and then convert to a double when interfacing with boost.

PR #326, opened 2024-01-23
fix: update pass computation epoch to improve accuracy

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
```diff
@@ -203,7 +203,10 @@ class Orbit : public Trajectory
     String generateFrameName(const Orbit::FrameType& aFrameType) const;
 
     static Instant GetCrossingInstant(
```
kyle-cochran, line 205: nit: should the epoch be optional here?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
```diff
@@ -203,7 +203,10 @@ class Orbit : public Trajectory
     String generateFrameName(const Orbit::FrameType& aFrameType) const;
```
kyle-cochran, line 204:
```suggestion

    /// @brief Find the Instant at which the return value of `getValue` crosses zero.
    ///
    ///              Use a bisection search to find the Instant between `previousInstant` and `nextInstant` at which the return value of `getValue` crosses zero. 
    ///
    /// @param anEpoch An orbit epoch from which to measure
    /// @param previousInstant lower bound of search
    /// @param getValue function that receives a duration [s] from `anEpoch` and returns a real value
    /// @return Instant
```

^nit: it's not super obvious what the extra Instant is for

kyle-cochran (approved):

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
vishwa2710: It's required to offset the instants to convert to double. So it's required, I don't want to have a fallback to J2000 or something due to the precision issues.

kyle-cochran: Okay. I missed that it was a private function. Disregard

PR #327, opened 2024-01-23
feat: update pass constructor to allow partial passes from start/end

vishwa2710:
- [x] Add tests to cover the remaining edge cases

The main lift here is having explicit definitions for all the parameters for a Pass. This was necessary as a Pass can be partial from both sides:
```
|        ----------------|
|--------------          |
```
Therefore we want explicit definitions for the ascending node, north point, descending node, south point and the pass break.
Where pass break is the ascending node for the _next_ pass.

This is also necessary to properly define whether the pass is complete or partial.

Note: While there are some breaking changes here, since this function was recently released and is not used anywhere (yet) we don't necessarily have to do a major version increase here.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
@@ -1040,7 +1040,9 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
 
     if (aStateArray.getSize() < 2)
     {
-        throw ostk::core::error::Runtime("Greater than 2 states required to compute passes: {}", aStateArray.getSize());
+        throw ostk::core::error::RuntimeError(
+            "Greater than 2 states required to compute passes: {}", aStateArray.getSize()
+        );
     }
 
     for (Index i = 1; i < aStateArray.getSize(); ++i)
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
@@ -53,6 +53,8 @@ using ostk::astro::RootSolver;
 static const Derived::Unit GravitationalParameterSIUnit =
     Derived::Unit::GravitationalParameter(Length::Unit::Meter, ostk::physics::units::Time::Unit::Second);
 
+static const Real epsilon = 1e-6;
+
 static const RootSolver rootSolver = RootSolver::Default();
 
 Orbit::Orbit(const orbit::Model& aModel, const Shared<const Celestial>& aCelestialObjectSPtr)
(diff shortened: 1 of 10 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -245,11 +247,12 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber) const
             Integer currentRevolutionNumber = currentPass.isDefined() ? currentPass.getRevolutionNumber()
                                                                       : this->modelPtr_->getRevolutionNumberAtEpoch();
             Instant previousInstant = currentPass.isDefined()
-                                        ? (currentPass.getInterval().accessEnd() + Duration::Microseconds(1.0))
+                                        ? (currentPass.accessInstantAtPassBreak() + Duration::Microseconds(1.0))
```
kyle-cochran, line 250: Sanity check:

From our conversation, I thought I remembered that `passBreak` is the instant immediately after the spacecraft crosses the ascending node. If that's true, would this logic make the next pass start one step after the actual start of the pass?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1034,13 +1036,12 @@ Map<Index, Pass> Orbit::GeneratePassMap(const Array<State>& aStateArray, const I
...
+    Array<Pair<Index, Pass>> passMap;
 
-    Integer revolutionNumber = anInitialRevolutionNumber;
-    Instant previousPassEndInstant = Instant::Undefined();
-
-    Index currentIndex = 0;
-    State const* previousStatePtr = nullptr;
+    if (aStateArray.getSize() < 2)
+    {
+        return passMap;
```
kyle-cochran, line 1043:
```suggestion
        throw ostk::core::error::Runtime("Too few states to compute passes. n_states = {}", aStateArray.getSize());
```

Returning an uninitialized value seems kinda sketch. Maybe good to just error here instead?

kyle-cochran: Especially since that's the way we handle states not being in chrono order on line 1050 below

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Pass.cpp
```diff
@@ -15,23 +15,70 @@ namespace orbit
...
+    if (!anInstantAtNorthPoint.isDefined() && !anInstantAtSouthPoint.isDefined() &&
+        !anInstantAtDescendingNode.isDefined() && !anInstantAtAscendingNode.isDefined() &&
+        !anInstantAtPassBreak.isDefined() && !revolutionNumber_.isDefined())
+    {
+        type_ = Pass::Type::Undefined;
+    }
+    else if (!anInstantAtNorthPoint.isDefined() || !anInstantAtSouthPoint.isDefined() ||
+        !anInstantAtDescendingNode.isDefined() || !anInstantAtAscendingNode.isDefined() ||
+        !anInstantAtPassBreak.isDefined())
+    {
```
kyle-cochran, line 41:
```suggestion
    else if ((!anInstantAtNorthPoint.isDefined() || !anInstantAtSouthPoint.isDefined() ||
        !anInstantAtDescendingNode.isDefined() || !anInstantAtAscendingNode.isDefined() ||
        !anInstantAtPassBreak.isDefined()) && revolutionNumber_.isDefined())
    {
```

Because revolution number always has to be defined right?

Otherwise i think this allows revolution number to be undefined and have a pass with a `Partial` type

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: `passBreak` is the ascending node (within accuracy). Since `passBreak` is at ascending node, We want to start propagating from a microsecond after so that we don't start before the ascending node.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: Yeah I went back and forth on this. I think raising an error is fair, users can handle it.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Pass.cpp
vishwa2710: Yeah good catch!

kyle-cochran (approved):

PR #330, opened 2024-01-29
feat: add new validation scenarios

apaletta3:
This MR builds on top of the one refactoring all the validation tests by adding a bunch more scenarios to validation framework.

The types of scenarios I included are meant to be more or less "linearly independent" so that if in the future if we break something we will have a pretty good idea of the where the error is coming from.

The naming convention I followed is:
- 001-009 (force-model) is reserved for validation tests trying out different force models and combinations of force models, so that if something in the dynamics contribution or ostk-physics breaks, it shows up here
- 010-019 (thruster) is reserved for validation tests trying out different thrusting orientations and parameters associated with the thruster and spacecraft, so that if something breaks in the local orbital frames, the thruster/spacecraft classes, or the dynamics classes it shows up here
- 020-029 (sequence) is reserved for complex sequences of multiple burn and coast arcs, so that if anything breaks in the segment and sequence classes it should show up here


phc1990 (approved):
LGTM

Derollez (approved):
Nice! In the future, we might have another discussion about scenario naming/description as the complexity will grow with the number of cases/scenarios but that's for later!

PR #333, opened 2024-02-02
feat: add maneuver class

apaletta3:
This MR will add a `Maneuver` class, who's goal is to essentially be an interface that stores a maneuver profile as well as metadata associated with that maneuver, and provides utility methods to compute extra metrics about the maneuver. Connecting a Maneuver with a `SegmentSolution` and `Propagator` happens in [this seperate MR](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/340) currently on top of this one.

Progress:

- [x] Create prototype of Maneuver class with API and mem vars and interactions with other OSTk objects
- [x] Get design reviewed and iterate on it
- [x] Merge [this MR](https://github.com/open-space-collective/open-space-toolkit-mathematics/pull/110) in OSTk math and update corresponding `InterpolationType`
- [x] Add all internal logic in Maneuver 
- [x] Add bindings
- [x] Add unit tests
- [x] Remove calculateXAt methods
- [x] Add convenient operators like == and !=
- [x] Look into not accepting Local orbital frame contribution profiles


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -105,6 +105,11 @@ Mass Segment::Solution::computeDeltaMass() const
     return Mass::Kilograms(getInitialMass().inKilograms() - getFinalMass().inKilograms());
 }
 
+Array<Maneuver> Segment::Solution::getManeuvers() const
+{
+    // Construct maneuvers from the contributions inside of a maneuvering segment
```
apaletta3, line 110: Going to use the logic from dynamics_interpreter from the guidance service here to split thrusting accelerations produced during a maneuvering segment into multiple maneuvers

apaletta3 (commented):
Let me know what you guys think of this rough prototype. 

What I've found while prototyping this is that a Maneuver class should be very specific to a maneuver acceleration profile and a mass flow rate (assumed to have one mass flow rate for the entire maneuver for now), without anything else like a propulsion system or other stuff because I think the maneuver class should be very pure since it should be able to be reconstructed with no specific information about the prop system or spacecraft etc, so things like the deltaV and deltaMass can be "estimated" from the length of the maneuver and accelerations occuring during the maneuver.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+    Pair<Array<Instant>, Array<Vector3d>> getInstantsAndAccelerationProfile();
+    Real getMassFlowRate();
+    Instant getStartInstant();
+    Instant getEndInstant();
+    Duration getDuration();
+
+    Real estimateDeltaV();
+    Mass estimateDeltaMass();
+
+    Vector3d estimateThrustAccelerationAt(
```
vishwa2710, line 108:
```suggestion
    Vector3d calculateAccelerationAt(
```
I prefer calculate to be more consistent with the rest of the repo.
And we can remove the thrust here as it's obvious from the object what the acceleration contribution is.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+        const Instant& anInstant,
+        const Shared<const Frame>& aFrameSPtr,
+        const TabulatedDynamics::InterpolationType& anInterpolationType = DEFAULT_INTERPOLATION_TYPE
+    );
+
+    Shared<TabulatedDynamics> toTabulatedDynamics(
+        const Shared<const Frame>& aFrameSPtr,
```
vishwa2710, line 115:
```suggestion
```
Is this required?
There's a Frame that is passed when creating the Maneuver object that is associated with the profile.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+    );
+
+    static Tabulated Load(const Maneuver::Type& aManeuverType, const File& aFile);
+
+   private:
+    const Type type_;
+    const Array<Instant> instants_;
+    const Array<Vector3d>& accelerationProfile_;
+    const Real massFlowRate_;
+    Shared<Interpolator> interpolatorSPtr_;
```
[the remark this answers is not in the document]
vishwa2710: - I think by having a constant mass flow rate we are unnecessarily limiting ourselves.
We can always have 2 constructors, that accept a Real, or a VectorXd, and then it's patched together under the hood.
- I don't see the value in having it's own interpolator...
- Agreed we should always store in an inertial frame like GCRF

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -145,6 +147,8 @@ class Propagator
     /// @endcode
     void clearDynamics();
 
+    void addManeuvers(const Array<Maneuver>& aManeuverArray);
```
vishwa2710, line 150: Should add a method to get/access maneuvers as well.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -105,6 +107,8 @@ class Segment
         /// @return Delta mass
         Mass computeDeltaMass() const;
 
+        Array<Maneuver> getManeuvers() const;
```
vishwa2710, line 110: I would call this `generateManeuvers` as we are generating them rather than it being a strict getter

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -120,6 +121,17 @@ void Propagator::setDynamics(const Array<Shared<Dynamics>>& aDynamicsArray)
...
 }
 
+void Propagator::addManeuvers(const Array<Maneuver>& aManeuverArray)
+{
+    for (const Maneuver& aManeuver : aManeuverArray)
+    {
+        const Shared<TabulatedDynamics> tabulatedDynamics =
+            aManeuver.toTabulatedDynamics(IntegrationFrameSPtr, aManeuver.getInterpolationType());
+
+        this->addDynamics(std::dynamic_pointer_cast<Dynamics>(tabulatedDynamics));
```
vishwa2710, line 131:
```suggestion
        this->addDynamics(tabulatedDynamics);
```
Don't think the dynamic pointer cast is necessary.

vishwa2710 (commented):
First pass.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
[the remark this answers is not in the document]
apaletta3: 1. Yeah that's true, i'll change it back to an Array of either Reals or a VectorXd depending on whether I think its worth it to have ISP as well as Mass being specified
2. The reason I wanted to have its own interpolator is that I don't feel like the interpolator is going to change much, so I don't want to have to specify it as an argument everytime I want a quantity from the maneuver interpolated. I have as a memvar for the same reason that TabulatedDynamics has it as a memvar. Do you think it would be better just to leave the "best" interpolator for maneuvers (that we find from testing) as a default argument for any function requiring interpolating the internal quantities of the object? 
@vishwa2710

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
apaletta3:
```suggestion
        Array<Maneuver> generateManeuvers() const;
```

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: True, I can make it a default, but I wanted to give the user the option to express the tabulated dynamics in another frame (for plotting etc)

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+    };
+
+    enum class InterpolationType
+    {
+
+        Linear,
+        BarycentricRational,
+        CubicSpline
+
+    };
```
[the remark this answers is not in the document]
apaletta3: Cool, I can do that

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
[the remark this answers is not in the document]
phc1990: Sorry just left a comment on why flowmassrate was a real
I see it's being agreed upon
so +1 on what @vishwa2710 has mentioned

Going back to the interpolator question or in general, they 'memvar' point that you bring up.

If we leave implementation aside, just from a "design" perspective, if a maneuver is a historical set of time-based numbers, it should be just that.

Any byproduct of that (for instance, a tabulated dynamics) could be created with many different "flavours" (the interpolator type being one of them for now). Having these "biproducts" as memvars "forces" them to be frozen with one "flavour" or another, whereas the "beauty" (imo) of the Maneuver object is that you store only what you need to recreate said bi-products. 

I see no issue in having default argument values for these "flavour options".

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+        const Array<Vector3d>& anAccelerationProfile,
+        const Shared<const Frame>& aFrameSPtr,
+        const Real& aMassFlowRate,
+        const InterpolationType& anInterpolationType = DEFAULT_INTERPOLATION_TYPE
+    );
+
+    bool isDefined();
+
+    Maneuver::Type getType();
+    Pair<Array<Instant>, Array<Vector3d>> getInstantsAndAccelerationProfile();
```
phc1990, line 99: Unless we provide the desired output frame as an argument, we should expose the frame somehow (e.g. `getFrame()`)

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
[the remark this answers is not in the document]
apaletta3: So I should make the `Interpolator` class in cpp have an `InterpolatorType` enum in its constructor, and then it selects the correct interpolator type under the hood based on that argument?

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -22,11 +22,12 @@ Tabulated::Tabulated(
...
-    const Shared<const Frame>& aFrameSPtr
+    const Shared<const Frame>& aFrameSPtr,
+    const InterpolationType& anInterpolationType
 )
     : Dynamics("Tabulated"),
-      contributionProfile_(aContributionProfile),
       instants_(anInstantArray),
+      contributionProfile_(aContributionProfile),
       writeCoordinatesSubsets_(aWriteCoordinatesSubsets),
       frameSPtr_(aFrameSPtr)
```
phc1990, line 32: We should check if the `aFrameSPtr` is inertial or pseudo-inertial

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -105,6 +107,8 @@ class Segment
         /// @return Delta mass
         Mass computeDeltaMass() const;
 
+        Array<Maneuver> generateManeuvers() const;
```
phc1990, line 110: Just asking here whether `generate` is the right prefix. Would the be generated on-demand or are they stored beforehand?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
phc1990: On this
Whereas I agree that this is to be done
Maybe for this PR is good enough to simply return an error if there are more than 1 items?

Just saying as that particular implementation, whilst "easy on paper" will increase the scope and size of this PR

phc1990 (commented):
Nice stuff!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
apaletta3: oh yeah I guess you're right, you only need to do that when you want to dereference the object afterwords right?

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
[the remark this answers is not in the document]
apaletta3: Okay agreed, and I guess `TabulatedDynamics` is slightly different cause from a software perspective, it behaves like any other dynamics, and "needs" to have its own interpolator in order to produce computedContributions at any requested instant. But since the Maneuver object is free on its own and doesn't have to conform to the function signature of the `computeContributions` function from a parent `Dynamics` class, it makes more sense to keep it lighter-weight without its own interpolator that can be specified as an argument. That is my understanding.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
apaletta3: Good point, will think about this

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: Good point

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: hmmm what difference would that make? I plan on supporting all frames (even non-international ones) for creating the object from acceleration profiles.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
apaletta3: They would be generated on demand (to not bloat the segmentSolution with an `Array<Maneuver>` memvar that it might not need for coast segments.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    Shared<TabulatedDynamics> toTabulatedDynamics(
+        const Shared<const Frame>& aFrameSPtr = AccelerationFrameSPtr,
+        const TabulatedDynamics::InterpolationType& anInterpolationType = DEFAULT_INTERPOLATION_TYPE
+    );
+
+    static Maneuver Load(const Maneuver::Type& aManeuverType, const File& aFile);
+
+   private:
+    const Type type_;
```
Derollez, line 136: Where do we use it?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
Derollez: I think Pau is asking: should we use generate, or rather extract or sth else

src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+Maneuver Maneuver::Load(const Maneuver::Type& aManeuverType, const File& aFile)
+{
+    // Assert on the type of file (OEM, OCM, etc)
+
+    // Sanitize the inputs and convert to GCRF frame for contribution profile
+
+    // TBI: Load the tabulated dynamics from the file
+}
```
Derollez, line 152: This is a good idea. We should probably do this as a separate MR later

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+#define DEFAULT_INTERPOLATION_TYPE Tabulated::InterpolationType::BarycentricRational
+
+/// @brief Stores information and metadata about a maneuver as performed by a spacecraft.
+class Maneuver
+{
+   public:
+    static const Shared<const Frame> AccelerationFrameSPtr;
+
+    enum class Type
```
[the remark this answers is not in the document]
apaletta3: Yeah, I'm removing this since honestly its not that useful.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
apaletta3: Yeah extract sounds good too, I'll change it to that

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
apaletta3: So actually I don't think we should do this because the logic to extract a maneuver from a `propagator` will be pretty complicated because the Maneuver object isn't stored as a memvar inside of `Propagator` (it should stay that way), so the logic would have to decide what `TabulatedDynamics` is "considered" a maneuver and what isn't (cause of course `TabulatedDynamics` don't just have to be maneuvers.

apaletta3:
Notes to reviewers @vishwa2710 @Derollez 
- Pls ignore all instances of the `InterpolationType` being defined as enums in the `Maneuver`, `TabulatedDynamics` or `TabulatedTrajectory` class, as this is temporary while I propagate [this PR](https://github.com/open-space-collective/open-space-toolkit-mathematics/pull/110) up from OSTk Math.
- Pls ignore commented out py unit tests in `test_maneuver.py`. They are commented out because they rely on my point above about `InterpolationType` as well as adding a `Scalar` binding in OSTk physics and propagating that up to astro

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
```diff
@@ -45,6 +48,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
                         contribution_profile (numpy.ndarray): A contribution profile.
                         coordinates_subsets (list[CoordinatesSubset]): An array of coordinates subsets related to the contribution profile.
                         frame (Frame): A frame.
+                        interpolation_type (Interpolator.Type, optional): The interpolation type. Defaults to default tabulated dynamics interpolation type.
```
vishwa2710, line 51:
```suggestion
                        interpolation_type (Interpolator.Type, optional): The interpolation type. Defaults to Barycentric Rational.
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
```diff
@@ -96,7 +100,19 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
...
+            )
+
+            .def(
+                "get_interpolation_types",
+                &Tabulated::getInterpolationTypes,
+                R"doc(
+                    Get the interpolation types for each row of the contribution profile.
+
+                    Returns:
+                        list[Interpolator.Type]: The interpolation types.
```
vishwa2710, line 114: It looks like we only set one, so they should all be same? Unless the constructor accepts a list of interpolator types?

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Maneuver.cpp
```diff
@@ -0,0 +1,268 @@
...
+        .def(
+            "get_end_instant",
+            &Maneuver::getEndInstant,
+            R"doc(
+                Get the end instant of the maneuver.
+
+                Returns:
+                    Instant: The end instant.
+            )doc"
+        )
```
vishwa2710, line 105: Maybe store these an interval instead of start and end instant?

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+using TabulatedDynamics = ostk::astro::dynamics::Tabulated;
+using ostk::astro::flight::system::PropulsionSystem;
+using ostk::astro::trajectory::state::CoordinatesSubset;
+using ostk::astro::trajectory::state::coordinatessubsets::CartesianVelocity;
+
+#define DEFAULT_MANEUVER_INTERPOLATION_TYPE Interpolator::Type::BarycentricRational
+
+/// @brief Stores information and metadata about a maneuver as performed by a spacecraft.
+class Maneuver
```
vishwa2710, line 68: IMO most of the `getters` here should be `accessors`, per our discussion on this thread: https://loftorbital.slack.com/archives/C03GCNLG8G5/p1704928993120699
With the python methods as `get_*`

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,251 @@
...
+    /// @return The mass flow rate profile
+    Array<Real> getMassFlowRateProfile() const;
+
+    /// @brief Get the start instant of the maneuver
+    /// @return The start instant
+    Instant getStartInstant() const;
+
+    /// @brief Get the end instant of the maneuver
+    /// @return The end instant
+    Instant getEndInstant() const;
```
vishwa2710, line 124: Let's return an interval instead of start and end to match patterns with some other objects.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,251 @@
...
+    ///
+    /// @param anInstant The instant at which to calculate the acceleration
+    /// @param aFrameSPtr The frame in which the acceleration is to be defined
+    /// @param anInterpolationType The interpolation type to use
+    ///
+    /// @return The acceleration
+    Vector3d calculateAccelerationAt(
+        const Instant& anInstant,
+        const Shared<const Frame>& aFrameSPtr,
+        const Interpolator::Type& anInterpolationType = DEFAULT_MANEUVER_INTERPOLATION_TYPE
```
vishwa2710, line 162: We shouldn't be live constructing the interpolator everytime this function is called. It's O(N) to construct a barycentric rational interpolator. And O(N) to call it.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,251 @@
...
+    ) const;
+
+    /// @brief Calculate the (interpolated) mass flow rate at a given instant during the maneuver
+    ///
+    /// @param anInstant The instant at which to calculate the mass flow rate
+    /// @param anInterpolationType The interpolation type to use
+    ///
+    /// @return The mass flow rate
+    Real calculateMassFlowRateAt(
+        const Instant& anInstant, const Interpolator::Type& anInterpolationType = DEFAULT_MANEUVER_INTERPOLATION_TYPE
```
vishwa2710, line 185: Same as above.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+    /// @brief Calculate the (interpolated) mass flow rates at given instants during the maneuver
+    ///
+    /// @param anInstantArray The instants at which to calculate the mass flow rates
+    /// @param anInterpolationType The interpolation type to use
+    ///
+    /// @return The mass flow rates
+    Array<Real> calculateMassFlowRatesAt(
+        const Array<Instant>& anInstantArray,
+        const Interpolator::Type& anInterpolationType = DEFAULT_MANEUVER_INTERPOLATION_TYPE
```
vishwa2710, line 196: Same as above.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Models/Tabulated.hpp
```diff
@@ -48,7 +48,7 @@ using ostk::physics::time::Scale;
 using ostk::astro::trajectory::Model;
 using ostk::astro::trajectory::State;
 
-#define DEFAULT_TABULATED_INTERPOLATION_TYPE Tabulated::InterpolationType::Linear
+#define DEFAULT_TABULATED_INTERPOLATION_TYPE Tabulated::InterpolationType::Linear  // TBM: once OSTk math MR is merged
```
vishwa2710, line 51: Looks like we can remove this now?
Can do it in a follow up MR.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -61,10 +62,11 @@ Tabulated::Tabulated(
     }
 
     interpolators_.reserve(aContributionProfile.cols());
-    for (i = 0; i < (Index)aContributionProfile.cols(); ++i)
+    for (i = 0; i < (Index)aContributionProfile.cols(); i++)
```
vishwa2710, line 65: Why change this?
https://stackoverflow.com/questions/24853/what-is-the-difference-between-i-and-i

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -105,6 +107,17 @@ Shared<const Frame> Tabulated::getFrame() const
     return accessFrame();
 }
 
+Array<Interpolator::Type> Tabulated::getInterpolationTypes() const
```
vishwa2710, line 110: I think this function is obsolete because we provide only a single interpolation type so there's no point returning an array.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -105,6 +107,17 @@ Shared<const Frame> Tabulated::getFrame() const
...
+{
+    Array<Interpolator::Type> interpolationTypes = Array<Interpolator::Type>::Empty();
+
+    for (auto& interpolationSPtr : interpolators_)
```
vishwa2710, line 114:
```suggestion
    for (const auto& interpolationSPtr : interpolators_)
```
Always const unless we are modifying it.

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -251,10 +276,7 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, ComputeContribution)
             loadData("/app/test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated_Earth_Gravity.csv");
 
         const Tabulated tabulated = {
-            instants,
-            contributionProfile,
-            defaultWriteCoordinatesSubsets_,
-            defaultFrameSPtr_,
+            instants, contributionProfile, defaultWriteCoordinatesSubsets_, defaultFrameSPtr_, defaultInterpolationType_
```
vishwa2710, line 279:
```suggestion
            instants, contributionProfile, defaultWriteCoordinatesSubsets_, defaultFrameSPtr_, defaultInterpolationType_,
```
And fix the formatting.

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -211,14 +226,22 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Getters)
...
 {
     const VectorXd x;  // Not used
     {
         Tabulated tabulated = {
-            defaultInstants_, contributionProfile_, defaultWriteCoordinatesSubsets_, defaultFrameSPtr_
+            defaultInstants_,
+            contributionProfile_,
+            defaultWriteCoordinatesSubsets_,
+            defaultFrameSPtr_,
+            defaultInterpolationType_
```
vishwa2710, line 244:
```suggestion
            defaultInterpolationType_,
```

bindings/python/test/flight/test_maneuver.py
```diff
@@ -0,0 +1,237 @@
...
+        assert maneuver.get_end_instant() == instants[-1]
+
+        assert maneuver.get_duration() == instants[-1] - instants[0]
+
+    def test_calculators(
+        self,
+        maneuver: Maneuver,
+    ):
+        pass
+        # assert maneuver.calculate_delta_v().get_value() == pytest.approx(
```
vishwa2710, line 117: TBD?

vishwa2710 (commented):
Did a first pass.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
apaletta3: Good call

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Maneuver.cpp
apaletta3: They're not stored as interval, they're stored as an array of instants, but yes I can return an interval

bindings/python/test/flight/test_maneuver.py
apaletta3: Mentioned this in my comment to reviewers, waiting on releasing `Scalar` in python in OSTk phys

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: yeah, lets resolve this once and for all because I believe we should use getters instead of accessors

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Models/Tabulated.hpp
apaletta3: Yup forgot this one
```suggestion

```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: They're effectively identical when its used like that in a loop, so its just a style choice here. But you're right, no need to change it.
```suggestion
    for (i = 0; i < (Index)aContributionProfile.cols(); ++i)
```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: Ah yes good point, even more important when dealing with references

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: @vishwa2710 How else would you recommend it be done? 

The only way I could imagine how to avoid this is to have an interpolator type supplied as an argument as construction time of the `Maneuver` class and then stored as a member variable. Then you could leverage that interpolator (with its internally stored x's and y's) for each interpolation call. However you recommended against doing this in a previous comment.

I agree, if a user calls the `calculateAccelerationAt` function in a loop it will be slower, but that's what the `calculateAccelerationsAt` function is for (similar to this pattern in the `Propagator.calculateState(s)At`). The user is unlikely to call the `calculateAccelerationsAt` function multiple times in a loop, and the maneuver object itself will not be used in propagation (TabulatedDynamics will), so I don't think we should prematurely optimize for performance here.

Ditto for your comments on `calculateMassFlowRate(s)At` below (resolved those threads so we can keep the discussion on this one). Let me know what you think

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: Sure, may be worth aligning `Segment` and `Sequence` with this paradigm as well then

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
vishwa2710: In general I would agree that we should not prematurely optimize. But from a user perspective there is no knowledge that one is more performant than the other or any such contract.
If we provide a way to calculate accelerations, users can use it do what they see fit.

1 way of dealing with this is to indeed provide an interpolation type that goes with the profile to specify how the profile should be interpolated, I'm not a huge fan of this.

Another way of handling this is to remove the `calculateAcceleration*` methods and just enforce users to create a tabulated dynamics and compute their accelerations from that.
In this way users are forced to reason about the interpolation type at the correct interface, and are guaranteed to get performance as the interpolator is created once and the right time.

We can also forgo the performance issues and leave it as is. IMO Maneuvers being mostly data objects, we are overloading the interpolation and "calculation" concepts whereas we should be separating our concerns. If you want to interpolate, create a tabulated dynamics. But feel free to seek further opinions from the team, or resolve it if you feel strongly.

apaletta3: @vishwa2710 I see what you mean, that makes sense. I don't like the idea of  removing the calculateAcceleration* methods because they add a good amount of convenience for the user for plotting and stuff.

I think this "warning about not optimized performance" point is not that critical, because we already have something like this out [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp#L139) for propagator.calculateStateAt where a user can call it over and over in a loop and it will reduce and expand the state between each call which not great efficiency wise.

I think the potential for misuse of performance decrease in this case (as in the propagator case) is outweighed by the convenience of having both the calculateXAt and calculateXsAt methods that we can leverage in the services, etc.

However, happy to have a team consensus of this sort of debate moving forward. @Derollez @phc1990 @kyle-cochran any opinions on this?

Also I'm curious why you're not a fan of the maneuver class having its own interpolator (from a software correctness interface perspective)

vishwa2710: Maneuver to me is a data object. In general, we have the ability move "up" all methods from child objects to the parent. However that is not great practice as if and when the interfaces change you are locked into multiple levels of breaking interfaces.
Ideally you just return a child object and allow the user to manipulate the information that they want.

In this case, I already feel like we are tacking on too much to the Maneuver object, if a user wants to interpolate in a specific way it's quite simple to:
`maneuver.to_tabulated_dynamics().compute_contribution(instant, Frame.GCRF())`

The added "simplicity" of a `calculateThrustAccelerationAt` IMO is outweighed by the mixing of a data model with a computational model, additional unit tests, boiler plate code, code maintenance and updates etc.

For example, if a class has an interval as a member variable, we don't additionally save the duration, we just return the interval, and then allow the user the get the duration from the interval as that interface is locked in and not the responsibility of the original class.

apaletta3: Ok yes that makes a lot of sense, thanks for the clear explanation. I'm actually in favor of removing the `calculateAcceleration(s)At` and `calculateMassFlowRate(s)At` methods to force users to use `TabulatedDynamics` to interpolate.

But now, just to play devil's advocate here 🗡️ , there's no plural version  of `computeContribution`, so the user would have to loop over it to interpolate at multiple points which would be slower if the non-default (non-eci) frame is requested because a new frame would be created on every call. What do you think about this? I don't think its a big deal, but just something I noticed

Derollez: > For example, if a class has an interval as a member variable, we don't additionally save the duration, we just return the interval, and then allow the user the get the duration from the interval as that interface is locked in and not the responsibility of the original class.

This is a good point and great example to illustrate it.

> there's no plural version of computeContribution, so the user would have to loop over it to interpolate at multiple points which would be slower

That's also a good point.

I think you guys made some good points. I understand where Vishwa is coming from to keep objects and methods as elementary and modular as possible and give the responsibility to the user to introduce some performance hits. What about this to move forward on the short term -> We keep it simple and modular, remove those methods for now, deploy and start using in the service and in notebooks. We assess the potential pain or disconnect at user level and if it's worth it, we can always come back here and add the code (which already exists ^^).

@phc1990 Let us know if you have an opinion on this: You might have a different angle from the Guidance service development :)

Derollez: Actually Pau's comment below on different lines:
> If we leave implementation aside, just from a "design" perspective, if a maneuver is a historical set of time-based numbers, it should be just that.

Any byproduct of that (for instance, a tabulated dynamics) could be created with many different "flavours" (the interpolator type being one of them for now). Having these "biproducts" as memvars "forces" them to be frozen with one "flavour" or another, whereas the "beauty" (imo) of the Maneuver object is that you store only what you need to recreate said bi-products.


Seems to indicate that he is leaning towards the same conclusion but let's confirm and move 🙏 :)

phc1990: Just posting a new comment, since it's a slightly different but related topic. 

I think there is yet a 3rd group of functions that were considering:

- "Non-deterministic" getters/computer: this involves the computation of what might look like unambiguous properties, but, in reality, they are not. Examples of these are:
  - calculateDeltaV
  - calculateDeltaMass
  - calculateAverageThrust
  - calculateAverageSpecificImpulse
  - etc.

Whilst form the outside, one could do `maneuver.calculateDeltaV()` and simply get a number, I think this might be confusing:

On the one hand, we are providing a "raw but nice" way to the user to create a `TabulatedDynamics` where they are in full control with arguments such as the interpolator type...

...and on the other hand, we are "hiding" that we are using that behind each `calculateX` function an Euler type numerical integration is taking place.

What I would suggest:

1. We keep this "compute" functions to the bare minimum (for now) (e.g. are averageThrust and averageSpecificImpulse needed atm?, if so great 👍 )
2. For the functions we do know we actually need (maybe calculateDeltaV, calculateDeltaMass?), we already make the API explicit: we go from `calculateX()` to `calculateY(anIntegrationType)`. It does not matter if we only offer an EULER_TYPE integration at the moment, but this way we are consistent.
3. In the future (definitively not now), we should explore the option of moving the numerical integration capabilities to the `TabulatedDynamics` class. E.g:
  `tabulatedDynamcis.integrateContribution(aSubset , anIntegrationType)`

phc1990: (Apologies if these comments come in late, I lost track of this PR. All in all, the simpler we keep the `ManeuverObject` the better imo)

phc1990 (commented):
Left some comments on ManeuverObject API

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
phc1990: Maybe local orbital frame does not extend from `Frame` (can't remember right now), if that's the case we are good : )

apaletta3: @phc1990 I think we can accept them as long as they are converted into the inertial frame at construction time (and thus not stored in a local frame). That is essentially what I have done in Maneuver [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/feat/add-maneuver-class/src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp#L56).

But good point, I can add this to TabulatedDynamics so that the interpolation and storing of the contribution profile under the hood is always in the default (GCRF) frame, good idea.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
apaletta3: Great points guys, thanks for the feedback. 
- Agreed, I will removed the `calculateX(s)At(instant(s))` methods
- I will change the getters to be accessors following our slack thread (unrelated to the conversation above and doesn't change the behavior at all)
- I see your point @phc1990 about being more explicit regarding the fact that we're actually doing euler integration inside the calculateX methods, but I feel like specifying that is overkill at this point. These methods will be used in the guidance service to compute metrics and analyze the maneuvers that are planned or have been performed, and I'm not sure we will need to make them more accurate by upping the order of the integration happening under the hood. That being said, I like your idea of adding a  `tabulatedDynamcis.integrateContribution(aSubset , anIntegrationType)` in a future refactor when we have more time, I think that's very clean. Also, these methods can be used by the mission analysis team for convenience as well. Therefore, I plan on keeping these four `calculate` methods for now cause I think they will be useful in the guidance service, but I'm happy to remove them in the future if they become confusing. Let me know if you agree with this.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
phc1990: okok I think we might be talking about different things

What I meant is that the referenceFrame given to the Maneuver Object (at construction time) or the referenceFrame given to the function toTabulatedDynamcis(...) should not be a local orbital frame.

Why?

Because the scope of the Maneuver Object is to be a historical and **reproducible** item. Having a local orbital frame will break this, as the Maneuver Objet would no longer be self contained.

So here:

```
    Maneuver(
        const Array<Instant>& anInstantArray,
        const Array<Vector3d>& anAccelerationProfile,
        const Shared<const Frame>& aFrameSPtr,
        const Array<Real>& aMassFlowRateProfile
    );
```

The `aFrameSPtr` should not be a local orbital frame.

And I don't think we should allow it at any other function either (e.g. `toTabultedDynamics(frame)`).

Just to be clear: I was not suggesting we should have `TabulatedDynamics` use GCRF by default.

Does this make sense ?

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
phc1990: All good on my end @apaletta3 👍

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: Ah yes ok I agree. I need to double check how to prevent this cause if I recall correctly local orbital frames aren't always represented as `Shared<Frame>`.

But now that I think about it, I also think TabulatedDynamics needs to use GCRF by default, because if it uses a non-inertial frame or even a local frame we run into the same problem where it is no longer self contained, right? And since its used in the propagator, we shouldn't allow the possibility of numerically integrating in a non-inertial frame, which I think would be currently allowed.

apaletta3: @phc1990 I took a brief look into this, and with the code from `LocalOrbitalFrameFactory` as it currently stands it looks like we are protected from this happening, but I took an action in JIRA to investigate how we can prevent the wrong frames from being used to specify accelerations for maneuver in my next steps

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -117,14 +125,14 @@ Array<Shared<const CoordinateSubset>> Tabulated::getReadCoordinateSubsets() cons
...
-    return writeCoordinateSubsets_;
+    return writeCoordinateSubset_;
 }
 
 VectorXd Tabulated::computeContribution(
     const Instant& anInstant, [[maybe_unused]] const VectorXd& x, const Shared<const Frame>& aFrameSPtr
 ) const
 {
-    // TBI: Eventually we can check if the values can be converted using the subset inFrame methods.
+    // TBM: Convert this using the Maneuver class' conversion mehtod.
```
apaletta3, line 135:
```suggestion
    // TBM: Convert this using the Maneuver class' conversion method. Also, find a way to check and vet whether or not the frame is local, or quasi-inertial
```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -117,14 +125,14 @@ Array<Shared<const CoordinateSubset>> Tabulated::getReadCoordinateSubsets() cons
 
 Array<Shared<const CoordinateSubset>> Tabulated::getWriteCoordinateSubsets() const
 {
-    return writeCoordinateSubsets_;
+    return writeCoordinateSubset_;
```
vishwa2710, line 128:
```suggestion
    return writeCoordinateSubsets_;
```
This is plural as we are returning subsets, like the function name says.
I apologize if the refactor was unclear but here are semantics:
CoordinateSubset -> A single Coordinate Subset
CoordinateSubsets -> an Array with each element of type Coordinate Subset

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -21,13 +21,14 @@ using ostk::mathematics::object::VectorXd;
...
+    const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubset,
+    const Shared<const Frame>& aFrameSPtr,
+    const Interpolator::Type& anInterpolationType
 )
     : Dynamics("Tabulated"),
-      contributionProfile_(aContributionProfile),
       instants_(anInstantArray),
-      writeCoordinateSubsets_(aWriteCoordinateSubsets),
+      contributionProfile_(aContributionProfile),
+      writeCoordinateSubset_(aWriteCoordinateSubset),
```
vishwa2710, line 31:
```suggestion
      writeCoordinateSubsets_(aWriteCoordinateSubsets),
```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -21,13 +21,14 @@ using ostk::mathematics::object::VectorXd;
 Tabulated::Tabulated(
     const Array<Instant>& anInstantArray,
     const MatrixXd& aContributionProfile,
-    const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubsets,
-    const Shared<const Frame>& aFrameSPtr
+    const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubset,
```
vishwa2710, line 24:
```suggestion
    const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubsets,
```

include/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.hpp
```diff
@@ -45,13 +48,15 @@ class Tabulated : public Dynamics
     ///
     /// @param anInstantArray An array of instants, must be sorted
     /// @param aContributionProfile A contribution profile, one row for each instant
-    /// @param aWriteCoordinateSubsets An array of coordinate subsets to write to
+    /// @param aWriteCoordinateSubset An array of coordinate subsets to write to
```
vishwa2710, line 51:
```suggestion
    /// @param aWriteCoordinateSubsets An array of coordinate subsets to write to
```

include/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.hpp
```diff
@@ -45,13 +48,15 @@ class Tabulated : public Dynamics
...
     /// @param aFrameSPtr A frame
+    /// @param anInterpolationType An interpolation type
     Tabulated(
         const Array<Instant>& anInstantArray,
         const MatrixXd& aContributionProfile,
-        const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubsets,
-        const Shared<const Frame>& aFrameSPtr
+        const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubset,
```
vishwa2710, line 57:
```suggestion
        const Array<Shared<const CoordinateSubset>>& aWriteCoordinateSubsets,
```

include/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.hpp
```diff
@@ -130,11 +140,11 @@ class Tabulated : public Dynamics
     virtual void print(std::ostream& anOutputStream, bool displayDecorator = true) const override;
 
    private:
-    const MatrixXd contributionProfile_;
     const Array<Instant> instants_;
-    const Array<Shared<const CoordinateSubset>> writeCoordinateSubsets_;
+    const MatrixXd contributionProfile_;
+    const Array<Shared<const CoordinateSubset>> writeCoordinateSubset_;
```
vishwa2710, line 145:
```suggestion
    const Array<Shared<const CoordinateSubset>> writeCoordinateSubsets_;
```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -37,7 +38,7 @@ Tabulated::Tabulated(
         );
     }
 
-    if (aWriteCoordinateSubsets
+    if (aWriteCoordinateSubset
```
vishwa2710, line 41:
```suggestion
    if (aWriteCoordinateSubsets
```

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -152,7 +160,7 @@ void Tabulated::print(std::ostream& anOutputStream, bool displayDecorator) const
 
     Dynamics::print(anOutputStream, false);
 
-    for (const auto& subset : writeCoordinateSubsets_)
+    for (const auto& subset : writeCoordinateSubset_)
```
vishwa2710, line 163:
```suggestion
    for (const auto& subset : writeCoordinateSubsets_)
```

src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp
```diff
@@ -0,0 +1,224 @@
...
+{
+    aManeuver.print(anOutputStream);
+
+    return anOutputStream;
+}
+
+bool Maneuver::isDefined() const
+{
+    return true;
+}
```
vishwa2710, line 81: Maybe we should just delete this if it's not being used at all and it's always returning true?

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -51,8 +52,9 @@ class OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated : public ::testing::Test
         Instant::DateTime(DateTime(2018, 01, 01, 0, 0, 10, 000), Scale::UTC),
         Instant::DateTime(DateTime(2018, 01, 01, 0, 0, 15, 000), Scale::UTC),
     };
-    const Array<Shared<const CoordinateSubset>> defaultWriteCoordinateSubsets_ = {CartesianVelocity::Default()};
+    const Array<Shared<const CoordinateSubset>> defaultWriteCoordinateSubset_ = {CartesianVelocity::Default()};
```
vishwa2710, line 55:
```suggestion
    const Array<Shared<const CoordinateSubset>> defaultWriteCoordinateSubsets_ = {CartesianVelocity::Default()};
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -90,16 +92,26 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Constructor)
 {
     {
         EXPECT_NO_THROW(
-            Tabulated(defaultInstants_, contributionProfile_, defaultWriteCoordinateSubsets_, defaultFrameSPtr_)
+            Tabulated(defaultInstants_, contributionProfile_, defaultWriteCoordinateSubset_, defaultFrameSPtr_)
```
vishwa2710, line 95:
```suggestion
            Tabulated(defaultInstants_, contributionProfile_, defaultWriteCoordinateSubsets_, defaultFrameSPtr_)
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -90,16 +92,26 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Constructor)
...
         );
     }
 
+    {
+        EXPECT_NO_THROW(Tabulated(
+            defaultInstants_,
+            contributionProfile_,
+            defaultWriteCoordinateSubset_,
```
vishwa2710, line 103:
```suggestion
            defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -90,16 +92,26 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Constructor)
...
+        ));
+    }
+
     {
         EXPECT_THROW(
             {
                 try
                 {
-                    Tabulated({}, contributionProfile_, defaultWriteCoordinateSubsets_, defaultFrameSPtr_);
+                    Tabulated({}, contributionProfile_, defaultWriteCoordinateSubset_, defaultFrameSPtr_);
```
vishwa2710, line 114:
```suggestion
                    Tabulated({}, contributionProfile_, defaultWriteCoordinateSubsets_, defaultFrameSPtr_);
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -143,8 +155,9 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, StreamOperator)
     Tabulated tabulated = {
         defaultInstants_,
         contributionProfile_,
-        defaultWriteCoordinateSubsets_,
+        defaultWriteCoordinateSubset_,
```
vishwa2710, line 158:
```suggestion
        defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -159,8 +172,9 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Print)
...
```
vishwa2710, line 175:
```suggestion
        defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -174,7 +188,7 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Accessors)
...
```
vishwa2710, line 191:
```suggestion
        defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -196,8 +210,9 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Getters)
...
```
vishwa2710, line 213:
```suggestion
        defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -211,14 +226,22 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Getters)
...
 
 TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, ComputeContribution)
 {
     const VectorXd x;  // Not used
     {
         Tabulated tabulated = {
-            defaultInstants_, contributionProfile_, defaultWriteCoordinateSubsets_, defaultFrameSPtr_
+            defaultInstants_,
+            contributionProfile_,
+            defaultWriteCoordinateSubset_,
```
vishwa2710, line 242:
```suggestion
            defaultWriteCoordinateSubsets_,
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -253,8 +278,9 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, ComputeContribution)
         const Tabulated tabulated = {
             instants,
             contributionProfile,
-            defaultWriteCoordinateSubsets_,
+            defaultWriteCoordinateSubset_,
```
vishwa2710, line 281:
```suggestion
            defaultWriteCoordinateSubsets_,
```

vishwa2710 (approved):
Just some nits, feel free to apply and merge. LGTM! Great stuff!

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
phc1990: all good!

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Maneuver.cpp
```diff
@@ -0,0 +1,183 @@
...
+        .def("__repr__", &(shiftToString<Maneuver>))
+
+        .def(
+            "is_defined",
+            &Maneuver::isDefined,
+            R"doc(
+                Check if the maneuver is defined.
+
+                Returns:
+                    bool: True if the maneuver is defined, False otherwise. (Always returns true).
```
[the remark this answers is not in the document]
apaletta3: @vishwa2710 I wasn't actually planning on changing that in the future. My idea with the maneuver object, is that since its just a holder of maneuver information, it can't be constructed with partial information, therefore once it is constructed it will always be defined. What do you think about this?

src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp
apaletta3: I've seen other objects do this as well, but yea that might be a good idea. I'll leave it here for consistency for now thought.

Issue #334, opened 2024-02-06
[fix]  Build Failure in FiniteDifferenceSolver Due to Eigen Matrix Explicit Constructor Initialization


DhruvJ22:
**Describe the bug**

The bug occurs in the FiniteDifferenceSolver.cpp file, specifically when attempting to initialize an Eigen::MatrixXd object with a brace-enclosed initializer list. This operation fails due to the explicit constructor of Eigen::Matrix, which does not allow implicit conversions that would be necessary for this kind of initialization. The error is triggered at the following line:

 `const MatrixXd coordinatesMatrix = {coordinates};`

This line attempts to copy-initialize an Eigen::MatrixXd object with a VectorXd (or similar), which is not directly supported due to the explicit nature of the Eigen::Matrix constructor. The error message provided by the compiler points out that the chosen constructor is explicit in copy-initialization, referencing the explicit constructor declaration in the Eigen library's Matrix.h file.

The issue stems from a misunderstanding or oversight regarding Eigen's initialization requirements, leading to a build failure when attempting to compile the FiniteDifferenceSolver.cpp source file.

**Steps to reproduce**
1. Fork and clone the repository. 
2. Start the dev environment in container: `make start-development`
```
cmake ..
make
```
Screenshot of the error message:
![Screenshot from 2024-02-06 00-38-08](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/assets/33181026/0d86f492-4553-411f-8220-75fce293d542)

**Potential suggestion**

To resolve this issue, directly construct the MatrixXd object using a method that Eigen supports for converting or initializing from a VectorXd. If the intent is to create a matrix where coordinates are column elements then update to:
`const MatrixXd coordinatesMatrix = coordinates;`
If supposed to be a row in the matrix then:
`const MatrixXd coordinatesMatrix = coordinates.transpose();`

vishwa2710:
Hi @DhruvJ22 . Thanks for making this ticket and the thorough write up. I'm surprised this is happening as the CI tests are passing and I cannot re-produce this locally. I even made a new fork and tried to build from scratch.
Quite confused how this is happening.
Can you share some details on your environment?
- What OS are you running this on?
- What version of docker are you using?
- Can you share the CMake settings that you used?

From within the docker shell can you try running:
- ostk-clean
- ostk-build

And see what you get?

DhruvJ22:
Hi @vishwa2710, I hope the following details help: 

[...]

Issue #249, opened 2023-10-16
[fix] dead links in docuementation


kyle-cochran:
**Describe the bug**

From a list compiled by Dhruv Jain

	1.	OTSK Astrodynamics:
	1.	https://open-space-collective.github.io/open-space-toolkit-astrodynamics/html/index.html 
	1.	Documentation >>Python >> 404
	2.	Tutorials >> C++ >> 404
	3.	Tutorials >> C++ >> 404
	4.	Contribution >> Contributing guide >> blank
	5.	Dependencies >> Eigen >> http://eigen.tuxfamily.org/index.php “website seems to be down”

	1.	README 
	1.	Documentation >> Python >> “I do not believe the link is landing on the correct page”
	2.	Tutorials >> C++ >> “sensor modelling example lacks description and comments to be easy to understand”
	3.	Tutorials >> Python >> “no tutorial” -> route it to notebooks folder

PR #335, opened 2024-02-06
fix: contributing guide link in docs and Eigen Matrix constructor

DhruvJ22:
- Update the contributing guide link to be absolute instead of relative. Markdown and Doxygen handle relative links differently, hence, the Doxygen-generated doc didn't correctly link the *contributing guide* in README to the desired page. One of the deadlinks pointed out in  #249.

- Fix for issue in #334


DhruvJ22:
@kyle-cochran and @vishwa2710, proposed two fixes. I am unable to add you two as the reviewers. How do I go about it when I create the PR?

PR #338, opened 2024-02-09
refactor!: remove profile state

vishwa2710:


apaletta3 (approved):
Looks great!

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Models/Tabulated.cpp
apaletta3, line 1: Would you mind refactoring the logic from L89-L126 to use references for the stateRange instead of raw pointers?

vishwa2710: Can do that in a separate MR.

PR #344, opened 2024-02-13
build: add ssh-client to docker image

vishwa2710:
This allows us to git push etc. from within the docker container


vishwa2710:
Going to add it to the base image instead.

PR #353, opened 2024-02-15
refactor: propulsion system to use Reals instead of Scalar

apaletta3:
After going down a rabbit hole of trying to add python bindings for the `physics::data::Scalar` class (which currently lacks them) in OSTk physics, I have deemed that much harder than it seems on the surface and come up with an alternative medium term solution.

Context on the cpp side for OSTk physics:
- There is lots of convoluted code deep down inside of physics::Unit, which would take a non-trivial amount of time to modify if needed
- the tests for physics::Unit are weak, and the tests for physics::unit::Derived are non-existant (which is scary because we use physics::unit::Derived in a lot of places in astro
- the tests for data::Scalar and data::Vector and data::Direction are non-existant
- all derived units should be defined inside of physics::Unit, not as the Thrust and SpecificImpulse and MassFlowRate ones are currently defined inside of `PropulsionSystem`

Context on the python side for OSTk physics
-  physics::Unit,  physics::data::Scalar, physics::data::Vector, and physics::data::Direction are not bound to python at all as classes, and in order to add bindings to physics::data::Scalar, you need to add bindings to physics::Unit, which is not trivial at all due to the aforementioned convolutedness of physics::Unit
- if data::Scalar and data::Vector were to be bound to python, cpp tests would need to be added for them as well, which would again take time

Context on how the above relates to OSTk astro:
- ostk::physics::Unit is only used in two places in OSTk astro (AtmosphericDragDynamics (not at the API level) and PropulsionSystem (indeed at the API level)
- data::Scalar is only used in PropulsionSystem and data::Vector is used nowhere in OSTk astro
- SatelliteSystem, dynamics, and propagation all do their computations and store their variables in unitless Reals or VectorXds
- after the bonanza of ostk releases over the past couple days, no more breaking changes 🙏 

Path forward:
- One day I think we will need a complete overhaul of the unit system in OSTk, by either a) completely removing it, b) replacing it with templated compile-time unit checking or boost::units 
- **However, in the meantime, and based on all the context above I think the best solution is to remove the use of data::Scalar from PropulsionSystem and store its quantities in SI for 1) consistency with everything else in astro and 2) the fact that you can't use the getters in Python for `PropulsionSystem` without adding Scalar bindings, and we are going to be using the `PropulsionSystem` object a lot in the near future in the guidance service and everything related to it**

There is a corresponding [MR](https://github.com/open-space-collective/open-space-toolkit-physics/pull/205) in physics that comments out pybound methods that won't work because they return data::Scalar and data::Vector.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Flight/System/PropulsionSystem.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Flight/System/PropulsionSystem.hpp
@@ -5,11 +5,7 @@
 
 #include <OpenSpaceToolkit/Core/Type/Real.hpp>
 
-#include <OpenSpaceToolkit/Physics/Data/Direction.hpp>
-#include <OpenSpaceToolkit/Physics/Data/Scalar.hpp>
-#include <OpenSpaceToolkit/Physics/Unit.hpp>
-#include <OpenSpaceToolkit/Physics/Unit/Derived.hpp>
-#include <OpenSpaceToolkit/Physics/Unit/Time.hpp>
+#include <OpenSpaceToolkit/Physics/Unit/Mass.hpp>
 
 namespace ostk
 {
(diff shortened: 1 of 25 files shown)


include/OpenSpaceToolkit/Astrodynamics/Flight/System/PropulsionSystem.hpp
```diff
@@ -22,34 +18,12 @@ namespace system
...
-    /// @brief Constructor
-    ///
-    /// @code{.cpp}
-    ///              PropulsionSystem propulsion = { ... };
-    /// @endcode
-    ///
-    /// @param aThrust Thrust (scalar)
-    /// @param aSpecificImpulse Specific impulse (scalar)
-    PropulsionSystem(const Scalar& aThrust, const Scalar& aSpecificImpulse);
-
```
apaletta3, line 52: Even though I am removing this constructor, its not really a breaking change because we don't use it anywhere downstream in cpp (except for the `PropulsionSystem.test.cpp` that I also refactored in this MR, and it doesn't exist in Python.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/AtmosphericDrag.cpp
```diff
@@ -29,9 +29,6 @@ using ostk::astrodynamics::trajectory::state::CoordinateSubset;
 using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianPosition;
 using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianVelocity;
 
-static const Derived::Unit GravitationalParameterSIUnit =
```
apaletta3, line 32: This doesn't change anything, some of these were just defined in files they're not actually used in, so I went ahead and removed other examples like this.

Derollez (approved):
Thanks for the investigation details and the cleaning 🙏

PR #362, opened 2024-02-21
fix: fix binding docstring default variable logging

vishwa2710:


apaletta3 (commented):
@vishwa2710 So this is supposed to swap every instance of `arg("XX") = YY` with the `arg_v` notation in pybind, because only that works with the sphinx docs?

Cause if that's the case I think you missed 50 or so instances across the repo, I checked with this regex below:
`arg\("(\w+)"\) =`

vishwa2710:
@apaletta3 sorry I should have put a better description. We only need to update with `arg_v` for things that have the OSTk pretty print. This is because the pretty print contains `======` headers and footers `------` or something equivalent which breaks the .rst format as those are key characters in rst.

apaletta3 (approved):

PR #364, opened 2024-02-22
refactor: add dynamics validation to propagator

apaletta3:
This MR refactors the propagator slightly by adding some dynamics validation logic so that propagation can't be done with a set of dynamics that doesn't make sense. 
- It defines a maximum and minimum number of each type of dynamics that can be used for propagation in a map, and perform a "propagation-time" check that occurs as the first step of the `calculateXAt()` methods being called, so that the user can add whatever dynamics they want one by one and/or remove dynamics using `.addDynamics` and `.setDynamics` and `.clearDynamics` until they have a good dynamics set, which is only validated before it will actually be used to calculate a state.

This MR refactors the `Propagator` unit tests in some places to be more concise, and also adds a deep copy constructor and deep copy assignement operator to the `Propagator` so that if a `Propagator` needs to be duplicated, the dynamics inside of it are also duplicated because if they're not both propagators would still be sharing the same dynamics which could be an insidious issue.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
@@ -72,13 +72,6 @@ class Propagator
    public:
     /// @brief Default integrator frame
     static const Shared<const Frame> IntegrationFrameSPtr;
-    /// @brief Map of valid dynamics set for propagation
-    ///
-    /// @code{.cpp}
-    ///              Map<String, Pair<Size, Size>> validDynamicsSet ={"TypeOfDynamics", {Min number allowed, Max number
-    ///              allowed}};
-    /// @endcode
-    static const Map<String, Pair<Size, Size>> ValidDynamicsSet;  // Map<String, Pair<Size, Size>>
 
     /// @brief Constructor
     ///
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
@@ -4,6 +4,9 @@
 #define __OpenSpaceToolkit_Astrodynamics_Trajectory_Propagator__
 
 #include <OpenSpaceToolkit/Core/Container/Array.hpp>
+#include <OpenSpaceToolkit/Core/Container/Map.hpp>
+#include <OpenSpaceToolkit/Core/Container/Pair.hpp>
+#include <OpenSpaceToolkit/Core/Type/Index.hpp>
 #include <OpenSpaceToolkit/Core/Type/Integer.hpp>
 #include <OpenSpaceToolkit/Core/Type/Real.hpp>
 #include <OpenSpaceToolkit/Core/Type/Shared.hpp>
@@ -42,6 +45,9 @@ namespace trajectory
 {
 
 using ostk::core::container::Array;
+using ostk::core::container::Map;
+using ostk::core::container::Pair;
+using ostk::core::type::Index;
 using ostk::core::type::Integer;
 using ostk::core::type::Real;
 using ostk::core::type::Shared;
@@ -53,18 +59,26 @@ using ostk::physics::coordinate::Velocity;
 using ostk::physics::time::Duration;
 using ostk::physics::time::Instant;
 
-using ostk::astrodynamics::trajectory::state::NumericalSolver;
+using ostk::astrodynamics::Dynamics;
 using ostk::astrodynamics::EventCondition;
+using ostk::astrodynamics::flight::system::SatelliteSystem;
 using ostk::astrodynamics::trajectory::State;
 using ostk::astrodynamics::trajectory::StateBuilder;
-using ostk::astrodynamics::Dynamics;
-using ostk::astrodynamics::flight::system::SatelliteSystem;
+using ostk::astrodynamics::trajectory::state::NumericalSolver;
 
 /// @brief Define a propagator to be used for numerical propagation
 class Propagator
 {
    public:
+    /// @brief Default integrator frame
     static const Shared<const Frame> IntegrationFrameSPtr;
+    /// @brief Map of valid dynamics set for propagation
+    ///
+    /// @code{.cpp}
+    ///              Map<String, Pair<Size, Size>> validDynamicsSet ={"TypeOfDynamics", {Min number allowed, Max number
+    ///              allowed}};
+    /// @endcode
+    static const Map<String, Pair<Size, Size>> ValidDynamicsSet;  // Map<String, Pair<Size, Size>>
 
     /// @brief Constructor
     ///
@@ -78,6 +92,17 @@ class Propagator
         const NumericalSolver& aNumericalSolver,
         const Array<Shared<Dynamics>>& aDynamicsArray = Array<Shared<Dynamics>>::Empty()
     );
+
+    /// @brief Copy constructor
+    ///
+    /// @param aPropagator A propagator (deep copy)
+    Propagator(const Propagator& aPropagator);
+
+    /// @brief Copy assignment operator
+    ///
+    /// @param aPropagator A propagator (deep copy)
+    /// @return A reference to this propagator
+    Propagator& operator=(const Propagator& aPropagator);
 
     /// @brief Equal to operator
     ///
@@ -213,7 +238,7 @@ class Propagator
     Array<Dynamics::Context> dynamicsContexts_ = Array<Dynamics::Context>::Empty();
     mutable NumericalSolver numericalSolver_;
 
-    void registerDynamicsContext(const Shared<Dynamics>& aDynamicsSPtr);
+    void validateDynamicsSet() const;
 };
 
 }  // namespace trajectory
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
@@ -412,12 +412,6 @@ void Propagator::validateDynamicsSet() const
             throw ostk::core::error::RuntimeError("Invalid Dynamics Set.");
         }
     }
-
-    // Special case to disallow both TabulatedDynamics and ThrusterDynamics
-    if (currentDynamicsSets["Tabulated"] > 0 && currentDynamicsSets["Thruster"] > 0)
-    {
-        throw ostk::core::error::RuntimeError("Invalid Dynamics Set.");
-    }
 }
 
 }  // namespace trajectory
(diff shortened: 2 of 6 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -315,27 +361,57 @@ Propagator Propagator::FromEnvironment(const NumericalSolver& aNumericalSolver,
...
 }
 
-void Propagator::registerDynamicsContext(const Shared<Dynamics>& aDynamicsSPtr)
+void Propagator::validateDynamicsSet() const
 {
-    // Store read coordinate subsets information
-    Array<Pair<Index, Size>> readInfo = Array<Pair<Index, Size>>::Empty();
-    readInfo.reserve(aDynamicsSPtr->getReadCoordinateSubsets().getSize());
-    for (const Shared<const CoordinateSubset>& subset : aDynamicsSPtr->getReadCoordinateSubsets())
+    Map<String, Size> currentDynamicsSets = {
```
phc1990, line 366: ...continuing from the first comment of the review:

1. Just a quick thing, if you need to use the type of instances for your logic, you might want to use `type_index`, which should be cleaner (imo) and you don't need to "hardcode" class names. For example:

```cpp
#include <typeindex>

std::map<std::type_index, int> typeCount;
```

Then you can keep track of number of instance types:

```cpp
for (const auto& dynamics : this->dynamics) {
  ++typeCount[typeid(*dynamics)];
}
```

I had to search the syntax online, but the idea is not to hardcode class names (as those can change) but use Types instead.

2. This is the actual continuation of the first comment in the review: I think rather than looping and trying to find a common rule-set, maybe we can just implement here the logic. It's ok if we can't loop over a map, this is not to be re-used and you can have something like this (be warned this is pseudo-code):

```
// Keep track of celestial bodies for which gravity has been considered
gravityCelestialBodies = []


// Check there is only one CentralBodyGravity
if typeCouns[typeid(CentralBodyGravity)] != 1 {
  throw error...
} 

// Find and add the celestial body of the CentralBodyGravity
gravityCelestialBodies.add(centralBodyGravity.getCelestialBody())

// Check that the atmosphere, if any is the same as the Gravity Central Body
if centralBodyGravity.getCelestialBody() != atmosphericDrag.getCelestialBody() {
  throw error...
}

// Check there is one ThrusterDynamics mas
if typeCounts[typeid(ThrusterDynamics)] > 1 {
   throw error...
}

// Check no duplications amongst gravity bodys
for thirdBodyGravity in Dynacmis {
  if thirdBodyGravity.getCelestialBody() in gravityCelestialBodies {
    throw error...
  } else {
    gravityCelestialBodies.add(thirdBodyGravity.getCelestialBody();
  }
}

// Etc.
```

Anyway, not saying we need to do all of this now, that was just an example of ruleset. The message was that the ruleset is going to be heterogeneous enough for us to simply code it. 

⚠️ Also -> keep in mind that `Dynamics` abstract class was designed so that users can write their own implementation and simply provide it to the propagator and make full use of the rest of astro. This is just a reminder that we can validate as much as we want, but it's not 100% guaranteed that the dynamics set makes sense. `Propagator` is probably to be moved to Physics in the future, together with `Dynamics`. The validation we are doing here is astro-related, which is fine for now, and will probably be kept within the astro package in the future.

phc1990 (commented):
Nice stuff. Left some comments 👍

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
phc1990: Just an idea, if it becomes unmanageable, then we can always revert 👍

apaletta3: I did what you suggested and its not super clean, and it also doesn't fully cover all the cases above, but that's fine because in some cases you can't even create a ThirdBodyDynamics for Earth for example.

apaletta3: But I think its definitely better than what I had before and its good enough for now. Lmk what you think

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -53,18 +59,26 @@ using ostk::physics::coordinate::Velocity;
...
    public:
+    /// @brief Default integrator frame
     static const Shared<const Frame> IntegrationFrameSPtr;
+    /// @brief Map of valid dynamics set for propagation
+    ///
+    /// @code{.cpp}
+    ///              Map<String, Pair<Size, Size>> validDynamicsSet ={"TypeOfDynamics", {Min number allowed, Max number
+    ///              allowed}};
+    /// @endcode
+    static const Map<String, Pair<Size, Size>> ValidDynamicsSet;  // Map<String, Pair<Size, Size>>
```
phc1990, line 81: Maybe leftover?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -315,27 +355,58 @@ Propagator Propagator::FromEnvironment(const NumericalSolver& aNumericalSolver,
...
+        if (dynamicsMap.at(std::type_index(typeid(CentralBodyGravity))).getSize() != 1)
+        {
+            throw ostk::core::error::RuntimeError("Propagator needs exactly one Central Body Gravity Dynamics.");
+        }
+        if (dynamicsMap.at(std::type_index(typeid(PositionDerivative))).getSize() != 1)
+        {
+            throw ostk::core::error::RuntimeError("Propagator needs exactly one Position Derivative Dynamics.");
+        }
+    }
+    catch (const std::out_of_range& e)
```
phc1990, line 379: Nit: you can use map's `.find()` to check if a key exists, this way you avoid nesting/repeating error messages:

std::map<int, std::string> myMap;

```cpp
    // Insert some key-value pairs into the map
    myMap[1] = "One";
    myMap[2] = "Two";
    myMap[3] = "Three";

    // Check if the map contains a key
    int keyToFind = 2;
    auto it = myMap.find(keyToFind);
    if (it != myMap.end()) {
        std::cout << "Key " << keyToFind << " found in the map. Value: " << it->second << std::endl;
    } else {
        std::cout << "Key " << keyToFind << " not found in the map." << std::endl;
    }
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -315,27 +355,58 @@ Propagator Propagator::FromEnvironment(const NumericalSolver& aNumericalSolver,
...
-    this->dynamicsContexts_.add({aDynamicsSPtr, readInfo, writeInfo});
+    // Check for optional dynamics number limits
+    try
+    {
+        if (dynamicsMap.at(std::type_index(typeid(AtmosphericDrag))).getSize() > 1)
+        {
+            throw ostk::core::error::RuntimeError("Propagator can have at most one Atmospheric Drag Dynamics.");
+        }
+    }
+    catch (const std::out_of_range& e)
```
phc1990, line 394: You wouldn't need this either

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -315,27 +355,58 @@ Propagator Propagator::FromEnvironment(const NumericalSolver& aNumericalSolver,
...
+    }
+
+    try
+    {
+        if (dynamicsMap.at(std::type_index(typeid(Thruster))).getSize() > 1)
+        {
+            throw ostk::core::error::RuntimeError("Propagator can have at most one Thruster Dynamics.");
+        }
+    }
+    catch (const std::out_of_range& e)
```
phc1990, line 406: nor this : )

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -841,3 +915,134 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, FromEnv
         EXPECT_TRUE(static_cast<const PositionDerivative&>(*propagator.getDynamics()[4]).isDefined());
     }
 }
+
+TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, ValidateDynamicsSet)
```
phc1990, line 919: Very nice, atomic, isolated testing 👍

phc1990 (approved):
Nice 👍 
Left some nit comments: imo, it can be merged once addressed 👍

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
apaletta3: Lol that's what I did at first, and then I undid it and replaced it with `try catch` cause I thought it made the code more readable 😢 😭 .

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
apaletta3: good catch

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -841,3 +915,134 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, FromEnv
...
+                    {
+                        EXPECT_EQ("Propagator needs exactly one Position Derivative Dynamics.", e.getMessage());
+                        throw;
+                    }
+                },
+                ostk::core::error::RuntimeError
+            );
+        }
+
+        // Missing PositionDerivatives or CentralBodyGravity
```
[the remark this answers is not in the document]
apaletta3: > This probably deserves 2 tests, one for Missing PositionDerivatives whilst having CentralBodyGravity (which is what you have) and another for missing CentralBodyGravity (whilst having PositionDerivatives). This way we cover all the branches.

Done

> (Also, super nit: might be worth sorting test cases to reflect the order in which rules are checked in your actual function, this way is more readable when you have both files side-by-side)

that's what I did, no?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
apaletta3: But good idea, i'll use the iterator itself to check the value like you did in your example

PR #365, opened 2024-03-07
refactor!: tabulated dynamics getters into accessors only and add a specific getter for returning a matrixxd from coordinate subsets

apaletta3:
This MR is in response to @phc1990 's comment [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/340#discussion_r1500867108) about adding a `getContributionProfileFromCoordinateSubsets` to TabulatedDynamics as a stopgap to properly refactoring the way Dynamics are handled (and expanded and reduced) in the future.

Also I removed the redundant getters in `TabulatedDynamics` to keep with the pattern:
- accessors for member variables
- getters for member variables that you modify before returning or anything computed on the fly

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
@@ -69,10 +69,10 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
             )
 
             .def(
-                "get_instants",
-                &Tabulated::getInstants,
+                "access_instants",
+                &Tabulated::accessInstants,
                 R"doc(
-                    Get the contribution instants.
+                    Access the contribution instants.
 
                     Returns:
                         list[Instant]: The contribution instants.
@@ -81,10 +81,10 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
             )
 
             .def(
-                "get_contribution_profile",
-                &Tabulated::getContributionProfile,
+                "access_contribution_profile",
+                &Tabulated::accessContributionProfile,
                 R"doc(
-                    Get the contribution profile.
+                    Access the contribution profile.
 
                     Returns:
                         np.ndarray: The contribution profile.
@@ -93,10 +93,26 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
             )
 
             .def(
-                "get_frame",
-                &Tabulated::getFrame,
+                "get_contribution_profile_from_coordinate_subsets",
+                &Tabulated::getContributionProfileFromCoordinateSubsets,
+                arg("coordinate_subsets"),
                 R"doc(
-                    Get the reference frame.
+                    Get the contribution profile corresponding to a subset of coordinates.
+
+                    Args:
+                        coordinate_subsets (list[CoordinateSubset]): The coordinate subsets.
+
+                    Returns:
+                        numpy.ndarray: The contribution profile.
+
+                )doc"
+            )
+
+            .def(
+                "access_frame",
+                &Tabulated::accessFrame,
+                R"doc(
+                    Access the reference frame.
 
                     Returns:
                         Frame: The reference frame.
--- a/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
@@ -40,12 +40,12 @@ Tabulated::Tabulated(
 
     if (aWriteCoordinateSubsets
             .map<Index>(
-                [](const auto& coordinatesSubset)
+                [](const auto& coordinateSubset)
                 {
-                    return coordinatesSubset->getSize();
+                    return coordinateSubset->getSize();
                 }
             )
-            .reduce(std::plus<Index>()) != (Index)aContributionProfile.cols())
+            .reduce(std::plus<Index>()) != static_cast<Index>(aContributionProfile.cols()))
     {
         throw ostk::core::error::RuntimeError(
             "Contribution profile must have the same number of columns as the sum of the sizes of the write "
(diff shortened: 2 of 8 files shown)


test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -207,24 +209,116 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Accessors)
...
+
+        EXPECT_EQ(
+            tabulated.getContributionProfileFromCoordinateSubsets(
+                {CartesianPosition::Default(), CartesianVelocity::Default(), CoordinateSubset::Mass()}
+            ),
+            extendedContributionProfile
+        );
     }
 
+    // Test differently ordered subsets
```
phc1990, line 285: ❤️

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -207,24 +209,116 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Accessors)
 
 TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, Getters)
 {
+    MatrixXd positionContributionsProfile_(defaultInstants_.getSize(), 3);
+    MatrixXd velocityContributionsProfile_(defaultInstants_.getSize(), 3);
+    MatrixXd massContributionsProfile_(defaultInstants_.getSize(), 1);
+    positionContributionsProfile_ << 1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0;
```
phc1990, line 215: I see that you didn't overlap numbers between subsets to test better.
Still I would make entirely unique, since we are extracting subsets in a different order than the one they were originally stored in the TabulatedDynamics instance. E.g.:

```cpp
positionContributionsProfile_ << 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0;
velocityContributionsProfile_ << -1.0, -2.0, -3.0, -4.0, -5.0, -6.0, -7.0, -8.0, -9.0, -10.0, -11.0, -12.0;
massContributionsProfile_ << 0.001, 0.002, 0.003, 0.004;
```

phc1990 (approved):
LGTM @apaletta3 
Left some minor comments. Feel free to resolve and merge 👍

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
apaletta3: good point

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
```diff
@@ -81,10 +81,10 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Dynamics_Tabulated(pybind11::module&
             )
 
             .def(
-                "get_contribution_profile",
-                &Tabulated::getContributionProfile,
+                "access_contribution_profile",
```
vishwa2710, line 84: From our discussion, in python we are using the `get_*` keywords as we are not truly "accessing" the memory values by reference in python. If and when we implement that in the bindings we can have both get and access based methods if/where appropriate.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -40,12 +40,12 @@ Tabulated::Tabulated(
...
             .map<Index>(
-                [](const auto& coordinatesSubset)
+                [](const auto& coordinateSubset)
                 {
-                    return coordinatesSubset->getSize();
+                    return coordinateSubset->getSize();
                 }
             )
-            .reduce(std::plus<Index>()) != (Index)aContributionProfile.cols())
+            .reduce(std::plus<Index>()) != static_cast<Index>(aContributionProfile.cols()))
```
vishwa2710, line 48: For my knowledge, why is a static cast necessary here?

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -82,31 +82,69 @@ const Array<Instant>& Tabulated::accessInstants() const
...
+                [](const auto& coordinateSubset)
+                {
+                    return coordinateSubset->getSize();
+                }
+            )
+            .reduce(std::plus<Index>())
+    );
+
+    Index customMatrixColIndex = 0;
+    for (const auto desiredCoordinateSubsetSPtr : aCoordinateSubsetArray)
```
vishwa2710, line 112: for (const auto& desiredCoordinateSubsetSPtr : aCoordinateSubsetArray)

vishwa2710 (commented):
Great stuff, IMO removing the getters in C++ is fine, but changing the name in the python doesn't align with what we had discussed on slack.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
apaletta3: I think its more confusing to have a different name for the python binding than for the cpp method.

And there are already a bunch of methods called `access_*` in python in OSTk already. So I think its best to keep the name consistent between cpp and python

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: Because its safer and more "searchable" and expresses the programmer's intention more clearly than a c-style cast

https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting
https://www.reddit.com/r/cpp_questions/comments/f26n5x/why_use_static_cast_over_c_style_cast_for/

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Dynamics/Tabulated.cpp
vishwa2710: Moving the conversation to slack.

PR #373, opened 2024-03-22
feat: add maneuver cross and self validation

apaletta3:
Part of Maneuver validation effort.

This MR:
- [x] Adds self validation of tabulated dynamics and maneuver to make sure they produce similar results
- [x] Adds cross validation of tabulated dynamics and maneuver to make sure they produce similar results

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
@@ -6,6 +6,7 @@
 #include <OpenSpaceToolkit/Core/Type/Index.hpp>
 
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateBroker.hpp>
 
 namespace ostk
 {
@@ -17,6 +18,10 @@ namespace dynamics
 using ostk::core::type::Index;
 
 using ostk::mathematics::object::VectorXd;
+
+using ostk::astrodynamics::trajectory::state::CoordinateBroker;
+
+const Shared<const Frame> Tabulated::DefaultContributionFrameSPtr = Frame::GCRF();
 
 Tabulated::Tabulated(
     const Array<Instant>& anInstantArray,
@@ -51,6 +56,11 @@ Tabulated::Tabulated(
             "Contribution profile must have the same number of columns as the sum of the sizes of the write "
             "coordinate subsets."
         );
+    }
+
+    if (aFrameSPtr != DefaultContributionFrameSPtr)
+    {
+        throw ostk::core::error::RuntimeError("Contributions must be expressed in an inertial frame.");
     }
 
     VectorXd timestamps(anInstantArray.getSize());
@@ -170,11 +180,11 @@ VectorXd Tabulated::computeContribution(
     const Instant& anInstant, [[maybe_unused]] const VectorXd& x, const Shared<const Frame>& aFrameSPtr
 ) const
 {
-    // TBM: Convert this using the Maneuver class' conversion method. Also, find a way to check and vet whether or not
-    // the frame is local, or quasi-inertial
+    // TBM: Allow frame conversion through `CoordinateSubset.inFrame` method, once we have a `CartesianAcceleration`
+    // class
     if (aFrameSPtr != frameSPtr_)
     {
-        throw ostk::core::error::runtime::Wrong("Frame");
+        throw ostk::core::error::RuntimeError("Contribution Frame conversion to non-inertial not yet supported.");
     }
 
     if (anInstant < instants_.accessFirst() || anInstant > instants_.accessLast())
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
@@ -2,16 +2,24 @@
 
 #include <typeindex>
 
+#include <OpenSpaceToolkit/Core/Container/Map.hpp>
+#include <OpenSpaceToolkit/Core/Container/Pair.hpp>
 #include <OpenSpaceToolkit/Core/Error.hpp>
+#include <OpenSpaceToolkit/Core/Type/Index.hpp>
 #include <OpenSpaceToolkit/Core/Utility.hpp>
+
+#include <OpenSpaceToolkit/Mathematics/Object/Vector.hpp>
+
+#include <OpenSpaceToolkit/Physics/Coordinate/Position.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Object/Celestial.hpp>
 
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/AtmosphericDrag.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/CentralBodyGravity.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/PositionDerivative.hpp>
-#include <OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.hpp>
-#include <OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/Thruster.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.hpp>
 
 namespace ostk
 {
@@ -20,17 +28,21 @@ namespace astrodynamics
 namespace trajectory
 {
 
+using ostk::core::container::Map;
+using ostk::core::container::Pair;
+using ostk::core::type::Index;
+
 using ostk::mathematics::object::VectorXd;
 
+using ostk::physics::coordinate::Position;
 using ostk::physics::environment::object::Celestial;
 
 using ostk::astrodynamics::dynamics::AtmosphericDrag;
 using ostk::astrodynamics::dynamics::CentralBodyGravity;
 using ostk::astrodynamics::dynamics::PositionDerivative;
-using TabulatedDynamics = ostk::astrodynamics::dynamics::Tabulated;
-using ostk::astrodynamics::dynamics::ThirdBodyGravity;
 using ostk::astrodynamics::dynamics::Thruster;
 using ostk::astrodynamics::trajectory::state::CoordinateSubset;
+using ostk::astrodynamics::trajectory::StateBuilder;
 
 const Shared<const Frame> Propagator::IntegrationFrameSPtr = Frame::GCRF();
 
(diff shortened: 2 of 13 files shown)


src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -190,6 +185,33 @@ VectorXd Tabulated::computeContribution(
...
+        }
+
+        Size coordinateSubsetIndex = 0;
+        for (const auto& coordinateSubset : writeCoordinateSubsets_)
+        {
+            if (coordinateSubset->getSize() == 3)
+            {
+                contribution.segment(coordinateSubsetIndex, coordinateSubsetIndex + 3) =
+                    frameSPtr_->getTransformTo(aFrameSPtr, anInstant)
+                        .applyToVector(contribution.segment(coordinateSubsetIndex, coordinateSubsetIndex + 3));
```
vishwa2710, line 209: I don't think will apply the correct transformations to position and velocity?
https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/src/OpenSpaceToolkit/Physics/Coordinate/Transform.cpp#L311
https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/src/OpenSpaceToolkit/Physics/Coordinate/Transform.cpp#L333

Instead we should use the internally defined `inFrame` method for the subset to transform the frame in my opinion.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -49,9 +49,30 @@ Propagator::Propagator(
...
     : Propagator(aNumericalSolver, aDynamicsArray)
 {
-    // TBM: in the future maybe sanitize the instants of each maneuver to make sure none of them are overalapping
     for (const Maneuver& maneuver : aManeuverArray)
     {
+        // Check if the maneuver's first and last instants overlap with any other maneuver
+        Size duplicateManeuverCounter = 0;
+        for (const Maneuver& otherManeuver : aManeuverArray)
+        {
+            if (maneuver != otherManeuver)  // Skip self-comparison
```
vishwa2710, line 58:
```suggestion
            if (maneuver != otherManeuver)
```
The comment makes it more confusing as you aren't skipping in that block.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -49,9 +49,30 @@ Propagator::Propagator(
...
+            {
+                if (maneuver.getInstants()[0] <=
+                        otherManeuver.getInstants()[otherManeuver.getInstants().getSize() - 1] &&
```
vishwa2710, line 61:
```suggestion
                        otherManeuver.getInstants().accessLast() &&
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -49,9 +49,30 @@ Propagator::Propagator(
...
-    // TBM: in the future maybe sanitize the instants of each maneuver to make sure none of them are overalapping
     for (const Maneuver& maneuver : aManeuverArray)
     {
+        // Check if the maneuver's first and last instants overlap with any other maneuver
+        Size duplicateManeuverCounter = 0;
+        for (const Maneuver& otherManeuver : aManeuverArray)
+        {
+            if (maneuver != otherManeuver)  // Skip self-comparison
+            {
+                if (maneuver.getInstants()[0] <=
```
vishwa2710, line 60:
```suggestion
                if (maneuver.getInstants().accessFirst() <=
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -49,9 +49,30 @@ Propagator::Propagator(
...
+                        otherManeuver.getInstants()[otherManeuver.getInstants().getSize() - 1] &&
+                    maneuver.getInstants()[maneuver.getInstants().getSize() - 1] >= otherManeuver.getInstants()[0])
```
vishwa2710, line 62:
```suggestion
                    maneuver.getInstants().accessLast() >= otherManeuver.getInstants().accessFirst())
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.cpp
```diff
@@ -49,9 +49,30 @@ Propagator::Propagator(
...
+            }
+            else
+            {
+                duplicateManeuverCounter++;
+            }
+        }
+        if (duplicateManeuverCounter != 1)
+        {
+            throw ostk::core::error::RuntimeError("Maneuvers cannot be duplicated.");
+        }
```
vishwa2710, line 75:
```suggestion

        if (duplicateManeuverCounter != 1)
        {
            throw ostk::core::error::RuntimeError("Maneuvers cannot be duplicated.");
        }
```

test/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.test.cpp
```diff
@@ -355,25 +355,54 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Dynamics_Tabulated, ComputeContribution)
...
+            };
+
+            testing::internal::CaptureStdout();
+
+            const VectorXd computedContribution =
+                tabulatedNonQuasiInertialFrame.computeContribution(defaultInstants_[0], x, defaultFrameSPtr_);
+
+            EXPECT_FALSE(testing::internal::GetCapturedStdout().empty());
+
+            EXPECT_NE(computedContribution, contributionProfile_.row(0));
```
vishwa2710, line 391: Instead of a NE check, we should do an Eq check with the correct expected rotation.

vishwa2710 (changes requested):

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: Good catch, agreed.

Although looking for that code I found this which seems like it should raise an error or at least check if the dimension of the coordinate subset it has is single dimensional before raising an error, cause it doesn't actually do the frame conversion, right?
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.cpp#L87-L90

apaletta3: Also, I don't think I need to change this [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/validate/self-and-cross-validation-of-maneuver-and-tabulated-dynamics-non-inertial-frames-addition/src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp#L308-L309), right? 

Since this will only ever be for converting accelerations (and no other coordinate subsets) from inertial to a non-inertial frame (and not the other way around).

Agree?

apaletta3: @vishwa2710 Turns out its not that simple, because here we're using the `CartesianVelocity` subset as a derivative and not as itself, so the `CartesianSubset.inFrame()` method expects a `CartesianPosition` subset within the coordinate vector to reference so that it can correctly apply the Coriolis term if going to or from a non-inertial frame. So it raises an error when that happens.

So I might have to:
- add a `applyToAcceleration` method in `Transform` [open-space-collective/open-space-toolkit-physics@main/src/OpenSpaceToolkit/Physics/Coordinate/Transform.cpp#L333](https://github.com/open-space-collective/open-space-toolkit-physics/blob/main/src/OpenSpaceToolkit/Physics/Coordinate/Transform.cpp?rgh-link-date=2024-04-04T00%3A34%3A16Z#L333), and take into account all terms for the acceleration transform from non-inertial to inertial
-  add another `CoordinateSubset` derived class for `CartesianAcceleration` that we can use to actually represent accelerations for cases like these, in order to use the `.inFrame` method. Orrr I could ski revert to the way I had it 

Butttt that's a decent amount of work, so it may be easier to just keep the error raising here for now and not allow conversion to other frames.

Also come to think of it, is my [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/validate/self-and-cross-validation-of-maneuver-and-tabulated-dynamics-non-inertial-frames-addition/src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp?rgh-link-date=2024-04-05T11%3A51%3A33Z#L308-L309) then technically wrong because I'm using `applyToVector` which doesn't take into account higher order effects (angular acceleration) of non inertial frames?

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
```diff
@@ -190,6 +188,35 @@ VectorXd Tabulated::computeContribution(
...
+        Index coordinateSubsetIndex = 0;
+        for (const auto& coordinateSubset : writeCoordinateSubsets_)
+        {
+            coordinateSubsetIndex = coordinatesBrokerSPtr->addSubset(coordinateSubset);
+
+            if (coordinateSubset->getSize() == 3)
+            {
+                contribution.segment(coordinateSubsetIndex, coordinateSubsetIndex + 3) =
+                    coordinateSubset->inFrame(anInstant, contribution, frameSPtr_, aFrameSPtr, coordinatesBrokerSPtr);
+            }
```
vishwa2710, line 216:
```suggestion
            contribution.segment(coordinateSubsetIndex, coordinateSubsetIndex + 3) =
                coordinateSubset->inFrame(anInstant, contribution, frameSPtr_, aFrameSPtr, coordinatesBrokerSPtr);
```
What's the reason we're only applying this to size 3?
IMO if it's a size 1 it can just return the unconverted scalar value that way we don't have to go around checking.

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
vishwa2710: Honestly yeah I think it's easier to just keep it all in inertial and raise an error, let the user convert.

vishwa2710: https://space.stackexchange.com/questions/58123/conversion-of-acceleration-in-inertial-frame-to-non-inertial-rotating-frame
https://space.stackexchange.com/questions/58123/conversion-of-acceleration-in-inertial-frame-to-non-inertial-rotating-frame
My intuition says that we need to account for the rotational acceleration.
But from the comments in the stack overflow you can just convert the acceleration using just the Rotation matrix. As we do for position.

Which is what we do in OSTk Physics when converting acceleration from ITRF to GCRF.
@kyle-cochran @lucas-bremond @Derollez @phc1990 What do you guys think here (specifically on converting accelerations from one frame to another)

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: @vishwa2710 because I think there may be an issue with the `CoordinateSubset` frame conversion.
[See my comment](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/373#discussion_r1553478373)

What do you think?

src/OpenSpaceToolkit/Astrodynamics/Dynamics/Tabulated.cpp
apaletta3: I have gone deep down this rabbit hole, and this explanation helps, but doesn't fully explain it sufficiently...
image


I do think that you can indeed convert accelerations without considering the fictitious forces that arise from non-inertial frames. I think this is only possible when the acceleration is being "applied inertially or freely" (I don't have a good way to describe this). Basically what I'm trying to say is that I think the fictitious forces only apply in the case of rigid body dynamics. But that's not a very conclusive answer and not rigorous at all...

There has to be a fundamental reason why we don't have to account for fictitious forces in our acceleration vector conversions from inertial to non-inertial (and vice versa) reference frames - otherwise OSTk would note line up with GMAT and STK so well...

vishwa2710 (approved):

PR #377, opened 2024-03-26
build: arm64 with debug symbols off

apaletta3:


.github/workflows/build-test.yml
apaletta3, line 1: Change to be removed before merge into main

guillaumeloftorbital (approved):

PR #384, opened 2024-04-05
refactor!: remove local orbital frame factory parent frame getter and fix accessor

phc1990:


phc1990 (commented):
LGTM

We should mark it as a breaking change `!` since this was already released.

apaletta3 (approved):

PR #385, opened 2024-04-05
fix: converters coerce_to_iso typing

phc1990:


apaletta3 (approved):
Would be good to add type checking like mypy to OSTk python to catch this. I'll add an issue for this

PR #387, opened 2024-04-08
fix: coordinate subset unsafe frame conversion

apaletta3:
Users can create a `CoordinateSubset`, giving it a specific name and number of coordinates in the subset. However, if they use the `inFrame` conversion method on the `CoordinateSubset` class itself, it doesn't actually do the conversion they want, it just returns the coordinates as they are.

This is fine for scalar subsets, but not fine for vector coordinate subsets because it is a hidden behavior the user may not be aware of. Therefore, if someone tries to do this, I will raise an error.

If the user wants to make their own non-scalar coordinate subset by inheriting from `CoordinateSubset`, then they can get around this error by acutally implementing the `virtual` `inFrame` method and overriding it in their child class.

### DIFF
--- a/test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.test.cpp
+++ b/test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.test.cpp
@@ -132,6 +132,39 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset, InFrame
 
         EXPECT_EQ(expected, actual);
     }
+
+    // Attempt to transform a non-scalar subset without implementing the inFrame method
+    {
+        const Shared<const CoordinateSubset> nonScalarCoordinateSubset = std::make_shared<CoordinateSubset>("NAME", 3);
+        const Shared<const CoordinateBroker> nonScalarCoordinateBroker =
+            std::make_shared<CoordinateBroker>(Array<Shared<const CoordinateSubset>>({nonScalarCoordinateSubset}));
+
+        const Instant instant = Instant::J2000();
+        const Shared<const Frame> fromFrame = Frame::GCRF();
+        const Shared<const Frame> toFrame = Frame::TEME();
+        VectorXd fullCoordinatesVector(3);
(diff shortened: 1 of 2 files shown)


test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.test.cpp
```diff
@@ -132,6 +132,39 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset, InFrame
...
+        const Shared<const CoordinateBroker> nonScalarCoordinateBroker =
+            std::make_shared<CoordinateBroker>(Array<Shared<const CoordinateSubset>>({nonScalarCoordinateSubset}));
+
+        const Instant instant = Instant::J2000();
+        const Shared<const Frame> fromFrame = Frame::GCRF();
+        const Shared<const Frame> toFrame = Frame::TEME();
+        VectorXd fullCoordinatesVector(3);
+        fullCoordinatesVector << 1.0e7, -1e7, 5e6;
+
+        EXPECT_THROW(
```
phc1990, line 148: Nit: don't most of other tests simply assert that an exception is thrown?

Nothing wrong with asserting the message string. I'm not a big fan because it makes the test much longer and it involves reworking the test for a tangential change in an error message.

phc1990 (approved):

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.test.cpp
apaletta3: Not necessarily, I think its more common to assert the full error message because that way you're sure its the correct error being thrown.

And if you make a change to the cpp code that throws the error you can just ctrl+f to change that message across the repo, easy.

Check other examples here
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/test/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.test.cpp
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/blob/main/test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp#L1211-L1355

PR #389, opened 2024-04-11
fix: LVLHGDGT orientation

vishwa2710:
This was aligned with anti-velocity, while it should be aligned with velocity. I messed it up cause STK defines it the other way.


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -704,10 +704,10 @@ Shared<const Frame> Orbit::getOrbitalFrame(const Orbit::FrameType& aFrameType) c
                              std::pow(std::cos(inclination.inRadians()), 2.0);
                 temp1 = temp1 > 0.0 ? std::sqrt(temp1) : 0.0;
                 const Real temp2 = k - std::cos(inclination.inRadians());
-                const Real delta = (isPassDescending ? -1.0 : 1.0) * std::atan2(temp1, temp2) + Real::Pi();
+                const Real delta = (isPassDescending ? 1.0 : -1.0) * std::atan2(temp1, temp2);
 
                 const Quaternion q_LVLHGDGT_LVLHGD =
-                    Quaternion::RotationVector(RotationVector::X(Angle::Radians(delta))).toConjugate();
+                    Quaternion::RotationVector(RotationVector::X(Angle::Radians(delta)));
```
kyle-cochran, line 710: Doesn't removing the conjugation have the effect or reversing the transformation? i.e. instead of:

`LVLHGDGT` -> `LVLHGD`

it is:
`LVLHGD` -> `LVLHGDGT`

?

kyle-cochran (approved):

PR #402, opened 2024-06-03
feat: make coordinate subsets hashable

apaletta3:
This will allow us to use `CoordinateSubset` as a key in a python dictionary in our services

### DIFF
--- a/bindings/python/test/trajectory/state/test_coordinate_subset.py
+++ b/bindings/python/test/trajectory/state/test_coordinate_subset.py
@@ -30,6 +30,9 @@ class TestCoordinateSubset:
     def test_ne(self, coordinate_subset: CoordinateSubset):
         assert (coordinate_subset != coordinate_subset) == False
 
+    def test_hash(self, coordinate_subset: CoordinateSubset):
+        assert hash(coordinate_subset) is not None
+
     def test_get_id(self, coordinate_subset: CoordinateSubset):
         assert coordinate_subset.get_id() is not None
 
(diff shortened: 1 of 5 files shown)


bindings/python/test/trajectory/state/test_coordinate_subset.py
```diff
@@ -30,6 +30,9 @@ def test_eq(self, coordinate_subset: CoordinateSubset):
     def test_ne(self, coordinate_subset: CoordinateSubset):
         assert (coordinate_subset != coordinate_subset) == False
 
+    def test_hash(self, coordinate_subset: CoordinateSubset):
+        assert hash(coordinate_subset) is not None
```
bodzas-loft, line 34: Should we pop in a couple coordinates to verify behavior?

To cover the cases of:
- Different key, value insertions can be grabbed as intended
- Different key objects but `==` and `hash` produce true and the same value, insert both and insure only the value for the second instance.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.test.cpp
```diff
@@ -75,6 +75,13 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset, NotEqua
     }
 }
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset, Hash)
+{
+    {
+        EXPECT_NO_THROW(defaultCoordinateSubset_.hash());
+    }
```
vishwa2710, line 82: We should add some test cases for this like the ones Austin mentioned above.

vishwa2710 (commented):
Looks great!
Instead of using a specific method, we should make our class hashable like:
```
namespace std {
    template <>
    struct hash<CoordinateSubset> {
        Size operator()(const CoordinateSubset& aCoordinateSubset) const {
            return std::hash<String>()(aCoordinateSubset.id_);
        }
    };
}
```

And then update the bindings as:
```
        .def("__hash__", [](const CoordinateSubset& aCoordinateSubset) {
            return std::hash<CoordinateSubset>()(aCoordinateSubset);
        });
```

This is a more generic pattern, that also allows us to hash the C++ class and use it in maps etc. (which will be required when we move some of the python code to C++)

PR #407, opened 2024-06-07
fix: prevent State addition and subtraction based on coordinates broker not being equal

phc1990:
We are not performing subset-to-subset +/- operations at the moment whenever a State +/- operation is carried out.

This MR fails when attempting to perform a +/- operations between states where the underlying coordinate subsets are not in the same and in the same order.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
@@ -230,9 +230,9 @@ State State::operator+(const State& aState) const
         throw ostk::core::error::runtime::Wrong("Frame");
     }
 
-    if (!this->coordinatesBrokerSPtr_ != aState.coordinatesBrokerSPtr_)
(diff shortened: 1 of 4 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -230,9 +230,9 @@ State State::operator+(const State& aState) const
         throw ostk::core::error::runtime::Wrong("Frame");
     }
 
-    if (this->getSize() != aState.getSize())
+    if (this->coordinatesBrokerSPtr_ != aState.coordinatesBrokerSPtr_)
```
vishwa2710, line 233: So for now, this will also check that the coordinates are in the exact same order yet?

vishwa2710 (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
phc1990: correct

apaletta3 (approved):

PR #410, opened 2024-06-12
feat: add backward propagation get pass with rev number

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -185,192 +185,217 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+            .accessCoordinates()
+            .z();
+    };
+
+    while ((!currentPass.isDefined()) || (currentPass.getRevolutionNumber() != aRevolutionNumber))
+    {
+        Instant previousInstant = Instant::Undefined();
+
+        Duration stepDuration = aStepDuration;
+        if (currentPass.isDefined() && currentPass.isComplete())
```
kyle-cochran, line 256: Small Q:

Is the `currentPass.isComplete()` still necessary? If we don't save incomplete passes, then can't we assume that if a pass is defined that it is also complete?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -185,192 +185,217 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+    };
+
+    const auto getZDot = [this, &epoch](const double& aDurationInSeconds) -> Real
+    {
+        return this->modelPtr_->calculateStateAt(epoch + Duration::Seconds(aDurationInSeconds))
+            .getVelocity()
+            .accessCoordinates()
+            .z();
+    };
+
```
kyle-cochran, line 250: Maybe here
```suggestion
    // If `currentPass` is undefined, then the `_passMap` is empty and we start filling it. Otherwise, fill the `_passMap` starting from the closest defined Pass until we reach the desired revolution number.
```

^ Since there are a lot of `if (currentPass.isDefined())` checks in the next part, I think it's super helpful to call out that it is all to handle the case where the pass map is empty.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -185,192 +185,217 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+                }
             }
+
+            previousStateCoordinates_ECI_z = currentStateCoordinates_ECI_z;
+            previousStateCoordinates_ECI_zdot = currentStateCoordinates_ECI_zdot;
+            previousInstant = currentInstant;
         }
 
-        if (currentPass.getRevolutionNumber() == aRevolutionNumber)
+        if (currentPass.isDefined())
```
kyle-cochran, line 345: Not asking for a change in this MR, but we might be able to simplify a bit by separating out the initial case into a separate check at the beginning:
```
if _passMap.is_empty() {
   // Do all of the flow that is necessary when currentPass is not defined
}
```

Which may let us separate that logic into a separate function. Additionally, if we separate the:
```
while(true) {
    // Fill out incomplete passes
}
```
into it's own function, could be re-used by both flows.

Again, not asking for this. Just tossing ideas around to make this algorithm a little more transparent.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -185,192 +185,217 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
-                    closestPasSPtr = &(lowerBoundMapIt->second);
-                }
+                closestPasSPtr = &(lowerBoundMapIt->second);
             }
         }
-        else if (this->passMap_.size() > 0)
+    }
+    else if (this->passMap_.size() > 0)
+    {
+        closestPasSPtr = &(this->passMap_.begin()->second);
```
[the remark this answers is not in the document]
vishwa2710: I've cleaned up this code significantly with examples. Let me know what you think.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: Good catch!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -179,54 +181,79 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+        // {5, 6, 9, 10} -> aRevolutionNumber=4 -> return 5
+
+        if (lowerBoundMapIt == this->passMap_.begin())
         {
-            closestPasSPtr = &(this->passMap_.begin()->second);
+            return this->passMap_.begin()->second;
         }
 
-        Pass currentPass = Pass::Undefined();
+        // Cloases revolution number is within the map
```
kyle-cochran, line 216:
```suggestion
        // Closest revolution number is within the map
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -179,54 +181,79 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+
+        const auto closestPassMapIt = std::prev(lowerBoundMapIt);
 
-        if (closestPasSPtr != nullptr)  // Pass with closest revolution number found
+        // {5, 6, 9, 10} -> aRevolutionNumber=8 -> return 9
+        // lowerBoundMapIt = 9, closestPassMapIt = 6
```
kyle-cochran, line 221:
```suggestion
        // {5, 6, 9, 10} -> aRevolutionNumber=7 -> return 6
        // lowerBoundMapIt = 9, closestPassMapIt = 6
```

Is this comment supposed to be switched with the other one? Since `closestPassMapIt` is the one before the revolution number

Feel free to close these if I'm just confused

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -179,54 +181,79 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+
+        if ((aRevolutionNumber - closestPassMapIt->first) < (lowerBoundMapIt->first - aRevolutionNumber))
         {
-            currentPass = *closestPasSPtr;
+            return closestPassMapIt->second;
         }
 
+        // {5, 6, 9, 10} -> aRevolutionNumber=7 -> return 6
+        // lowerBoundMapIt = 9, closestPassMapIt = 6
```
kyle-cochran, line 229:
```suggestion
        // {5, 6, 9, 10} -> aRevolutionNumber=8 -> return 9
        // lowerBoundMapIt = 9, closestPassMapIt = 6
```

kyle-cochran (approved):

PR #414, opened 2024-06-18
feat: add more pass computation methods

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,106 +1241,185 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
-        return tabulated.calculateStateAt(epoch + Duration::Seconds(aDurationInSeconds))
+        throw ostk::core::error::runtime::Undefined("Model");
+    }
+
+    if (!anInterval.isDefined())
+    {
+        throw ostk::core::error::runtime::Undefined("Interval");
+    }
+
+    Array<Pass> passes;
```
apaletta3, line 1312:
```suggestion
    Array<Pass> passes = Array<Pass>::Empty();
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.test.cpp
apaletta3, line 1: What are these changes for? Seems unrelated

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.test.cpp
```diff
@@ -570,6 +576,195 @@ TEST(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit, ComputePasses)
...
+            const Instant endInstant = Instant::DateTime(DateTime::Parse("2018-01-01 02:30:00"), Scale::UTC);
+            const Interval interval = Interval::Closed(startInstant, endInstant);
+
+            const Array<Pass> passes = Orbit::ComputePassesWithModel(keplerianModel, interval);
+
+            for (const auto &pass : passes)
+            {
+                EXPECT_TRUE(pass.isDefined());
+            }
+            EXPECT_EQ(passes.accessLast().getType(), Pass::Type::Partial);
```
apaletta3, line 638: I thought you wanted to drop support for partial passes, will this be in a followup MR?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,106 +1241,185 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
+    const RootSolver::Solution solution =
+        rootSolver.bisection(getValue, (previousInstant - anEpoch).inSeconds(), (currentInstant - anEpoch).inSeconds());
+
+    if (!solution.hasConverged)
+    {
+        throw ostk::core::error::RuntimeError("Root solver did not converge.");
+    }
+
+    return anEpoch + Duration::Seconds(solution.root);
+}
```
apaletta3, line 1296: perhaps keep the method order the same as in the hpp?

apaletta3 (changes requested):
I think there are a few things to polish up, but otherwise looks good!

bindings/python/requirements.txt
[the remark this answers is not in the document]
vishwa2710: Technically we need to apply to `mathematics` as that is the base repo that uses numpy + eigen conversions.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.test.cpp
vishwa2710: Sorry these snuck in, just small fixes to use `const` where appropriate.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,106 +1241,185 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
-                }
-            }
+            break;
+        }
+
+        const State currentState = aModel.calculateStateAt(currentInstant);
 
-            // Descending node
-            if ((previousPositionCoordinates_ECI.z() > 0.0) && (currentPositionCoordinates_ECI.z() <= 0.0))
+        const Vector3d previousPositionCoordinates_ECI = previousState.getPosition().accessCoordinates();
```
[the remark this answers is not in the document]
vishwa2710: We don't need to specify frame as the model state computation method returns in ECI. Adding frame conversions everywhere will become too tedious, I think at this low a level it is okay to assume this. If this ever changes for some reason (highly doubt it, we will never propagate in a fixed frame) our unit tests will explode so we're good.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -181,7 +181,7 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
 
     const std::lock_guard<std::mutex> lock {this->mutex_};
 
-    const auto getClosestPass = [this](const Integer& aRevolutionNumber) -> Pass
+    const auto getClosestPass = [this](const Integer& revolutionNumber) -> Pass
```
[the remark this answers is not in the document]
vishwa2710: This causes a warning where it shadows a previously defined variable :/

bindings/python/requirements.txt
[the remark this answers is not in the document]
apaletta3: Ah crap, you right

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,106 +1241,185 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
+
+    const Array<Pass> passes = Orbit::ComputePassesWithModel(tabulated, interval);
+
+    Array<Pair<Index, Pass>> passMap;
+    passMap.reserve(passes.getSize());
+
+    Index stateIndex = 0;
+    Index passIndex = 0;
+
+    Pass const* currentPassPtr = &passes[passIndex];
```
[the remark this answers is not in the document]
vishwa2710: Why would you use a shared or unique pointer here? We are not allocating memory here we are merely just getting a pointer to the element.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,106 +1241,185 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
+    if (!solution.hasConverged)
+    {
+        throw ostk::core::error::RuntimeError("Root solver did not converge.");
+    }
+
+    return anEpoch + Duration::Seconds(solution.root);
+}
 
-    const Instant& epoch = aStateArray.accessFirst().accessInstant();
+Array<Pass> Orbit::ComputePassesWithModel(const orbit::Model& aModel, const Interval& anInterval)
```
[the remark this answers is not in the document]
vishwa2710: The other one returns a pass map as it provides a list of states as input. Where the desired output is to also provide all the states that are within a pass. In this case we are given a model, so we can only return a list of passes.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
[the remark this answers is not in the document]
apaletta3: Ah ok good point, I thought you could just simply create a shared pointer to an object that already exists more simply, but it turns out you have to use a custom deleter or something like that which complicates things.

However, my point still stands, why use raw pointers unless absolutely necessary? I think we should avoid them unless we have a good reason to use them. In this case, I get that you're trying to not copy the pass at `passes[passIndex]` to a `currentPass` variable at each iteration of the loop, but I don't think this needs a pointer, why not just use the  `passes[passIndex]` itself like this:
```cpp
while (stateIndex < aStateArray.getSize())
{
    const Instant& stateInstant = aStateArray[stateIndex].accessInstant();

    if (passes[passIndex]->accessInstantAtPassBreak().isDefined() &&
        stateInstant >= passes[passIndex]->accessInstantAtPassBreak())
    {
        passMap.add({stateIndex, passes[passIndex]});
        ++passIndex;
    }

    stateIndex++;
}
```
Also, I feel like with all optimizations turned on the compiler probably optimizes out the "copying" to the currentPass variable on each iteration.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
apaletta3, line 184: Why not extract this lambda and make it its own private function so its easier to unit test

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -415,6 +318,55 @@ Pass Orbit::getPassWithRevolutionNumber(const Integer& aRevolutionNumber, const
...
+    {
+        throw ostk::core::error::runtime::Undefined("Interval");
+    }
+
+    Instant currentInstant = anInterval.getStart();
+
+    Integer revolutionNumber = Integer::Undefined();
+
+    // Check if the Pass is already within Pass map
+    for (const auto& passPair : this->passMap_)
```
apaletta3, line 338:
```suggestion
    for (const auto& passPair : passMap_)
```
Shouldn't need this-> here cause of the unique naming. Can apply to the rest of the file or just ignore this comment so that its consistent

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -1196,121 +1144,131 @@ Array<Pair<Index, Pass>> Orbit::ComputePasses(const Array<State>& aStateArray, c
...
+                (pass.accessInstantAtDescendingNode() - pass.accessInstantAtNorthPoint()),
+                (pass.accessInstantAtSouthPoint() - pass.accessInstantAtDescendingNode()),
+                (pass.accessInstantAtPassBreak() - pass.accessInstantAtSouthPoint()),
+            };
+            stepDuration = *std::min_element(durations.begin(), durations.end()) / 2.0;
         }
 
-        previousStatePtr = &state;
-        currentIndex++;
+        previousInstant = isForwardPropagated ? pass.accessInstantAtPassBreak() + Duration::Microseconds(1.0)
```
apaletta3, line 1264: Microsecond is a good enough time interval? It doesn't work with something like a nanosecond? or 10 nanoseconds?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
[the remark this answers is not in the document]
vishwa2710: Your solution would work fine, but it would result in copying the pass object, it also results in repeated indexing of the passes array, which is fine cause it's constant complexity.
However, I don't believe we should live in fear of pointers. Pointers like any other programming paradigm have their place. You can segfault your program and memory leak it without using raw pointers. You can segfault it with Smart pointers as well.
Instead I would suggest understanding how they work and when they are effective. We are utilizing C++ for it's ability to do precise memory management, in this case it's relatively unnoticeable as we're operating on 100s of passes at most, but if this pattern was copied for something else that is 1000s or more we would start to see the impact.
I'll update the code but it's food for thought.

apaletta3 (approved):
If you can address my comments, otherwise looks dope man

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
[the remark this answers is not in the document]
apaletta3: To me the cost benefit of trying to do precise memory management at that granular of a level is not worth it, especially with the impact it can have on dev velocity. To me, the reason we use C++ is for code performance, that's it, I wouldn't say that about precise memory management. Using C++ for OSTk impacts us negatively in virtually every other way, but we stick with it because it is so damn fast and we need it to be that fast.

Sure you can segfault without pointers, but screwing something up with pointers is easy to do if you don't 100% know what you're doing.

Also can't you do it with a non-const reference instead of a raw pointer?
```cpp 
Pass& currentPass = passes[passIndex];

while (stateIndex < aStateArray.getSize())
{
    const Instant& stateInstant = aStateArray[stateIndex].accessInstant();

    if (currentPass.accessInstantAtPassBreak().isDefined() &&
        stateInstant >= currentPass.accessInstantAtPassBreak())
    {
        passMap.add({stateIndex, currentPass});
        ++passIndex;
		currentPass = passes[passIndex]
    }

    stateIndex++;
}
```

Also also, are you sure the Pass object is not being copied at this step anyways? `passMap.add({stateIndex, *currentPassPtr});`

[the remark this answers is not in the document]
vishwa2710: > the reason we use C++ is for code performance
This doesn't come for free. You can write code that is not performant in C++. OSTk for example is atleast 1-3 orders of magnitude slower than propagators I've used in the past.

> Using C++ for OSTk impacts us negatively in virtually every other way, but we stick with it because it is so damn fast and we need it to be that fast.

? I'm not quite sure I understand what you mean here.

> Also also, are you sure the Pass object is not being copied at this step anyways? passMap.add({stateIndex, *currentPassPtr});
It's being copied there once as it has to be added to the map. It's not making an additional copy of the `currentPass`.

> I wouldn't say that about precise memory management
The hallmark of modern performance optimization **is** memory management.

You can have the exact same issues you have with a non-const reference that you can with a pointer. Why is that any "safer"? If you make a reference to something that goes out of scope you can segfault.
Either way. It's been amended.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: It would be a private function that would not be unit tested in either case. We're unit testing this by testing the public interface.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: I'd rather reformat the repo to either use `this` or not use `this` when referring to private variables consistently outside of this MR.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: Yeah unless we have passes that are less than 1 microsecond long. Which I can live with :P
This is more than sufficient, we could even do 1 second or 100 seconds and be fine.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
apaletta3: We do unit test our private functions though. But anyways feel free to resolve

PR #418, opened 2024-06-24
build: add option to build executable

vishwa2710:


CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ OPTION (BUILD_CODE_COVERAGE "Build code coverage" OFF)
 OPTION (BUILD_DOCUMENTATION "Build documentation" OFF)
 OPTION (BUILD_WITH_DEBUG_SYMBOLS "Build with debug symbols" ON)
 OPTION (BUILD_BENCHMARK "Build benchmark" ON)
+OPTION (BUILD_UTILITY "Build utility" ON)
```
vishwa2710, line 26:
```suggestion
OPTION (BUILD_UTILITY "Build utility" OFF)
```

CMakeLists.txt
```diff
@@ -394,6 +395,30 @@ IF (BUILD_STATIC_LIBRARY)
 
 ENDIF ()
 
+IF (BUILD_UTILITY)
+
+    SET (UTILITY_NAME "${PROJECT_PACKAGE_NAME}")
+    SET (UTILITY_TARGET "${UTILITY_NAME}.exe")
+
+    FILE (GLOB_RECURSE UTILITY_HEADERS "${PROJECT_SOURCE_DIR}/include/${PROJECT_PATH}/*.hpp")
+    FILE (GLOB_RECURSE UTILITY_SRCS "${PROJECT_SOURCE_DIR}/share/utils/*.cxx")
```
apaletta3, line 404: Do we ever put stuff in the /share dir? I've never seen it used

apaletta3 (changes requested):
Don't we need something in the makefile to set the env var that is used to name the utility? 
Right now I'm not sure how to use this build utility, so could you also include a little guide in the README and/or a make target that does this? like make `build-dev-utility` or something

CMakeLists.txt
vishwa2710: This is not in /share, but in the main project dir `/share/utils/*.cxx`
I can move it to another folder, I was just mirroring what we had in OSTk core.
Maybe in project_srouce_dir/scripts/

vishwa2710:
> Don't we need something in the makefile to set the env var that is used to name the utility? Right now I'm not sure how to use this build utility, so could you also include a little guide in the README and/or a make target that does this? like make `build-dev-utility` or something

Since we would typically build this from within a docker image, it would require a new `ostk-build` command in the base image. Which is fine, I can add it, but I think for now just using `ccmake` and setting the option to ON and then running `ostk-build` is sufficient. I can definitely update the README.

CMakeLists.txt
apaletta3: Yeah that's what I meant, I've never seen us use the main project dir /share before, should we depracate that folder? I do indeed prefer putting it in either in a /scripts dir or even in the /test dir

CMakeLists.txt
```diff
@@ -394,6 +395,30 @@ IF (BUILD_STATIC_LIBRARY)
 
 ENDIF ()
 
+IF (BUILD_UTILITY)
+
+    SET (UTILITY_NAME "${PROJECT_PACKAGE_NAME}")
+    SET (UTILITY_TARGET "${UTILITY_NAME}.exe")
```
apaletta3, line 401: Do you have to set this name somewhere `UTILITY_NAME`?

apaletta3 (approved):

CMakeLists.txt
vishwaLoft: This is setting the UTILITY_NAME to `open-space-toolkit-astrodynamics`

CMakeLists.txt
vishwaLoft: Yeah /scripts might be good.

PR #419, opened 2024-06-24
feat: add custom profile generators

vishwa2710:


include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -59,6 +74,35 @@ class Profile
...
+    struct Target
+    {
+        TargetType type;
+        Axis axis;
+        bool antiDirection = false;
+
+        Integer getSign() const
+        {
+            return antiDirection ? -1 : 1;
+        }
```
lucas-bremond, line 103: I wonder if we shouldn't that struct "pure" and have a simple inline function defined where `getSign` is used.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -173,10 +217,52 @@ class Profile
...
+    static Vector3d GetTargetDirectionVector(const State& state, const Trajectory& targetTrajectory);
+
+    static Vector3d GetClockingAxisVector(const Vector3d& alignmentAxisVector, const Vector3d& clockingVector);
+
+    static RotationMatrix GetRotationMatrix(
+        const Axis& alignmentAxis,
+        const Axis& clockingAxis,
+        const Vector3d& alignmentAxisVector,
+        const Vector3d& clockingAxisVector
+    );
```
lucas-bremond, line 265: Shouldn't we name the input `aState`, `anAlignmentAxis`, etc to follow OSTk's conventions?

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -173,10 +217,52 @@ class Profile
...
+    /// @param anAlignmentAxis An alignment axis
+    /// @param aClockingAxis A clocking axis
+    /// @param anAngularOffset An angular offset
+    /// @param aTrajectory A trajectory
+
+    static std::function<Quaternion(const State&)> GenerateCustomOrientation(
+        const Target& anAlignmentTarget,
+        const Target& aClockingTarget,
+        const Angle& anAngularOffset = Angle::Zero(),
+        const Trajectory& aTrajectory = Trajectory::Undefined()
```
lucas-bremond, line 240: What if we supplied two Targets with `type = Trajectory`? I think this should be associated with the Target struct itself to avoid the potential confusion.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -173,10 +230,48 @@ class Profile
...
+
+    static Vector3d GetTargetDirectionVector(const State& aState, const Trajectory& aTrajectory);
+
+    static Vector3d GetClockingAxisVector(const Vector3d& anAlignmentAxisVector, const Vector3d& aClockingVector);
+
+    static RotationMatrix GetRotationMatrix(
+        const Axis& anAlignmentAxis,
+        const Axis& aClockingAxis,
+        const Vector3d& anAlignmentAxisVector,
+        const Vector3d& clockingAxisVector
```
lucas-bremond, line 273:
```suggestion
        const Axis& anAlignmentAxis,
        const Axis& aClockingAxis,
        const Vector3d& anAlignmentAxisVector,
        const Vector3d& aClockingAxisVector
```

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -59,6 +74,48 @@ class Profile
...
+    };
+
+    enum class TargetType
+    {
+        GeocentricNadir,
+        GeodeticNadir,
+        Trajectory,
+        Sun,
+        Moon,
+        Velocity
```
lucas-bremond, line 91: Perhaps not worth delaying this PR, but thought: we could consider tracking the Velocity vector in ECI and Velocity vector in ECEF (which aren't pointing in the same direction, in an absolute sense, since the Earth is rotating).

Clocking against V_ECEF is for example necessary to support line scans.

Even if our implementation doesn't support this today, should our interface already embrace that case?

vishwa2710: Ah interesting, that's a good idea. Can atleast add the enums and a raise a ToBeImplemented error 👍

vishwa2710: Done!

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -59,6 +82,111 @@ class Profile
...
+        X,
+        Y,
+        Z
+    };
+
+    enum class TargetType
+    {
+        GeocentricNadir,
+        GeodeticNadir,
+        Trajectory,
```
lucas-bremond, line 96: This one isn't obvious to me, perhaps worth documenting?

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -59,6 +82,111 @@ class Profile
...
     };
 
+    enum class Axis
+    {
+        X,
+        Y,
+        Z
+    };
+
+    enum class TargetType
```
lucas-bremond, line 92: Let'd document each enum entry? (a one liner for each is fine)

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -59,6 +82,111 @@ class Profile
...
+
+            for (Index i = 0; i < Size(coordinates.cols()); ++i)
+            {
+                interpolators_.add(Interpolator::GenerateInterpolator(
+                    Interpolator::Type::BarycentricRational, timestamps, coordinates.col(i)
+                ));
+            }
+        }
+
+        Vector3d getAlignmentVectorAt(const Instant& anInstant) const
```
lucas-bremond, line 172: Could we move the implementation to the `.cpp` side to keep headers cleaner?

PR #427, opened 2024-08-05
refactor!: remove unnecessary converters and add 'Z' parsing

vishwa2710:
@phc1990 I removed some of the position/velocity coercion methods as I feel like they are do not reduce/simplify the creation as you have to pass every single parameter to a 1 line static method.

### DIFF
--- a/bindings/python/tools/python/ostk/astrodynamics/converters.py
+++ b/bindings/python/tools/python/ostk/astrodynamics/converters.py
@@ -1,18 +1,10 @@
 # Apache License 2.0
 
-from datetime import datetime, timedelta, timezone
-
-import numpy as np
-
-from ostk.mathematics.geometry.d3.transformation.rotation import Quaternion
+from datetime import datetime, timezone
 
 from ostk.physics.time import Instant
 from ostk.physics.time import Interval
-from ostk.physics.time import Duration
 from ostk.physics.time import Scale
-from ostk.physics.coordinate import Position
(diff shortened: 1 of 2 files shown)


bindings/python/tools/python/ostk/astrodynamics/converters.py
```diff
@@ -33,6 +25,9 @@ def coerce_to_datetime(value: Instant | datetime | str) -> datetime:
         return value.get_date_time(Scale.UTC).replace(tzinfo=timezone.utc)
 
     if isinstance(value, str):
+        if value.endswith("Z"):
+            value = value.replace("Z", "+00:00")
+
```
phc1990, line 30: Nit: Don't know if this is necessary, tests cases were already covering the "Z" scenario and they seemed fine (see screenshot below):

image

phc1990 (approved):
Left one nit comment, LGTM 👍

bindings/python/tools/python/ostk/astrodynamics/converters.py
vishwa2710: So this works if you include the milliseconds, but it doesn't work if you do:
`coerce_to_datetime("2024-01-01T00:00:00Z")`

vishwa2710: Ah that's because of python3.10 vs python3.11 compatibility. This should make it compatible both ways.

PR #429, opened 2024-08-20
feat: getBodyFrame and getAxesAt for Tabulated Profile

vishwa2710:


loft-nicolas-sanchez (approved):
Looks good to me - from a not so experienced OSTk contributor 😅

bindings/python/test/flight/test_profile.py
```diff
@@ -1,6 +1,7 @@
 # Apache License 2.0
 
 from datetime import datetime
+from pyexpat import model
```
vishwa2710, line 4:
```suggestion
```

PR #432, opened 2024-08-27
wip:

vishwa2710:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+
+using ostk::mathematics::object::Vector3d;
+
+using ostk::physics::coordinate::Position;
+using ostk::physics::coordinate::Transform;
+using ostk::physics::coordinate::Velocity;
+
+CartesianAcceleration::CartesianAcceleration(
+    const Shared<const CartesianPosition>& aCartesianPositionSPtr, const String& aName
```
phc1990, line 25: We have 2 options here imo:

- Option 1, we provide no references to other subsets AND we dont support inFrame conversion of the acceleration subset
- Option 2, we provide all required subsets to support the inFrame conversion (position, velocity)

There is an option 3, which is what I think we should do (at least for now):
- Option 3, we provided the required subsets (position, velocity) but may raise a NonImplemented error in the inFrame function.

⚠️ see the other comment in the inFrame function

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+    const Vector3d accelerationCoordinates =
+        Vector3d::Map(aCoordinateBrokerSPtr->extractCoordinate(aFullCoordinatesVector, *this).data());
+
+    // As per
+    // https://space.stackexchange.com/questions/58123/conversion-of-acceleration-in-inertial-frame-to-non-inertial-rotating-frame
+    // acceleration due to gravity is the same in all Newtonian frames
+    const Transform transform = fromFrame->getTransformTo(toFrame, anInstant);
+
+    return VectorXd::Map(transform.applyToVector(accelerationCoordinates).data(), Eigen::Index(3));
```
phc1990, line 74: What we are doing here is simply rotating the acceleration vector into another frame.

Although it is probably intended this is not what other inFrame functions do (see the Cartesian Velocity one for instance). The inFrame function expresses the coordinates with respect to the new frame (i.e. as seen from that frame).

So as you pointed out in the link this would involve:
- position -> This is already included in your MR as part of the subset constructor
- velocity -> This should be included in your MR as part of the subset constructor
- frame-to-frame angular velocity -> this is included in the Physics Transform
- frame-to-frame angular acceleration -> this is not present in the Physics Transform

Imo, this should be something provided by the Transform, which already provides
- applyToVector
- applyToVelocity
- but does not provide applyToAcceleration (and currently can't as it's not storing angular acceleration)

src/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.cpp
```diff
@@ -257,62 +277,14 @@ void Maneuver::print(std::ostream& anOutputStream, bool displayDecorator) const
...
-        massFlowRateProfile.add(contributionProfile(i, 3));
-    }
-
     return {
-        aTabulatedDynamics.accessInstants(),
-        accelerationProfile,
-        aTabulatedDynamics.accessFrame(),
-        massFlowRateProfile,
+        aStateArray,
+        Array<Real>(aStateArray.getSize(), aMassFlowRate),
```
phc1990, line 284: If we create the MFR subset, here we can simply use a Builder to create a new expanded array of states and pass it to the constructor.

phc1990 (commented):
This looks great, so happy that we are going in this direction.
Left comments on two areas:
- going fully state-only (by creating a MFR subset)
- on the inFrame function

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -519,59 +531,65 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, AddDyna
...
-        overlappingManeuverInstants.add(
-            instant + defaultManeuver_.getInterval().getDuration() - Duration::Seconds(1.0)
-        );
-    }
-
-    const Maneuver overlappingManeuver = {
-        overlappingManeuverInstants, defaultManeuverAccelerationProfile_, gcrfSPtr_, defaultManeuverMassFlowRateProfile_
-    };
-}
+// TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, AddManeuvers)
```
alex-liang3, line 534: Just a reminder to fix/remove these

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -68,17 +69,12 @@ class Maneuver
...
     /// @param aFrameSPtr A frame in which the acceleration profile is defined
     /// @param aMassFlowRateProfile A mass flow rate profile of the maneuver (negative numbers expected), one Real per
     /// instant in kg/s
-    Maneuver(
-        const Array<Instant>& anInstantArray,
-        const Array<Vector3d>& anAccelerationProfile,
-        const Shared<const Frame>& aFrameSPtr,
-        const Array<Real>& aMassFlowRateProfile
-    );
+    Maneuver(const Array<State>& aStateArray, const Array<Real>& aMassFlowRateProfile);
```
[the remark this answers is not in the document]
apaletta3: Agreed with Pau. also need to change this in python
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/432/files#diff-f68dddf49949cb68a74c863f4c27c8f9329d0ba9e6904af5a0a9b81e5013a7daR23

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -210,43 +206,21 @@ class Maneuver
...
-    /// rate
-    ///
-    /// @code{.cpp}
-    ///                  Maneuver maneuver = Maneuver::TabulatedDynamics(tabulated);
-    /// @endcode
-    ///
-    /// @param aTabulatedDynamics A TabulatedDynamics object
-    ///
-    /// @return A maneuver
-    static Maneuver TabulatedDynamics(const Tabulated& aTabulatedDynamics);
```
apaletta3, line 223: Perhaps we could even change the `TabulatedDynamics` creation interface to allow creating tabulated dynamics from a State or something? That way you could still support this static method cause I thought we used it quite often, no?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
apaletta3, line 534: The addManeuver method is super useful, why are we removing it?

apaletta3 (commented):
Nice, but would look to discuss some of the changes and removal of some useful methods in more detail

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
vishwa2710: We don't use it at all anywhere.
It's not possible to create a Maneuver from just Tabulated dynamics, you would have to specify the position velocity, while a tabulated dynamics for thrusters typically only contains thrust acceleration and mass flow rate.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
vishwa2710: @phc1990 I'm confused. As mentioned in the comment, we don't need to apply those conversions when rotating acceleration due to gravity.
This is what was done here as well: https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/432/files#diff-fcf1fa13079a2431c7ab797be79b5f0d062701794b152f860c3e37bb9f06b52eL299
And also what is done when converting between ECEF and ECI in the CentralBodyGravity dynamics.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
vishwa2710: We're not removing the method, these tests were commented out by mistake will add them back, some of the other tests were removed as we removed the FromTabulatedDynamics method.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
phc1990: Ok I see what you mean. I think you are correct.

Because we're using an acceleration as a way to come up with the **force** exerted upon the satellite.

So effectively you're only interested in rotating the vector (i.e. applyToVector()), which makes sense.

Ok, then I think what we should do is call this subset `CartesianManeuverAcceleration` subset, to differentiate it from the actual `CartesianAcceleration` (understood as the derivative of `CartesianVelocity`, which would actually have the "complex" inFrame transformation in the  comment above).

The CartesianManeuverAcceleration would not need position nor velocity in its constructor and the inFrame function would simply rotate as you're doing already.

LMK what you think

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
phc1990: Edit: [see this comment ](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/432/files#r1925839466)

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -210,43 +206,21 @@ class Maneuver
...
-    );
+    static Maneuver ConstantMassFlowRateProfile(const Array<State>& aStatearray, const Real& aMassFlowRate);
 
    private:
-    Array<Instant> instants_;
-    Array<Vector3d> accelerationProfileDefaultFrame_;
+    Array<State> states_;
     Array<Real> massFlowRateProfile_;
-
-    Array<Vector3d> convertAccelerationProfileFrame(const Shared<const Frame>& aFrameSPtr) const;
```
phc1990, line 249: Nit: do you think it would be useful to keep this (but using states), should be straight forward:

```
Maneuver inFrame(const Shared<const Frame>& aFrameSPtr) const;
```

Simply loops over the states, running inFrame and returns a new Maneuver.

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
```diff
@@ -109,14 +105,14 @@ class Maneuver
...
+    ///                  Array<State> states = maneuver.getStates();
     /// @endcode
     ///
-    /// @return The instants
-    Array<Instant> getInstants() const;
+    /// @return The states
+    Array<State> getStates() const;
 
     /// @brief Get the acceleration profile of the maneuver
     ///
```
phc1990, line 118: See my other comment about having a `Maneuver.inFrame(frame)` function.

Could we then get rid of these two?
- Array<Vector3d> getAccelerationProfile(...)
- Array<Real> getMassFlowRateProfile() const

Just checking if they are required (now that you're making the breaking change to use States)

If still required maybe we can get rid of the (aFrameSPTr) in the getAccelerationProfile at least?

So just this -> Array<Vector3d> getAccelerationProfile()

If desired in a specific frame one can do `maneuver.inFrame(myFrame).getAccelerationProfile`

Although once we have access to the states, then all of this is not strictly needed, hence my initial question : )

phc1990 (commented):
Nice, replied and added 2 nits

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
vishwa2710: Hmmmmm I agree we could rename it. However, this is also true for the central body gravitation as I mentioned. From the thread [here](https://space.stackexchange.com/questions/58123/conversion-of-acceleration-in-inertial-frame-to-non-inertial-rotating-frame), acceleration due to gravity is the same in all (Newtonian) frames. Should we keep a CartesianAcceleration and a GravitationalAcceleration?

include/OpenSpaceToolkit/Astrodynamics/Flight/Maneuver.hpp
vishwa2710: Agreed, got rid of it and instead will add an in frame method.

bindings/python/test/trajectory/test_propagator.py
```diff
@@ -25,13 +23,13 @@
 
 from ostk.astrodynamics.trajectory import LocalOrbitalFrameFactory
 from ostk.astrodynamics.trajectory import LocalOrbitalFrameDirection
-
 from ostk.astrodynamics.trajectory.state import CoordinateSubset
+from ostk.astrodynamics.trajectory.state import CoordinateBroker
```
vishwa2710, line 27:
```suggestion
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/NewtonianAcceleration.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+
+using ostk::mathematics::object::Vector3d;
+
+using ostk::physics::coordinate::Position;
+using ostk::physics::coordinate::Transform;
+using ostk::physics::coordinate::Velocity;
+
+NewtonianAcceleration::NewtonianAcceleration(
+    const Shared<const CartesianPosition>& aCartesianPositionSPtr, const String& aName
```
phc1990, line 25: No need for CartesianPosition pointer, just the name (probably copy-paste from CartesianVelocity)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/NewtonianAcceleration.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // acceleration due to gravity is the same in all Newtonian frames
+    const Transform transform = fromFrame->getTransformTo(toFrame, anInstant);
+
+    return VectorXd::Map(transform.applyToVector(accelerationCoordinates).data(), Eigen::Index(3));
+}
+
+Shared<const NewtonianAcceleration> NewtonianAcceleration::Default()
+{
+    static const Shared<const NewtonianAcceleration> cartesianAcceleration =
+        std::make_shared<NewtonianAcceleration>(CartesianPosition::Default(), "CARTESIAN_ACCELERATION");
```
phc1990, line 80:
```suggestion
    static const Shared<const NewtonianAcceleration> cartesianAcceleration =
        std::make_shared<NewtonianAcceleration>("CARTESIAN_ACCELERATION");
```

phc1990 (approved):
Left a minor one, approved!

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/NewtonianAcceleration.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    return cartesianAcceleration;
```
phc1990, line 82:
```suggestion
    return newtonianAcceleration;
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/NewtonianAcceleration.hpp
```diff
@@ -0,0 +1,111 @@
+/// Apache License 2.0
+
+#ifndef __OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset_CartesianAcceleration__
+#define __OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset_CartesianAcceleration__
```
phc1990, line 4:
```suggestion
#ifndef __OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset_NewtonianAcceleration__
#define __OpenSpaceToolkit_Astrodynamics_Trajectory_State_CoordinateSubset_NewtonianAcceleration__
```

phc1990 (approved):
Added 2 nits

apaletta3 (commented):
Partial review, see my message on slack

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
```diff
@@ -558,19 +563,6 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Propagator, AddMane
...
-    for (const Instant& instant : defaultManeuverInstants_)
-    {
-        overlappingManeuverInstants.add(
-            instant + defaultManeuver_.getInterval().getDuration() - Duration::Seconds(1.0)
-        );
-    }
-
-    const Maneuver overlappingManeuver = {
-        overlappingManeuverInstants, defaultManeuverAccelerationProfile_, gcrfSPtr_, defaultManeuverMassFlowRateProfile_
-    };
```
apaletta3, line 573: I guess this unit test was never actually finished, but it would be good to test that the propagator raises an error here (if its not done anywhere else)

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Validation/Propagator.cross.validation.cpp
```diff
@@ -768,7 +770,29 @@ TEST_P(OpenSpaceToolkit_Astrodynamics_Validation_CrossValidation_Thruster, Force
         propagatorWithTabulated.calculateStatesAt(initialState, instantArray);
 
     // Create maneuver from tabulated and propagator with maneuver
-    const Maneuver maneuver = Maneuver::TabulatedDynamics(tabulated);
+    Array<State> maneuverStateArray = Array<State>::Empty();
```
apaletta3, line 773: Thinking about this, we actually could have kept this `TabulatedDynamics` static method and just required that states with position and velocity coordinate subsets be provided along with it

apaletta3 (commented):
Great stuff! I still think we should remove the ThrustAcceleration coordinate subset because it isn't necessary and a little confusing to the user.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    /// @brief Returns the default CartesianAcceleration object
+    ///
+    /// @return Shared pointer to the default CartesianAcceleration object
+    static Shared<const CartesianAcceleration> Default();
+
+    /// @brief Returns the thrust acceleration
+    ///
+    /// @return Shared pointer to the thrust acceleration
+    static Shared<const CartesianAcceleration> ThrustAcceleration();
```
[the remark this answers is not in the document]
vishwa2710: This allows us to have multiple types of accelerations in a state, for example AtmosphericDrag, ThrustAcceleration etc.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianAcceleration.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+Shared<const CartesianAcceleration> CartesianAcceleration::ThrustAcceleration()
+{
+    static const Shared<const CartesianAcceleration> cartesianAcceleration = std::make_shared<CartesianAcceleration>(
+        CartesianPosition::Default(), CartesianVelocity::Default(), "THRUST_ACCELERATION"
+    );
+
+    return cartesianAcceleration;
+}
```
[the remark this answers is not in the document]
vishwa2710: It's not a separate subset, it's a definition of a CartesianAcceleration subset.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -201,25 +198,43 @@ Array<flightManeuver> Segment::Solution::extractManeuvers(const Shared<const Fra
...
+            VectorXd coordinates(10);
+            coordinates.segment<6>(0) = state.extractCoordinates({
+                CartesianPosition::Default(),
+                CartesianVelocity::Default(),
+            });
+            coordinates.segment<3>(6) = maneuverContributionBlock.block<1, 3>(i, 0);
+            coordinates(9) = fullSegmentContributions(startStopPair.first + i, 3);
+
+            maneuverStatesBlock.add(stateBuilder.build(state.accessInstant(), coordinates));
+        }
```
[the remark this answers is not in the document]
vishwa2710: Not sure what you mean? We're extracting the coordinates and creating a maneuverCoordinates block as per the coordinate subsets are defined above?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.test.cpp
vishwa2710: Yeah I just removed the incomplete test, feel free to add it back in since I believe you had worked on this test.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Validation/Propagator.cross.validation.cpp
vishwa2710: That is a strange tabulated dynamic where you're tabulating states. We could add it but I don't see any value, and it wasn't being used anywhere so not sure it's worth.

apaletta3: Yeah fair I guess that is a little awkward. In the future we may want to revisit this with our plans to include acceleration subset in all states. Because all accelerations are applied at a certain state

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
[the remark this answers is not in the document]
apaletta3:
```suggestion
            const State& state = this->states[startStopPair.first + i].inFrame(aFrameSPtr);

            VectorXd coordinates(stateBuilder.getSize());
            coordinates.segment<6>(0) = state.extractCoordinates({
                CartesianPosition::Default(),
                CartesianVelocity::Default(),
            });
            coordinates.segment<3>(6) = maneuverContributionBlock.block<1, 3>(i, 0);
            coordinates(9) = fullSegmentContributions(startStopPair.first + i, 3);

            maneuverStatesBlock.add(stateBuilder.build(state.accessInstant(), coordinates));
        }
```

With my [PR](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/504) we can now do this. But i guess it is a little cumbersome to do that for each individual subsets later... so anyways feel free to resolve

PR #434, opened 2024-09-10
feat!: simplify orbit constructor and State conversion to COE

vishwa2710:
```
from ostk.physics.environment.object.celestial import Earth
from ostk.astrodynamics.trajectory import Orbit
from ostk.astrodynamics.trajectory.orbit.model.sgp4 import TLE
from ostk.astrodynamics.trajectory.orbit.model import SGP4

tle = TLE(...)
sgp4 = SGP4(tle)

orbit = Orbit.Builder().with_celestial(Earth.default()).with_model(sgp4).build()
```
Or even
```
from ostk.astrodynamics.trajectory import Orbit
from ostk.astrodynamics.trajectory.orbit.model.sgp4 import TLE

tle = TLE(...)
orbit = Orbit.Builder().with_earth().with_tle(tle).build()
```
and

```
from ostk.astrodynamics.trajectory import Orbit

states = [...]
orbit = Orbit.Builder().with_earth().with_tabulated(states).build()
```

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
@@ -82,36 +82,38 @@ class Orbit : public Trajectory
               // momentum)
     };
 
+    class Builder
+    {
+       public:
+        Builder();
+
+        ~Builder();
+
+        Builder& withCelestial(const Shared<const Celestial>& aCelestialObjectSPtr);
+
+        Builder& withEarth();
+
+        Builder& withModel(const orbit::Model& aModel);
+
+        Builder& withTLE(const TLE& aTLE);
+
+        Builder& withTabulated(const Array<State>& aStateArray, const Integer& anInitialRevolutionNumber = 0);
+
+        Orbit build() const;
+
+       private:
+        orbit::Model* modelPtr_;
+
+        Shared<const Celestial> celestialObjectSPtr_;
+    };
+
     typedef Array<Pass>::ConstIterator ConstPassIterator;
 
     /// @brief Constructor.
     ///
     /// @param aModel An orbit model.
     /// @param aCelestialObjectSPtr A shared pointer to a central celestial body. Defaults to Earth.
-    Orbit(
-        const orbit::Model& aModel,
-        const Shared<const Celestial>& aCelestialObjectSPtr = std::make_shared<Celestial>(Earth::Default())
-    );
-
-    /// @brief Constructor.
-    ///
-    /// @param aStateArray An array of states.
-    /// @param anInitialRevolutionNumber An initial revolution number.
-    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body. Defaults to Earth.
-    Orbit(
-        const Array<State>& aStateArray,
-        const Integer& anInitialRevolutionNumber,
-        const Shared<const Celestial>& aCelestialObjectSPtr = std::make_shared<Celestial>(Earth::Default())
-    );
-
-    /// @brief Constructor.
-    ///
-    /// @param aTLE Two-Line Element (TLE).
-    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body. Defaults to Earth.
-    Orbit(
-        const TLE& aTLE,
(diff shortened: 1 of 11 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
```diff
@@ -83,18 +87,30 @@ class Orbit : public Trajectory
...
+        const Shared<const Celestial>& aCelestialObjectSPtr = std::make_shared<Celestial>(Earth::Default())
+    );
+
+    /// @brief Constructor.
+    ///
+    /// @param aTLE Two-Line Element (TLE).
+    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body. Defaults to Earth.
+    Orbit(
+        const TLE& aTLE,
+        const Shared<const Celestial>& aCelestialObjectSPtr = std::make_shared<Celestial>(Earth::Default())
```
apaletta3, line 113: I think it might make more sense to make this a factory method instead of another constructor up here, because its specific to `TLE` models, and duplicates the functionality offered by the first constructor
```
Orbit::FromTLE(
	const TLE& aTLE,
)
```
Since you'd add this convenience factory method above, do you also want to add one for a Keplerian model and a Propagated model? Cause if not I don't see the need to prioritize creating one for TLE over the others, since the construction of an orbit from a TLE is already quite simple

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
```diff
@@ -239,7 +245,20 @@ class COE
...
+
+    /// @brief Construct a COE from a State
+    ///
+    /// @param aState A state
+    /// @param aGravitationalParameter A gravitational parameter
+    /// @return COE
+    static COE FromState(
+        const State& aState,
+        const Derived& aGravitationalParameter = EarthGravitationalModel::EGM2008.gravitationalParameter_
+    );
```
apaletta3, line 261: This feels like a little bit of a specific ad-hoc solution here.
I think a more generalizable solution could be to delete the `typedef Pair<Position, Velocity> CartesianState;` and replace the `COE::Cartesian` factory method with this new `COE::FromState` factory method you added. That way coordinate subset and reference frame checking are actually undertaken instead of assumptions made when parsing the pair of position and velocity like happens [here](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/434/files#diff-e928c48a82575ab8a2b99e3fa061796625187c58c18542ac932cfde16a4ca02dR449-R450)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
vishwa2710: Agreed that we can make it a factory method. I would then remove the state based Orbit constructor as well and do a `fromTabulated`

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
vishwa2710: I think that constructors should accept the minimum required information to create the object, and then we can have named constructors that simplify interfaces. There's no reason we can't have both. Also it would be a pretty big breaking change :)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
```diff
@@ -78,25 +82,39 @@ class Orbit : public Trajectory
...
+        Shared<const Celestial> celestialObjectSPtr_;
+    };
+
     typedef Array<Pass>::ConstIterator ConstPassIterator;
 
     /// @brief Constructor.
     ///
     /// @param aModel An orbit model.
-    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body.
+    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body. Defaults to Earth.
```
vishwa2710, line 115:
```suggestion
    /// @param aCelestialObjectSPtr A shared pointer to a central celestial body.
```

PR #442, opened 2024-10-01
feat: update pystate with cd, surface_area and mass

vishwa2710:


bindings/python/test/trajectory/test_state.py
```diff
@@ -313,7 +313,89 @@ def test_from_dict_with_angular_velocity(self):
...
+    def test_from_dict_with_drag_coefficient(self):
+        data = {
+            "timestamp": datetime.now(timezone.utc).isoformat(),
+            "rx_eci": 7000.0,
+            "ry_eci": 0.0,
+            "rz_eci": 0.0,
+            "vx_eci": 0.0,
+            "vy_eci": 7.5,
+            "vz_eci": 0.0,
+            "drag_coefficient": 2.2,
```
apaletta3, line 326: How do you know that the drag coefficient actually got added though? Instead of just getting ignored?
Might be worth asserting the number of of coordinates is equal to 7 here?

apaletta3 (approved):
Left one nit, otherwise LGTM

bindings/python/test/trajectory/test_state.py
vishwa2710: Great point.

PR #457, opened 2024-10-31
feat: add update_target binding to EventCondition

vishwa2710:
## Summary by CodeRabbit

- **New Features**
	- Introduced `update_target` method in the `EventCondition` class for updating target values.
	- Added a new read-only property `value_offset` in the `Target` class.

- **Bug Fixes**
	- Updated the `value` property type from `Real` to `float` for better accuracy.

- **Tests**
	- Added a new test for the `update_target` functionality.
	- Modified evaluation logic in tests to reflect changes in target definitions.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
@@ -141,7 +141,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_EventCondition(pybind11::module& aMo
                 R"doc(
                     The value of the target.
 
-                    :type: Real
+                    :type: float
                 )doc"
             )
             .def_readonly(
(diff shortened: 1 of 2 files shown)


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
```diff
@@ -141,7 +141,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_EventCondition(pybind11::module& aMo
                 R"doc(
                     The value of the target.
 
-                    :type: Real
+                    :type: float
```
apaletta3, line 144: Are you sure it gets returned as a float? Don't we have to cast it in python to make sure its a float?

apaletta3 (approved):
LGTM pending one comment

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/EventCondition.cpp
vishwa2710: It's technically a Real you're right. But I think it's clearer in the docs to say float as they can be used almost interchangably (except during JSON serialization) and that way users don't have to check what a `Real` is.

PR #473, opened 2024-12-13
feat: frozen orbit conditions

alex-liang3:


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -641,6 +641,130 @@ COE COE::FromSIVector(const Vector6d& aCOEVector, const AnomalyType& anAnomalyTy
...
+    const Angle critical_aops[] = {Angle::Degrees(90.0), Angle::Degrees(270.0)};
+
+    const Real re = anEquatorialRadius.inMeters();
+    const Real j2 = aJ2;
+    const Real j3 = aJ3;
+
+    // ecc =~ ecc_coefficient * sin(incl)
+    const Real ecc_coefficient = -j3 * re / 2.0 / j2 / aSemiMajorAxis.inMeters();
+
+    const auto eccentricityFromInclination = [&](const Angle& inclination) -> Real
```
vishwa2710, line 693:
```suggestion
    const auto eccentricityFromInclination = [&eccCoefficient](const Angle& inclination) -> Real
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -641,6 +641,130 @@ COE COE::FromSIVector(const Vector6d& aCOEVector, const AnomalyType& anAnomalyTy
...
+
+    const Angle critical_inclinations[] = {Angle::Degrees(63.4349), Angle::Degrees(116.5651)};
+    const Angle critical_aops[] = {Angle::Degrees(90.0), Angle::Degrees(270.0)};
+
+    const Real re = anEquatorialRadius.inMeters();
+    const Real j2 = aJ2;
+    const Real j3 = aJ3;
+
+    // ecc =~ ecc_coefficient * sin(incl)
+    const Real ecc_coefficient = -j3 * re / 2.0 / j2 / aSemiMajorAxis.inMeters();
```
vishwa2710, line 691:
```suggestion
    const Real eccCoefficient = -j3 * equatorialRadius_meters / 2.0 / j2 / aSemiMajorAxis.inMeters();
```
We camel case in the C++ code

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
```diff
@@ -252,6 +252,30 @@ class COE
     /// @return COE
     static COE FromSIVector(const Vector6d& aCOEVector, const AnomalyType& anAnomalyType);
 
+    /// @brief Construct a frozen orbit from an incomplete set of COEs
```
vishwa2710, line 255: Worth updating the docstrings here to specify what the various combinations of legal inputs are

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -387,6 +387,37 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Orbit_Model_Kepler_COE(py
             arg("anomaly_type")
         )
 
+        .def_static(
+            "frozen_orbit",
+            &COE::FrozenOrbit,
+            R"doc(
+                Build a `COE` model of a frozen orbit.
```
vishwa2710, line 394: Worth updating the docstrings here to specify what the various combinations of legal inputs are:
- just sma
- sma + inclination
- sma + AoP
- sma + inclination + aop (but some have to be critical angles)
etc.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -387,6 +387,37 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Orbit_Model_Kepler_COE(py
             arg("anomaly_type")
         )
 
+        .def_static(
```
vishwa2710, line 390: might be nice to also provide an overloaded constructor here (and the cpp) which accepts a Celestial from which you can get the gravitationalModel and access these parameters.

vishwa2710 (commented):
Excellent stuff! Left a few comments, mostly around documentation and style.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.hpp
```diff
@@ -250,6 +250,30 @@ class Orbit : public Trajectory
         const Angle& anArgumentOfLatitude = Angle::Zero()
     );
 
+    /// @brief Constructs a Frozen orbit
```
vishwa2710, line 253: Just like the COE one can you update the docstring here enumerating the various cases 🙏

vishwa2710 (approved):
1 nit comment. That's excellent!

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit.cpp
```diff
@@ -375,6 +375,54 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Orbit(pybind11::module& a
                 )doc"
             )
 
+            .def_static(
+                "frozen",
+                &Orbit::Frozen,
+                arg("epoch"),
+                arg("altitude"),
```
apaletta3, line 382: Shouldn't we provide SMA instead of altitude as the argument here since that is more standard?
Is it geodetic altitude, geocentric altitude, etc? Or at least specify that it is geocentric altitude in this case

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -641,6 +642,159 @@ COE COE::FromSIVector(const Vector6d& aCOEVector, const AnomalyType& anAnomalyTy
...
+
+    const auto inclinationFromEccentricity = [&eccCoefficient](const Real& eccentricity) -> Angle
+    {
+        return Angle::Radians(asin(eccentricity / eccCoefficient));
+    };
+
+    // TBI: this only works because neither set of critical angles are 0/360 degrees
+    const auto isCritical = [](const Angle& angle, const Array<Angle>& criticalAngles) -> bool
+    {
+        const Real epsilon = Angle::Arcseconds(1.0).inRadians();  // TBI: make configurable?
```
apaletta3, line 733: Indeed I do think the "criticality" angle tolerance should be configurable as an extra argument - preferably on the sooner side, because in reality if you area within a certain tolerance (I think much bigger than 1 Arcsec) of the conditions, your AOP and ecc will oscillate around the condition, instead of directly diverging - so it may actually be counterproductive to try and always correct and send us back to the perfect condition

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -641,6 +642,159 @@ COE COE::FromSIVector(const Vector6d& aCOEVector, const AnomalyType& anAnomalyTy
...
+            aop,
+            aTrueAnomaly,
+        };
+    }
+
+    // If AoP is (given) and not a critical angle, then the inclination must be critical
+    const Angle inclination = anInclination.isDefined() ? anInclination : criticalInclinations[0];
+    if (!isCritical(inclination, criticalInclinations))
+    {
+        throw ostk::core::error::RuntimeError("Provided inclination must be a critical value");
```
apaletta3, line 783: so if a user intended to provide a critical inclination and AoP - but was slightly off, they'll only get an error message saying "Provided inclination must be a critical value" - even though it wasn't necessarily just the inclination that was incorrectly non-critical?
Is that a correct interpretation of the code? If so I propose shifting around the logic / error message to yield a more accurate portrayal of what the user did incorrectly

apaletta3 (commented):
Great work man!

PR #484, opened 2025-01-05
refactor!: rename nadir_pointing profile to local_orbital_frame

vishwa2710:
## Release Notes

- **Refactor**
    - Renamed `NadirPointing` method to `LocalOrbitalFramePointing` across multiple components of the OpenSpaceToolkit Astrodynamics library.
    - Updated method signatures and documentation to reflect the new method name while maintaining existing functionality.

- **Tests**
    - Updated test suites to utilize the new `LocalOrbitalFramePointing` method.
    - Ensured consistent test coverage with the renamed method and adjusted expected outcomes accordingly.


lucas-bremond (changes requested):
I think the docstrings should be updated accordingly?

Also, we could make the PR non-breaking keeping "NadirPointing" with a warning being printed on use?

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -292,7 +292,7 @@ class Profile
     /// @param anOrbit An orbit
     /// @param anOrbitalFrameType An orbital frame type
     /// @return Flight profile
-    static Profile NadirPointing(
+    static Profile LocalOrbitalFrame(
         const trajectory::Orbit& anOrbit, const trajectory::Orbit::FrameType& anOrbitalFrameType
```
apaletta3, line 296: For clarity could you also rename this in Orbit to `LocalFrameType`?

vishwa2710:
> I think the docstrings should be updated accordingly?
> 
> Also, we could make the PR non-breaking keeping "NadirPointing" with a warning being printed on use?

Yeah we could provide a warning in this MR and then deprecate it in an upcoming patch. That's best. Will do!

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile/Model/Transform.cpp
```diff
@@ -127,8 +127,8 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile_Model_Transform(pybin
         )
 
         .def_static(
-            "nadir_pointing",
-            &Transform::NadirPointing,
+            "local_orbital_frame",
```
apaletta3, line 130: Perhaps rename to `local_orbital_frame_pointing` for consistency?
Also update the docstrings accordingly and apply to the rest of the files

apaletta3 (changes requested):
Approved pending requested changes

vishwa2710:
> > I think the docstrings should be updated accordingly?
> > Also, we could make the PR non-breaking keeping "NadirPointing" with a warning being printed on use?
> 
> Yeah we could provide a warning in this MR and then deprecate it in an upcoming patch. That's best. Will do!

Actually, upon further thought, since there are other breaking changes in this upcoming release, I would like to just club this together to simplify propagating the changes across downstream dependencies.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
vishwa2710: Rename the enum? I can do that in a separate MR, that will be a pretty big change to apply across repos...

vishwa2710: Will do this in a separate MR.

lucas-bremond (approved):

PR #493, opened 2025-01-13
refactor!: use state instead of instantposvel

### LINKED ISSUE #433: [feat] Prefer reference frame dependent information (like position and velocity vectors) to be encapsulated in State object for LocalOrbitalFrames and Profiles
For the `LocalOrbitalFrameProvider` for example, we should move away from {Instant, VectorXd, VectorXd} (basically, time, pos and vel), to simply use {State}  in the signature. Then we can ensure that pos and vel are extracted in the parent frame (which is something we don't do now).

vishwa2710:
Closes #433

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `State` object that consolidates instant, position, and velocity parameters across multiple components.

- **Refactor**
  - Streamlined method signatures in trajectory and frame-related classes by replacing multiple parameters with a single `State` object.
  - Updated method calls and test cases to utilize the new `State` object.

- **Documentation**
  - Updated method documentation to reflect changes in parameter types and method signatures.

- **Tests**
  - Modified test cases to accommodate the new `State` object.
  - Removed skipped test methods, indicating readiness for new implementations.

### DIFF
--- a/bindings/python/test/trajectory/test_local_orbital_frame_factory.py
+++ b/bindings/python/test/trajectory/test_local_orbital_frame_factory.py
@@ -7,7 +7,10 @@ from ostk.physics.time import DateTime
 from ostk.physics.time import Scale
 from ostk.physics.coordinate import Frame
 from ostk.physics.coordinate import Transform
+from ostk.physics.coordinate import Position
+from ostk.physics.coordinate import Velocity
 
+from ostk.astrodynamics.trajectory import State
 from ostk.astrodynamics.trajectory import LocalOrbitalFrameFactory
 from ostk.astrodynamics.trajectory import LocalOrbitalFrameTransformProvider
 
@@ -24,7 +27,7 @@ def local_orbital_transform_provider_type() -> LocalOrbitalFrameTransformProvide
 
 @pytest.fixture
 def transform_generator() -> callable:
-    return lambda instant, position, velocity: Transform.identity(Transform.Type.passive)
+    return lambda state: Transform.identity(Transform.Type.passive)
 
 
 @pytest.fixture
@@ -38,13 +41,24 @@ def instant() -> Instant:
 
 
 @pytest.fixture
-def position_vector() -> list:
-    return [7500000.0, 0.0, 0.0]
+def position() -> list:
+    return Position.meters([7500000.0, 0.0, 0.0], Frame.GCRF())
 
 
 @pytest.fixture
-def velocity_vector() -> list:
-    return [0.0, 5335.865450622126, 5335.865450622126]
+def velocity() -> list:
+    return Velocity.meters_per_second(
+        [0.0, 5335.865450622126, 5335.865450622126], Frame.GCRF()
+    )
+
+
+@pytest.fixture
+def state(
+    instant: Instant,
+    position: Position,
+    velocity: Velocity,
+) -> State:
+    return State(instant, position, velocity)
 
 
 class TestLocalOrbitalFrameFactory:
@@ -66,15 +80,9 @@ class TestLocalOrbitalFrameFactory:
     def test_generate_frame(
         self,
         local_orbital_frame_factory: LocalOrbitalFrameFactory,
-        instant: Instant,
-        position_vector: list,
-        velocity_vector: list,
+        state: State,
     ):
-        frame = local_orbital_frame_factory.generate_frame(
-            instant,
-            position_vector,
-            velocity_vector,
-        )
+        frame = local_orbital_frame_factory.generate_frame(state=state)
 
         assert frame is not None
         assert frame.is_defined()
--- a/bindings/python/test/trajectory/test_local_orbital_frame_factory.py
+++ b/bindings/python/test/trajectory/test_local_orbital_frame_factory.py
@@ -41,12 +41,12 @@ def instant() -> Instant:
 
 
 @pytest.fixture
-def position() -> list:
+def position() -> Position:
     return Position.meters([7500000.0, 0.0, 0.0], Frame.GCRF())
 
 
 @pytest.fixture
-def velocity() -> list:
+def velocity() -> Velocity:
     return Velocity.meters_per_second(
         [0.0, 5335.865450622126, 5335.865450622126], Frame.GCRF()
     )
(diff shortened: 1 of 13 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameFactory.cpp
```diff
@@ -50,18 +50,18 @@ struct SharedFrameEnabler : public Frame
     }
 };
 
-Shared<const Frame> LocalOrbitalFrameFactory::generateFrame(
-    const Instant& anInstant, const Vector3d& aPosition, const Vector3d& aVelocity
-) const
+Shared<const Frame> LocalOrbitalFrameFactory::generateFrame(const State& aState) const
 {
-    const String name = this->generateFrameName(anInstant, aPosition, aVelocity);
+    const State stateInParentFrame = aState.inFrame(parentFrameSPtr_);
```
phc1990, line 55: Nit: since we only need pos/vel and not all coordinate subsets might be portable to another frame without raising an error, we could reduce it first to avoid any issues

```suggestion
    const StateBuilder posVelStateBuilder = StateBuilder(state.getFrame(), {CartesianPosition::Default(), CartesianVelocity::Default()})
    const State posVelStateInParentFrame = posVelStateBuilder.reduce(aState).inFrame(parentFrameSPtr_);
```

phc1990 (approved):
Nice, I think this improves a lot the interface. 
Left a minor suggestion.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
```diff
@@ -167,18 +167,18 @@ Transform LocalOrbitalFrameTransformProvider::generateTransform(
         case LocalOrbitalFrameTransformProvider::Type::VVLH:
         {
             // Y axis along negative orbital momentum
-            // Z axis along negative position vector
-            const Vector3d transformPosition = -aPosition;
-            const Vector3d transformVelocity = -aVelocity;
-            const Vector3d zAxis = -aPosition.normalized();
-            const Vector3d yAxis = -aPosition.cross(aVelocity).normalized();
+            // X axis toward velocity vector
```
phc1990, line 170:
```suggestion
            // Z axis along negative position vector
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
```diff
@@ -145,17 +144,18 @@ Transform LocalOrbitalFrameTransformProvider::generateTransform(
         {
             // X axis along position vector
             // Z axis along orbital momentum
-            const Vector3d transformPosition = -aPosition;
-            const Vector3d transformVelocity = -aVelocity;
-            const Vector3d xAxis = aPosition.normalized();
-            const Vector3d zAxis = aPosition.cross(aVelocity).normalized();
+            // Y axis toward velocity vector
```
phc1990, line 147: Don't know if this was intended to clarify.

I'm happy with it but maybe we should also add the a 3rd axis "toward" comment in the other frames?

Don't know if people will understand the difference between "along" and "toward". Maybe we should do "align" vs "toward"?

I leave it up to you.

bindings/python/test/trajectory/test_local_orbital_frame_factory.py
```diff
@@ -38,13 +41,24 @@ def instant() -> Instant:
 
 
 @pytest.fixture
-def position_vector() -> list:
-    return [7500000.0, 0.0, 0.0]
+def position() -> list:
```
apaletta3, line 44:
```suggestion
def position() -> Position:
```

bindings/python/test/trajectory/test_local_orbital_frame_factory.py
```diff
@@ -38,13 +41,24 @@ def instant() -> Instant:
...
+    return Position.meters([7500000.0, 0.0, 0.0], Frame.GCRF())
 
 
 @pytest.fixture
-def velocity_vector() -> list:
-    return [0.0, 5335.865450622126, 5335.865450622126]
+def velocity() -> list:
```
apaletta3, line 49:
```suggestion
def velocity() -> Velocity:
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
```diff
@@ -119,18 +116,20 @@ String LocalOrbitalFrameTransformProvider::StringFromType(const LocalOrbitalFram
...
-    const Instant& anInstant,
-    const Vector3d& aPosition,
-    const Vector3d& aVelocity
+Transform LocalOrbitalFrameTransformProvider::GenerateTransform(
+    const LocalOrbitalFrameTransformProvider::Type& aType, const State& aState
 )
 {
+    const Instant& instant = aState.accessInstant();
+    const Vector3d positionCoordinates = aState.getPosition().accessCoordinates();
+    const Vector3d velocityCoordinates = aState.getVelocity().accessCoordinates();
```
apaletta3, line 125: These are sure to be given in GCRF right?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
apaletta3: Agree with Pau, either add the 3rd axis comment for all cases or no cases

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
apaletta3: Indeed, and then you can add the third axis at the end

apaletta3 (approved):
Very nice! Approved pending a couple nits

src/OpenSpaceToolkit/Astrodynamics/Trajectory/LocalOrbitalFrameTransformProvider.cpp
vishwa2710: They're not, they will be in whatever parent frame the user has provided, which can be GCRF. Typically these are provided by the localFrameFactory and in the correct frame. So I think we're covered.

apaletta3, line 125: I think we should explicitly put them in GCRF just in case, no? While we work with raw coordinates below that are assumed to be in GCRF
```suggestion
    const State& stateInGCRF = aState.inFrame(Frame::GCRF())
    const Vector3d positionCoordinates = stateInGCRF.getPosition().accessCoordinates();
    const Vector3d velocityCoordinates = stateInGCRF.getVelocity().accessCoordinates();
```

PR #495, opened 2025-01-14
fix: fix off nadir state missing frame transformation

phc1990:
**Context:**

Off Nadir data provider assumes that the State frame is given in the VVLH parent frame (GCRF). Whilst this is likely the case most of the time, it cannot be guaranteed. This results in a "wrong" Local Orbital Frame being built if the State's frame is not GCRF.

**This PR:**

This PR ensures that the State is previously transformed to the VVLH parent frame, before its coordinates are extracted in order to construct the Local Orbital Frame.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the logic for off-nadir angle calculations by improving state transformation and frame utilization.


apaletta3 (approved):

vishwa2710:
Closing this MR as https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/493 has been landed which takes care of this.

PR #500, opened 2025-01-17
feat: Least Squares Solver

vishwa2710:
## Release Notes

- **New Features**
  - Added Least Squares Solver to the Open Space Toolkit Astrodynamics module.
  - Introduced Python bindings for the Least Squares Solver, allowing interaction with its components.
  - Implemented comprehensive solver functionality for non-linear least squares problems, including step and analysis structures.

- **Improvements**
  - Enhanced solver with iteration tracking, RMS error calculation, and covariance estimation.
  - Added support for coordinate subset handling in solver computations.
  - Implemented robust error handling and input validation.

- **Bug Fixes**
  - Removed unsupported addition and subtraction methods from the CartesianAcceleration, CartesianPosition, and CartesianVelocity classes.

- **Testing**
  - Comprehensive unit tests added for Least Squares Solver, validating functionality and performance across various scenarios.
  - Enhanced test coverage for the `Step` and `Analysis` components of the solver.
  - Updated tests to validate the correctness of the `add` and `subtract` methods in the `CoordinateSubset` class.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
@@ -91,30 +91,44 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 const State&,
                 const MatrixXd&,
                 const MatrixXd&,
-                const MatrixXd&,
+                const Array<State>&,
                 const Array<LeastSquaresSolver::Step>&>(),
             R"doc(
                 Constructor.
 
                 Args:
-                    observation_count (int): The observation count.
+                    rms_error (float): The RMS error.
                     termination_criteria (str): The termination criteria.
-                    solution_state (State): The solution state.
-                    solution_covariance (np.ndarray): The solution covariance matrix.
-                    solution_frisbee_covariance (np.ndarray): The solution Frisbee covariance matrix.
-                    solution_residuals (np.ndarray): The solution residuals.
+                    estimate (State): The estimated state.
+                    estimated_covariance (np.ndarray): The estimated covariance matrix.
+                    estimated_frisbee_covariance (np.ndarray): The estimated Frisbee covariance matrix.
+                    computed_observations (list[State]): The computed observations of the final iteration.
                     steps (list[LeastSquaresSolver.Step]): The steps.
             )doc",
             arg("observation_count"),
             arg("termination_criteria"),
-            arg("solution_state"),
-            arg("solution_covariance"),
-            arg("solution_frisbee_covariance"),
-            arg("solution_residuals"),
+            arg("estimate"),
+            arg("estimated_covariance"),
+            arg("estimated_frisbee_covariance"),
+            arg("computed_observations"),
             arg("steps")
         )
         .def("__str__", &(shiftToString<LeastSquaresSolver::Analysis>))
         .def("__repr__", &(shiftToString<LeastSquaresSolver::Analysis>))
+        .def(
+            "compute_residual_states",
+            &LeastSquaresSolver::Analysis::computeResidualStates,
+            R"doc(
+            Compute the residual states.
+
+            Args:
+                observations (list[State]): The observations.
+
+            Returns:
+                list[State]: The residuals.
+        )doc",
+            arg("observations")
+        )
         .def_readonly(
             "rms_error",
             &LeastSquaresSolver::Analysis::rmsError,
@@ -152,37 +166,37 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
             )doc"
         )
         .def_readonly(
-            "solution_state",
-            &LeastSquaresSolver::Analysis::solutionState,
+            "estimate",
+            &LeastSquaresSolver::Analysis::estimatedState,
             R"doc(
-                The solution state.
+                The estimated state.
 
                 :type: State
             )doc"
         )
         .def_readonly(
-            "solution_covariance",
-            &LeastSquaresSolver::Analysis::solutionCovariance,
+            "estimated_covariance",
+            &LeastSquaresSolver::Analysis::estimatedCovariance,
             R"doc(
-                The solution covariance matrix.
+                The estimated covariance matrix.
 
                 :type: np.ndarray
             )doc"
         )
         .def_readonly(
-            "solution_frisbee_covariance",
-            &LeastSquaresSolver::Analysis::solutionFrisbeeCovariance,
+            "estimated_frisbee_covariance",
+            &LeastSquaresSolver::Analysis::estimatedFrisbeeCovariance,
             R"doc(
-                The solution Frisbee covariance matrix.
+                The estimated Frisbee covariance matrix.
 
                 :type: np.ndarray
             )doc"
         )
         .def_readonly(
-            "solution_residuals",
-            &LeastSquaresSolver::Analysis::solutionResiduals,
+            "computed_observations",
+            &LeastSquaresSolver::Analysis::computedObservationStates,
             R"doc(
-                The solution residuals.
+                The computed observations of the final iteration.
 
                 :type: np.ndarray
             )doc"
@@ -251,19 +265,19 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 Solve the non-linear least squares problem.
 
                 Args:
-                    initial_guess (State): Initial guess state.
+                    initial_guess (State): Initial guess state (the Estimated State is of the same domain as this state).
                     observations (list[State]): List of observations.
                     state_generator (callable[list[State],[State, list[Instant]]]): Function to generate states.
-                    estimation_sigmas (dict[CoordinateSubset, np.ndarray], optional): Dictionary of sigmas for initial guess.
+                    initial_guess_sigmas (dict[CoordinateSubset, np.ndarray], optional): Dictionary of sigmas for initial guess.
                     observation_sigmas (dict[CoordinateSubset, np.ndarray], optional): Dictionary of sigmas for observations.
 
                 Returns:
-                    LeastSquaresSolver::Analysis: The analysis of the solution.
+                    LeastSquaresSolver::Analysis: The analysis of the estimate.
             )doc",
             arg("initial_guess"),
             arg("observations"),
             arg("state_generator"),
-            arg_v("estimation_sigmas", DEFAULT_ESTIMATION_SIGMAS, "{}"),
+            arg_v("initial_guess_sigmas", DEFAULT_INITIAL_GUESS_SIGMAS, "{}"),
             arg_v("observation_sigmas", DEFAULT_OBSERVATION_SIGMAS, "{}")
         )
         .def_static(
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
@@ -251,20 +251,20 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 Solve the non-linear least squares problem.
 
                 Args:
-                    initial_guess_state (State): Initial guess state.
+                    initial_guess (State): Initial guess state.
                     observations (list[State]): List of observations.
                     state_generator (callable[list[State],[State, list[Instant]]]): Function to generate states.
-                    initial_guess_sigmas (dict, optional): Dictionary of sigmas for initial guess.
-                    observation_sigmas (dict, optional): Dictionary of sigmas for observations.
+                    estimation_sigmas (dict[CoordinateSubset, np.ndarray], optional): Dictionary of sigmas for initial guess.
+                    observation_sigmas (dict[CoordinateSubset, np.ndarray], optional): Dictionary of sigmas for observations.
 
                 Returns:
                     LeastSquaresSolver::Analysis: The analysis of the solution.
             )doc",
-            arg("initial_guess_state"),
+            arg("initial_guess"),
             arg("observations"),
             arg("state_generator"),
-            arg_v("initial_guess_sigmas", DEFAULT_INITIAL_GUESS_SIGMAS, "{}")
-            arg_v("observation_sigmas"), DEFAULT_OBSERVATION_SIGMAS, "{}")
+            arg_v("estimation_sigmas", DEFAULT_ESTIMATION_SIGMAS, "{}"),
+            arg_v("observation_sigmas", DEFAULT_OBSERVATION_SIGMAS, "{}")
         )
         .def_static(
             "calculate_empirical_covariance",
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
@@ -86,7 +86,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
     )
         .def(
             init<
-                const Size&,
+                const Real&,
                 const String&,
                 const State&,
                 const MatrixXd&,
@@ -105,7 +105,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                     computed_observations (list[State]): The computed observations of the final iteration.
                     steps (list[LeastSquaresSolver.Step]): The steps.
             )doc",
-            arg("observation_count"),
+            arg("rms_error"),
             arg("termination_criteria"),
             arg("estimate"),
             arg("estimated_covariance"),
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
@@ -3,7 +3,6 @@
 
 #include <pybind11/eigen.h>
 #include <pybind11/pybind11.h>
-#include <pybind11/stl.h>
 
 #include <OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.hpp>
 
@@ -49,11 +48,11 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 Constructor.
 
                 Args:
-                    aRmsError (float): The RMS error.
-                    anXHat (np.ndarray): The X hat vector.
+                    rms_error (float): The RMS error.
+                    x_hat (np.ndarray): The X hat vector.
             )doc",
-            arg("aRmsError"),
-            arg("anXHat")
+            arg("rms_error"),
+            arg("x_hat")
         )
         .def("__str__", &(shiftToString<LeastSquaresSolver::Step>))
         .def("__repr__", &(shiftToString<LeastSquaresSolver::Step>))
@@ -87,10 +86,10 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
     )
         .def(
             init<
-                const Real&,
                 const Size&,
                 const String&,
                 const State&,
+                const MatrixXd&,
                 const MatrixXd&,
                 const MatrixXd&,
                 const Array<LeastSquaresSolver::Step>&>(),
@@ -98,20 +97,20 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 Constructor.
 
                 Args:
-                    rms_error (float): The RMS error.
-                    iteration_count (int): The iteration count.
+                    observation_count (int): The observation count.
                     termination_criteria (str): The termination criteria.
                     solution_state (State): The solution state.
                     solution_covariance (np.ndarray): The solution covariance matrix.
                     solution_frisbee_covariance (np.ndarray): The solution Frisbee covariance matrix.
+                    solution_residuals (np.ndarray): The solution residuals.
                     steps (list[LeastSquaresSolver.Step]): The steps.
             )doc",
-            arg("rms_error"),
-            arg("iteration_count"),
+            arg("observation_count"),
             arg("termination_criteria"),
             arg("solution_state"),
             arg("solution_covariance"),
             arg("solution_frisbee_covariance"),
+            arg("solution_residuals"),
             arg("steps")
         )
         .def("__str__", &(shiftToString<LeastSquaresSolver::Analysis>))
@@ -123,6 +122,15 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
                 The RMS error.
 
                 :type: float
+            )doc"
+        )
+        .def_readonly(
+            "observation_count",
+            &LeastSquaresSolver::Analysis::observationCount,
+            R"doc(
+                The observation count.
+
+                :type: int
             )doc"
         )
         .def_readonly(
@@ -166,6 +174,15 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
             &LeastSquaresSolver::Analysis::solutionFrisbeeCovariance,
             R"doc(
                 The solution Frisbee covariance matrix.
+
+                :type: np.ndarray
+            )doc"
+        )
+        .def_readonly(
+            "solution_residuals",
+            &LeastSquaresSolver::Analysis::solutionResiduals,
+            R"doc(
+                The solution residuals.
 
                 :type: np.ndarray
             )doc"
@@ -235,19 +252,19 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
 
                 Args:
                     initial_guess_state (State): Initial guess state.
-                    reference_states (list[State]): List of reference states.
-                    generate_states_callback (callable[list[State],[State, list[Instant]]]): Function to generate states.
+                    observations (list[State]): List of observations.
+                    state_generator (callable[list[State],[State, list[Instant]]]): Function to generate states.
                     initial_guess_sigmas (dict, optional): Dictionary of sigmas for initial guess.
-                    reference_state_sigmas (dict, optional): Dictionary of sigmas for reference states.
+                    observation_sigmas (dict, optional): Dictionary of sigmas for observations.
 
                 Returns:
                     LeastSquaresSolver::Analysis: The analysis of the solution.
             )doc",
             arg("initial_guess_state"),
-            arg("reference_states"),
-            arg("generate_states_callback"),
+            arg("observations"),
+            arg("state_generator"),
             arg("initial_guess_sigmas") = DEFAULT_INITIAL_GUESS_SIGMAS,
-            arg("reference_state_sigmas") = DEFAULT_REFERENCE_STATE_SIGMAS
+            arg("observation_sigmas") = DEFAULT_OBSERVATION_SIGMAS
         )
         .def_static(
             "calculate_empirical_covariance",
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
@@ -35,9 +35,13 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Solver_LeastSquaresSolver(py::module
         )doc"
     );
 
-    class_<LeastSquaresSolver::Step>(leastSquaresSolver, "Step", R"doc(
-        Class representing a step in the least squares solver.
-    )doc")
+    class_<LeastSquaresSolver::Step>(
+        leastSquaresSolver,
+        "Step",
+        R"doc(
+            Class representing a step in the least squares solver.
+        )doc"
+    )
         .def(
             init<const Real&, const VectorXd&>(),
             R"doc(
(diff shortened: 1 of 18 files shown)


src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,419 @@
...
+    if (aReferenceStateArray.isEmpty())
+    {
+        throw ostk::core::error::runtime::Undefined("Reference states");
+    }
+
+    // Setup state builders
+    const StateBuilder estimationStateBuilder(anInitialGuessState);
+    const Shared<const Frame>& estimationStateFrame = anInitialGuessState.accessFrame();
+    const StateBuilder referenceStateBuilder(aReferenceStateArray[0]);
+    const Array<Shared<const CoordinateSubset>> referenceStateSubsets = referenceStateBuilder.getCoordinateSubsets();
```
phc1990, line 185: Since we are not supporting "heterogeneous" observables (nothing wrong with it, not in scope) I we should probably run a quick check looping over the reference states and ensuring they all have the same coordinate subset as the first one, raising an error otherwise.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    // Initialize state vectors
+    VectorXd XNom = anInitialGuessState.accessCoordinates();
+    VectorXd xHat = VectorXd::Zero(stateDimension);
+
+    // Get reference state coordinates
+    MatrixXd referenceStatesCoordinates(referenceStateCount, referenceStateDimension);
+    for (Size i = 0; i < referenceStateCount; ++i)
+    {
+        referenceStatesCoordinates.row(i) = aReferenceStateArray[i].inFrame(estimationStateFrame).accessCoordinates();
```
phc1990, line 234: I know it's the same, but just to be in case of future refactors:

```suggestion
        referenceStatesCoordinates.row(i) = aReferenceStateArray[i].inFrame(estimationStateFrame).extractCoordinates(referenceStateSubsets);
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    return (coordinates.transpose() * coordinates) / count;
+}
+
+LeastSquaresSolver LeastSquaresSolver::Default()
+{
+    return LeastSquaresSolver(20, 1.0);
+}
+
+VectorXd LeastSquaresSolver::expandSigmas(
```
phc1990, line 373: Don't know if this "expands" or "aligns" the coordinates with the builder.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,419 @@
...
+            throw ostk::core::error::RuntimeError("Sigma value must be greater than 0.");
+        }
+
+        const VectorXd scaledSigma = sigmaIt->second / std::sqrt(subsetSize);
+        sigmasCoordinates.segment(currentIndex, subsetSize) = scaledSigma;
+        currentIndex += subsetSize;
+    }
+
+    return sigmasCoordinates;
+}
```
phc1990, line 415: Maybe building a state from the sigma map might help you ensure that you're aligned coordinates wise:

⚠️ pseudocode
 
```
sigmaSubsets
sigmaCoordinates
for subset, coordinates in aSigmas:
    allSubsets.add(subset)
    allCoordinates.add(coordinates)
 
 State sigmaState = State(Instant::J2000(), allSubsets, aStateBuilder.getFrame(), allCoordinates);
 Vector alignedCoordinates = aStateBuilder.reduce(sigmaState) // This should ensure presence of subsets and reorder them. It does not raise an error however it there are more subsets in sigma than in the builder.
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    if (aReferenceStateArray.isEmpty())
+    {
+        throw ostk::core::error::runtime::Undefined("Reference states");
+    }
+
+    // Setup state builders
+    const StateBuilder estimationStateBuilder(anInitialGuessState);
+    const Shared<const Frame>& estimationStateFrame = anInitialGuessState.accessFrame();
```
phc1990, line 183: Estimation State Builder would use the estimationCoodinates, and we would nevertheless want a initialguessStateBuilder

```suggestion
    const Shared<const Frame>& estimationStateFrame = anInitialGuessState.accessFrame();
    const StateBuilder estimationStateBuilder(estimationStateFrame, estimationCoordinates);
    const StateBuilder initialGuessStateBuilder(anInnitialGuessState);
```

phc1990 (commented):
Amazing to have this in OSTk, left a few comments
 🙌

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
phc1990: Not for now, but food for thought: if we see room for the state logic to help us here (I think there is), then we could consider one or more of the following:
- have this function already return APrioriInv (unless "expanded sigmas" are actually used somewhere else)
- support partial sigmas (e.g. estimation is {pos, vel}, but we only know {pos} sigmas) -> don't know how feasible it is.

(again, just for the future)

vishwa2710: Let's discuss this in person, I'm a bit confused.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
vishwa2710: Hmmmm this feels like overkill as the state builder and coordinate subsets are created from the reference state array...

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+    if (aSigmas.size() != stateSubsets.getSize())
+    {
+        throw ostk::core::error::RuntimeError("Sigma count does not match State Coordinate Subset count.");
+    }
+
+    Size totalSize = 0;
+    for (const auto& subset : stateSubsets)
+    {
+        totalSize += subset->getSize();
+    }
```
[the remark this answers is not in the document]
apaletta3: https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/504

apaletta3 (commented):
Great stuff! Left quite a few comments to address 🏓

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        referenceStatesCoordinates.row(i) = aReferenceStateArray[i].inFrame(estimationStateFrame).accessCoordinates();
+    }
+
+    // Initialize arrays for the iteration
+    MatrixXd residuals = MatrixXd::Zero(referenceStateCount, referenceStateDimension);
+    Real previousRmsError = std::numeric_limits<double>::max();
+    String terminationCriteria = "Maximum Iteration Threshold";
+    Array<Step> steps;
+
+    MatrixXd Lambda;
```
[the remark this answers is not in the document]
vishwa2710: Doesn't matter as it will be dynamically allocated regardless as it's a MatrixD but done for cleanliness

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        // Generate states at reference instants
+        Array<Instant> referenceInstants;
+        for (const auto& state : aReferenceStateArray)
+        {
+            referenceInstants.add(state.getInstant());
+        }
+
+        // Build current state for propagation
+        const State currentState = estimationStateBuilder.build(anInitialGuessState.getInstant(), XNom);
+        const Array<State> computedStates = aGenerateStatesCallback(currentState, referenceInstants);
```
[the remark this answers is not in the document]
vishwa2710: These are not reference instants, these are computed states at the reference instants. Renamed to estimatedStates.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        // Convert computed states to matrix form
+        MatrixXd computedStatesCoordinates(referenceStateCount, referenceStateDimension);
+        for (Size i = 0; i < referenceStateCount; ++i)
+        {
+            computedStatesCoordinates.row(i) =
+                computedStates[i].inFrame(estimationStateFrame).extractCoordinates(referenceStateSubsets);
+        }
+
+        // Compute Jacobian matrix using finite differences
+        const Array<MatrixXd> HFull = finiteDifferenceSolver.computeStateTransitionMatrix(
```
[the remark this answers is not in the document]
vishwa2710: This has to be re-computed each time we update the state, so it cannot be done before the main loop.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+            // Extract current H matrix
+            const MatrixXd H = HFull[i];
+
+            // Accumulate normal equations
+            Lambda += H.transpose() * RInv * H;
+            N += H.transpose() * RInv * y;
+
+            // Accumulate Frisbee covariance
+            frisbeeCovariance += H.transpose() * RInv * y * y.transpose() * RInv * H;
```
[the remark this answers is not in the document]
vishwa2710: I think for now let's leave it and we can do an assessment for performance later, easy to remove the code later, we can always just leave it as null if we don't want to compute it.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        {
+            terminationCriteria = "RMS Update Threshold";
+            break;
+        }
+
+        previousRmsError = currentRmsError;
+    }
+
+    // Compute final covariance matrices
+    const MatrixXd PHat = Lambda.ldlt().solve(MatrixXd::Identity(Lambda.rows(), Lambda.cols()));
```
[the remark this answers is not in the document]
vishwa2710: Yeah, more numerically stable.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        previousRmsError,
+        steps.getSize(),
+        terminationCriteria,
+        estimationStateBuilder.build(anInitialGuessState.getInstant(), XNom),
+        PHat,
+        frisbeeCovariance,
+        steps
+    );
+}
+
```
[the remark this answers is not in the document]
vishwa2710: That was by design. In python it accepts multiple signatures, harder to implement that in C++. Feel free to add an MR on top with those methods.

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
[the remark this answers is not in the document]
vishwa2710: I think the tests I've added are sufficient and well defined, we can leave the tests on the Loft python side as is and when we swap the C++ bound solver for the python one you can see if the tests still align.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,427 @@
...
+        const State currentState = estimationStateBuilder.build(anInitialGuessState.getInstant(), XNom);
+        const Array<State> computedStates = aGenerateStatesCallback(currentState, referenceInstants);
+
+        // Convert computed states to matrix form
+        MatrixXd computedStatesCoordinates(referenceStateCount, referenceStateDimension);
+        for (Size i = 0; i < referenceStateCount; ++i)
+        {
+            computedStatesCoordinates.row(i) =
+                computedStates[i].inFrame(estimationStateFrame).extractCoordinates(referenceStateSubsets);
+        }
```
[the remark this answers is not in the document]
vishwa2710: I believe #2 is faster as a matrixXd than a loop on state subtractions because Eigen is more efficient.

I think we should just do #2 and allow the users to change everything to the same frame etc. at the higher level if they want. It's what I've done in the ODLeastSquaresSolver.

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,436 @@
...
+                solver_.solve(State::Undefined(), referenceStates_, generateStates_),
+                ostk::core::error::runtime::Undefined
+            );
+        }
+
+        {
+            EXPECT_THROW(solver_.solve(initialGuessState_, {}, generateStates_), ostk::core::error::runtime::Undefined);
+        }
+
+        {
```
apaletta3, line 285: Add a comment as to why this should error?

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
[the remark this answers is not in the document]
apaletta3: These tests are pretty good, but we're still missing some coverage of the math compared to what we had in python.
We should add:
- the same toy orbit problem scenario and each of its weighting test cases
- solving success with different input reference frames

apaletta3 (commented):
🏓

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,409 @@
...
+    return (coordinates.transpose() * coordinates) / count;
+}
+
+LeastSquaresSolver LeastSquaresSolver::Default()
+{
+    return LeastSquaresSolver(20, 1.0);
+}
+
+VectorXd LeastSquaresSolver::extractSigmas(
+    const std::unordered_map<CoordinateSubset, VectorXd>& aSigmas, const StateBuilder& aStateBuilder
```
phc1990, line 364: What I meant in my other unresolved comment, is that I see a Map<Subset, VectorXd> so I just though, would using a State make your life easier in this function?

State is nothing but a Map<Subset,VectorXd> (with a timestamp and frame that you wouldn't really care about here). 

So I was just wondering whether building a state rather than a map might simplify your logic below:

for instance this:

```
Size totalSize = 0;
    for (const auto& subset : stateSubsets)
    {
        totalSize += subset->getSize();
    }
```

can be done by simply accessing the number of coordinates in the state, 
or the iteration below might also be simply accessing a subset of choice

not saying you should change the signature of this function
but you can go from std::unordered_map<CoordinateSubset, VectorXd>& aSigmas to State aSigmasState easily if it helps

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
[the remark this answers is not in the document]
vishwa2710: Spoke in person and will attempt to add these tests in a follow up MR

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Solver/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,281 @@
...
+                Args:
+                    aRmsError (float): The RMS error.
+                    anXHat (np.ndarray): The X hat vector.
+            )doc",
+            arg("aRmsError"),
+            arg("anXHat")
+        )
+        .def("__str__", &(shiftToString<LeastSquaresSolver::Step>))
+        .def("__repr__", &(shiftToString<LeastSquaresSolver::Step>))
+        .def_readonly(
```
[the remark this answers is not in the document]
vishwa2710: Yeah I think it's fine to have them as mutable in C++, they're only really being created in the LeastSquaresSolver

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,436 @@
...
+    std::unordered_map<CoordinateSubset, VectorXd> initialStateSigmas_ = {};
+
+    Array<State> referenceStates_;
+    const Size maxIterationCount_ = 20;
+    const Real rmsUpdateThreshold_ = 1e-2;
+    const FiniteDifferenceSolver finiteDifferenceSolver_ = FiniteDifferenceSolver::Default();
+    const LeastSquaresSolver solver_ = {maxIterationCount_, rmsUpdateThreshold_, finiteDifferenceSolver_};
+
+    // Define generate states callback for harmonic oscillator
+    static constexpr auto generateStates_ = [](const State& state, const Array<Instant>& instants) -> Array<State>
```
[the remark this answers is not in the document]
vishwa2710: Yeah if you're defining this at the class definition level it needs to be a constexpr, vs if we define it in the `SetUp` method then it doesn't need to be a constexpr.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
vishwa2710: You could... but it's kind of strange storing uncertainty values in the state, atleast this feels a bit more explicit. I agree though that there is nothing that precludes that. But then we have to consider what instant should be for that state/uncertainty value. I think this warrants a deeper discussion further down the line.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+        {
+            return state.getInstant();
+        }
+    );
+
+    State currentEstimatedState = State::Undefined();
+
+    const auto computeObservationsCoordinates = [&](const State& state, const Array<Instant>& instants) -> MatrixXd
+    {
+        const Array<State> states = aStateGenerator(state, instants);
```
apaletta3, line 227:
```suggestion
        const Array<State> computedObservationsStates = aStateGenerator(state, instants);
```

include/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.hpp
```diff
@@ -0,0 +1,177 @@
...
+    /// @return Finite difference solver
+    FiniteDifferenceSolver getFiniteDifferenceSolver() const;
+
+    /// @brief Solve the non-linear least squares problem
+    ///
+    /// @param anInitialGuessState Initial guess state
+    /// @param anObservationArray Array of observations
+    /// @param anInitialGuessSigmas Dictionary of sigmas for initial guess
+    /// @param anObservationSigmas Dictionary of sigmas for observations
+    /// @param aStateGenerator Function to generate states
```
apaletta3, line 134:
```suggestion
    /// Ref: https://www.sciencedirect.com/book/9780126836301/statistical-orbit-determination (Chapter 4, pg 196 for algorithm used)
    ///
    /// @param anInitialGuessState Initial guess state
    /// @param anObservationArray Array of observations
    /// @param aStateGenerator Function to generate states
    /// @param anInitialGuessSigmas Dictionary of sigmas for initial guess
    /// @param anObservationSigmas Dictionary of sigmas for observations
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+LeastSquaresSolver::Analysis LeastSquaresSolver::solve(
+    const State& anInitialGuessState,
+    const Array<State>& anObservationArray,
+    const std::function<Array<State>(const State&, const Array<Instant>&)>& aStateGenerator,
+    const std::unordered_map<CoordinateSubset, VectorXd>& anInitialGuessSigmas,
+    const std::unordered_map<CoordinateSubset, VectorXd>& anObservationSigmas
+) const
+{
+    if (!anInitialGuessState.isDefined())
```
apaletta3, line 137:
```suggestion
    // Notation used: https://www.sciencedirect.com/book/9780126836301/statistical-orbit-determination (Chapter 4, pg 196 for algorithm used)

    if (!anInitialGuessState.isDefined())
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+        // y = Y - G(X∗) (observed - computed = residual for current observations)
+        residuals = observationCoordinates - computedStatesCoordinates;
+
+        // H(t,t₀) = ∂G(X∗)/∂X∗₀ (sensitivty matrix for all observations at tᵢ w.r.t. nominal trajectory at epoch
+        // t₀)
+        const Array<MatrixXd> HFull = finiteDifferenceSolver_.computeStateTransitionMatrix(
+            currentEstimatedState, observationInstants, computeObservationsCoordinates
+        );
+
+        // Loop through each observations
```
apaletta3, line 263:
```suggestion
        // Loop through each observation
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+        // Initialize the Pₑ of equation 19 in https://ntrs.nasa.gov/citations/20140011726
+        PHatFrisbee = MatrixXd::Zero(stateDimension, stateDimension);
+
+        // G(X∗ᵢ) (computed observations) for all observation instants
+        currentEstimatedState = estimationStateBuilder.build(anInitialGuessState.getInstant(), XNom);
+        const MatrixXd computedStatesCoordinates =
+            computeObservationsCoordinates(currentEstimatedState, observationInstants);
+
+        // Compute residuals
+        // y = Y - G(X∗) (observed - computed = residual for current observations)
```
apaletta3, line 254:
```suggestion
        // y = Y - G(X∗) (observed - computed = residuals)
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+        }
+        return coordinates;
+    };
+
+    // Main iteration loop
+    for (Size iteration = 0; iteration < maxIterationCount_; ++iteration)
+    {
+        // Initialize matrices that will be accumulated in the loop
+        Lambda = PAprioriInverse;                 // Λ = P̄⁻¹
+        VectorXd N = PAprioriInverse * xApriori;  // N = P̄⁻¹ x̄
```
apaletta3, line 243:
```suggestion
        Lambda = PAprioriInverse;                 // Λ = P̄⁻¹ (shorthand for Hᵀ R⁻¹ H)
        VectorXd N = PAprioriInverse * xApriori;  // N = P̄⁻¹ x̄ (shorthand for H^T R^-1 y)
```

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,389 @@
...
+        throw ostk::core::error::RuntimeError("Sigma count does not match State Coordinate Subset count.");
+    }
+
+    Size totalSize = 0;
+    for (const auto& subset : stateSubsets)
+    {
+        totalSize += subset->getSize();
+    }
+
+    VectorXd sigmasCoordinates(totalSize);
```
apaletta3, line 359:
```suggestion
    VectorXd sigmasCoordinates(aStateBuilder.getSize());
```

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    State trueState_ = State::Undefined();
+    State initialGuessState_ = State::Undefined();
+
+    VectorXd positionSigmas_ {3};
+    VectorXd velocitySigmas_ {3};
+
+    std::unordered_map<CoordinateSubset, VectorXd> observationSigmas_ = {};
+
+    std::unordered_map<CoordinateSubset, VectorXd> initialStateSigmas_ = {};
```
apaletta3, line 111:
```suggestion
    std::unordered_map<CoordinateSubset, VectorXd> initialStateSigmas_ = {};

    std::unordered_map<CoordinateSubset, VectorXd> observationSigmas_ = {};
```

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+    const LeastSquaresSolver::Analysis analysis_ = LeastSquaresSolver::Analysis(
+        observationCount_,
+        terminationCriteria,
+        solutionState_,
+        solutionCovariance_,
+        solutionFrisbeeCovariance_,
+        solutionResidualMatrix_,
+        steps_
+    );
+};
```
apaletta3, line 178: Can you put these two above the LSSolver for consistency?

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+        const std::string output = testing::internal::GetCapturedStdout();
+    }
+}
+
+TEST_F(OpenSpaceToolkit_Astrodynamics_Solver_LeastSquaresSolver_Analysis, Constructor)
+{
+    {
+        EXPECT_NO_THROW(
+            LeastSquaresSolver::Analysis(
```
apaletta3, line 213: Worth checking that the RMS error value is what you expect it to be since it is calculated in the constructor

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+        );
+    }
+
+    // Test empty observations
+    {
+        EXPECT_THROW(
+            solver_.solve(initialGuessState_, Array<State>::Empty(), generateStates_),
+            ostk::core::error::runtime::Undefined
+        );
+    }
```
apaletta3, line 326: If you agree with deleting the if statement that corresponds to this test, you can add one that corresponds the matrix dimension condition (to yield a full rank H matri)

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+            const Real time = timeStep * (Real)i;
+            const Real position = std::cos(time);
+            const Real velocity = -std::sin(time);
+
+            const Position truePosition = Position::Meters({position, 0.0, 0.0}, Frame::GCRF());
+            const Velocity trueVelocity = Velocity::MetersPerSecond({velocity, 0.0, 0.0}, Frame::GCRF());
+
+            // Add noise to the true state
+            std::mt19937 rng(42);  // Fixed seed for reproducibility
+            std::uniform_real_distribution<double> dist(-0.01, 0.01);
```
apaletta3, line 75: This is the wrong distribution to use, it should be a normal distribution, that way you can exactly match it with the sigmas

test/OpenSpaceToolkit/Astrodynamics/Solver/LeastSquaresSolver.test.cpp
```diff
@@ -0,0 +1,422 @@
...
+        positionSigmas_ << 0.1, 0.1, 0.1;
+        velocitySigmas_ << 0.1, 0.1, 0.1;
+
+        observationSigmas_ = {
+            {*CartesianPosition::Default(), positionSigmas_}, {*CartesianVelocity::Default(), velocitySigmas_}
+        };
+
+        initialStateSigmas_ = {
+            {*CartesianPosition::Default(), positionSigmas_}, {*CartesianVelocity::Default(), velocitySigmas_}
+        };
```
apaletta3, line 100: Needs to match the sigma specified in the normal distribution, and the initial guess sigmas should be different than the observation sigmas, so that you can actually test the weighting properly

apaletta3 (commented):
Last round of comments 🏓 and we're ready to land, I promise

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
vishwa2710: I think in the scope of the lambda we don't need to have such verbose naming, the name of the lambda clarifies this and it's a short lambda.

src/OpenSpaceToolkit/Astrodynamics/Solvers/LeastSquaresSolver.cpp
```diff
@@ -0,0 +1,420 @@
...
+            // N = N + (Hᵢᵀ R⁻¹ yᵢ)
+            N += H.transpose() * RInv * y;
+
+            // Pₑ = Pₑ + Hᵢᵀ R⁻¹ yᵢ yᵢᵀ R⁻¹ Hᵢ
+            PHatFrisbee += H.transpose() * RInv * y * y.transpose() * RInv * H;
+        }
+
+        // Solve normal equations
+        // x̂ = Λ⁻¹ N
+        xHat = Lambda.ldlt().solve(N);
```
apaletta3, line 313: CodeRabbit mentioned that .ldlt() doesn't work well for ill-conditioned matrices? Could be worth checking when solving I gues?

vishwa2710: Eh, I think it's fine. It'll complain in the solve regardless.

PR #504, opened 2025-01-27
chore: add convenience method for total size of state builder

apaletta3:
Avoids having to loop through all coordinate subsets and accumulate their sizes each time.
It also avoids the necessity of accessing the underlying coordinate broker, since at some point I believe we will remove it from the API and have it just be something intenral.

Example inconvenience that this helps alleviate:
```
Size totalSize = 0;
for (const auto& subset : stateSubsets)
{
    totalSize += subset->getSize();
}
```

or 
```
Size totalSize = stateBuilder.accessCoordinateBroker.getNumerOfCoordinates()
```

I called it `.getSize` because that is what the equivalent method in the `State` class does

## Summary by CodeRabbit

- **New Features**
  - Added `get_size()` method to `StateBuilder` class to retrieve total coordinate size.

- **Tests**
  - Enhanced test coverage for `StateBuilder` class.
  - Added assertions to validate `get_size()` method functionality.
  - Verified error handling for undefined `StateBuilder` instances.

- **Documentation**
  - Updated method signatures with explicit return type annotations.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
@@ -21,7 +21,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_StateBuilder(pybind11::mo
         "StateBuilder",
         R"doc(
             This class makes it convenient to build a `State` object.
-        
+
         )doc"
     )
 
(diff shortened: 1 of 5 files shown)


alex-liang3 (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
```diff
@@ -21,7 +21,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_StateBuilder(pybind11::mo
         "StateBuilder",
         R"doc(
             This class makes it convenient to build a `State` object.
-        
+
```
vishwa2710, line 24: I don't think we need all these extra lines here?

vishwa2710 (approved):

PR #510, opened 2025-02-03
feat: tle solver

vishwa2710:
## Summary by CodeRabbit

- **New Features**
  - Introduced advanced orbital estimation using Two-Line Element (TLE) data. This new capability, accessible via the enhanced Python interface, allows users to generate and analyze orbital elements more effectively.
  
- **Tests**
  - Expanded the testing framework to validate the TLE estimation process under various scenarios, ensuring reliable error handling and robust performance.


src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -0,0 +1,307 @@
...
+    const LeastSquaresSolver::Analysis analysis = solver_.solve(
+        initialGuessTLEState, observationsInEstimationFrame, generateStates, anInitialGuessSigmas, anObservationSigmas
+    );
+
+    // Convert solution state to TLE
+    const TLE determinedTLE = TLEStateToTLE(analysis.estimatedState);
+
+    return Analysis(determinedTLE, analysis);
+}
+
```
apaletta3, line 228: why not add an `estimateOrbit` method here too? Just like for `Propagated`

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -0,0 +1,307 @@
...
+      defaultBStar_(1e-4),
+      firstDerivativeMeanMotionDividedBy2_(0.0),
+      secondDerivativeMeanMotionDividedBy6_(0.0),
+      ephemerisType_(0),
+      elementSetNumber_(0),
+      tleStateBuilder_(StateBuilder::Undefined())
+{
+    // Setup coordinate subsets for TLE state
+    Array<Shared<const CoordinateSubset>> coordinateSubsets = {
+        std::make_shared<CoordinateSubset>("INCLINATION", 1),
```
apaletta3, line 80: tbh I think we should add these as actually coordinate subset classes

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -0,0 +1,307 @@
...
+using ostk::physics::coordinate::Velocity;
+using ostk::physics::environment::gravitational::Earth;
+using ostk::physics::time::Instant;
+
+using ostk::astrodynamics::trajectory::orbit::model::blm::BrouwerLyddaneMeanLong;
+using ostk::astrodynamics::trajectory::orbit::model::SGP4;
+
+TLESolver::Analysis::Analysis(const TLE& aDeterminedTLE, const LeastSquaresSolver::Analysis& anAnalysis)
+    : determinedTLE(aDeterminedTLE),
+      solverAnalysis(anAnalysis)
```
apaletta3, line 33: Same, I feel that either we expand all the fields to a native TLESolver Analysis, or we expand none of the fields and just use the LSSolver Analysis

include/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.hpp
```diff
@@ -0,0 +1,210 @@
...
+    Integer revolutionNumber_;
+    bool fitWithBStar_;
+    Shared<const Frame> estimationFrameSPtr_;
+
+    mutable Real defaultBStar_;
+    Real firstDerivativeMeanMotionDividedBy2_;
+    Real secondDerivativeMeanMotionDividedBy6_;
+    Integer ephemerisType_;
+    Integer elementSetNumber_;
+    StateBuilder tleStateBuilder_;
```
apaletta3, line 181: This feels like a lot of memeber variables, maybe we can group them better with an intermediate object or something?

include/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.hpp
```diff
@@ -0,0 +1,210 @@
...
+using ostk::physics::unit::Time;
+
+using ostk::astrodynamics::solver::LeastSquaresSolver;
+using ostk::astrodynamics::trajectory::orbit::model::sgp4::TLE;
+using ostk::astrodynamics::trajectory::State;
+using ostk::astrodynamics::trajectory::state::CoordinateSubset;
+using ostk::astrodynamics::trajectory::StateBuilder;
+
+/// @brief Class for solving TLE elements using least squares
+class TLESolver
```
apaletta3, line 59: This should technically be called a TLELeastSquaresSolver

apaletta3 (commented):
Initial interface review

include/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.hpp
vishwa2710: We could make a TLE Solver Configuration object but it'll kinda be the same thing, lots of members. Let me noodle with it.

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
vishwa2710: Same as the other comment, since a TLE is it's own object I think it's convenient having this for the user.

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
vishwa2710: Yeah we can add some static constructors. Not a priority IMO since it's only used internally in the solver.

PR #515, opened 2025-02-10
feat: improve TLE Solver interface

apaletta3:
Last small MR with a few small improvements to the TLE solver.


src/OpenSpaceToolkit/Astrodynamics/Estimator/OrbitDeterminationSolver.cpp
```diff
@@ -95,9 +95,9 @@ OrbitDeterminationSolver::Analysis OrbitDeterminationSolver::estimate(
     const State initialGuessStateInEstimationFrame = anInitialGuessState.inFrame(estimationFrameSPtr_);
 
     const Array<State> observationStatesInEstimationFrame = anObservationStateArray.map<State>(
-        [estimationFrameSPtr = estimationFrameSPtr_](const State& aState) -> State
+        [this](const State& aState) -> State
```
vishwa2710, line 98: Why make it copy the entire object when we just need 1 shared pointer?

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -79,17 +91,12 @@ TLESolver::TLESolver(
 {
     // Setup coordinate subsets for TLE state
     Array<Shared<const CoordinateSubset>> coordinateSubsets = {
-        std::make_shared<CoordinateSubset>("INCLINATION", 1),
-        std::make_shared<CoordinateSubset>("RAAN", 1),
-        std::make_shared<CoordinateSubset>("ECCENTRICITY", 1),
-        std::make_shared<CoordinateSubset>("AOP", 1),
-        std::make_shared<CoordinateSubset>("MEAN_ANOMALY", 1),
-        std::make_shared<CoordinateSubset>("MEAN_MOTION", 1)
+        InclinationSubset, RaanSubset, EccentricitySubset, AopSubset, MeanAnomalySubset, MeanMotionSubset
```
vishwa2710, line 94:
```suggestion
        InclinationSubset, RaanSubset, EccentricitySubset, AopSubset, MeanAnomalySubset, MeanMotionSubset,
```
This will make it format nicer.

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -304,13 +309,8 @@ State TLESolver::CartesianStateAndBStarToTLEState(const State& aCartesianState,
             .in(Derived::Unit::RevolutionPerDay())
     };
 
-    if (fitWithBStar_)
+    if (estimateBStar_)
     {
-        if (!aBStar.isDefined())
```
vishwa2710, line 309: Why remove this check?

src/OpenSpaceToolkit/Astrodynamics/Estimator/OrbitDeterminationSolver.cpp
apaletta3: Its not copying the entire object, `this` is a pointer, so it just captures the pointer by value, which is effectively the same as capturing the `estimationFrameSPtr` by value

apaletta3:
```suggestion
        [estimationFrameSPtr_](const State& aState) -> State
```
I guess we could also just do this
or 
```suggestion
        [&estimationFrameSPtr_](const State& aState) -> State
```
although we would have to check if the latter implementation is support in cpp versions before cpp20, cause I know [&this] capture is not supported before cpp20

PR #526, opened 2025-02-17
refactor!: profile target renaming

apaletta3:
- Renaming elements of the Profile class including classes within its namespace and method to more accurately reflect certain terminology in the attitude world
- Fixed a unit test in python for profile

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
@@ -72,16 +72,7 @@ using ostk::astrodynamics::trajectory::State;
 class Profile
 {
    public:
-    enum class PointingMode
-    {
-
-        Undefined,  ///< Undefined pointing mode
-        Inertial,   ///< Inertial pointing mode (the spacecraft points to a celestial object)
-        Nadir,      ///< Nadir pointing mode (the spacecraft points points "directly down")
-        Target,     ///< Target pointing mode (the spacecraft points to a given target position)
-        Custom      ///< Custom pointing mode
-    };
-
+    // @brief Represents an axis of your spacecraft.
     enum class Axis
     {
         X,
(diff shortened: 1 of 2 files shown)


include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -72,104 +72,109 @@ using ostk::astrodynamics::trajectory::State;
...
-    enum class PointingMode
-    {
-
-        Undefined,  ///< Undefined pointing mode
-        Inertial,   ///< Inertial pointing mode (the spacecraft points to a celestial object)
-        Nadir,      ///< Nadir pointing mode (the spacecraft points points "directly down")
-        Target,     ///< Target pointing mode (the spacecraft points to a given target position)
-        Custom      ///< Custom pointing mode
-    };
-
```
apaletta3, line 84: This is no longer used anywhere in ostk

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -72,104 +72,109 @@ using ostk::astrodynamics::trajectory::State;
...
-        GeodeticNadir,       /// Negative of the geodetic normal of the satellite in the ECI frame
-        Trajectory,          /// Points towards the provided trajectory, eg. Ground Station in ECEF
-        Sun,                 /// The position of the Sun
-        Moon,                /// The position of the Moon
-        VelocityECI,         /// The velocity vector in the ECI frame
-        VelocityECEF,        /// The velocity vector in the ECEF frame
-        OrbitalMomentum,     /// The orbital momentum vector of the satellite in the ECI frame
-        OrientationProfile,  /// Points towards a profile of orientations in the ECI frame
-        Custom,              /// Custom target
+        // Simple pre-defined targets
```
vishwa2710, line 86: Not sure if this comment adds much value, I think it's better to extend the explanations in the enums so that they appear as docstrings as well.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -72,104 +72,109 @@ using ostk::astrodynamics::trajectory::State;
...
+        VelocityECEF,     /// The velocity vector in the ECEF frame
+        OrbitalMomentum,  /// The orbital momentum vector of the satellite in the ECI frame
+
+        // More complex targets
+        Trajectory,  /// Points towards the provided trajectory, eg. Ground Station in ECEF or another satellite
+
+        // Custom pointing
+        CustomProfileECI,           /// Pointing pre-defined via a profile of vectors in the ECI frame
+        CustomProfileGeneratorECI,  /// Pointing generated at each timestep by a function of that returns a vector in
+                                    /// the ECI frame
```
vishwa2710, line 101:
```suggestion
        Trajectory,  /// Points towards the provided trajectory, eg. Ground Station in ECEF or another satellite
        Custom /// Pointing generated at each timestep by a function that returns a Quaternion in the ECI frame
```
Not sure if we want 2 different custom profiles here? They're both just technically generators?

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -72,104 +72,109 @@ using ostk::astrodynamics::trajectory::State;
...
     };
 
-    /// @brief Represents a target for alignment or pointing purposes.
-    class Target
+    /// @brief Represents pointing of one axis of your spacecraft towards a target.
+    class Pointing
```
vishwa2710, line 105: I agree target might not have been the best word, but let's remember that it's accessed via the Profile class so it's `Profile.Target`.
I don't think we should call it pointing as if you read the brief you're basically saying `pointing of one axis of your spacecraft towards a **target**`

vishwa2710 (commented):
Overall I agree we can improve some of the naming here. Let's be careful though to not make another round of breaking changes unless absolutely necessary, especially if it's just naming. We can make a  Github issue highlighting the naming we would like in the future and then do it in one big refactor

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
apaletta3: Ah good point

PR #539, opened 2025-04-15
feat: add state isNear method

apaletta3:
This is something we do quite often, so might as well formalize it here to shortcut it in the future

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
@@ -499,6 +499,31 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
     };
 }
 
+bool State::isNear(const State& aState, const std::unordered_map<CoordinateSubset, Real>& aToleranceMap) const
+{
+    if (aToleranceMap.size() != coordinatesBrokerSPtr_->accessSubsets()->getSize())
+    {
+        throw ostk::core::error::runtime::Wrong("Tolerance Map");
+    }
+
+    const State deltaState = aState - *this;
+
+    bool isNear = true;
+    for (const auto& [subset, tolerance] : aToleranceMap) {
+        if (!coordinatesBrokerSPtr_->hasSubset(subset)) {
+            throw ostk::core::error::runtime::Wrong("Tolerance Map Coordinate Subset");
+        }
+
+        const VectorXd deltaCoordinates = deltaState.extractCoordinate(subset);
+
+        if (deltaCoordinates.norm() > tolerance) {
+            isNear = false;
+            break;
+        }
+    }
+    return isNear;
+}
+
 void State::print(std::ostream& anOutputStream, bool displayDecorator) const
 {
     using ostk::core::type::String;
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State.cpp
@@ -36,7 +36,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State(pybind11::module& a
             init<const Instant&, const Position&, const Velocity&>(),
             R"doc(
                  Utility constructor for Position/Velocity only.
-                 
+
                  Args:
                      instant (Instant): An instant
                      position (Position): The cartesian position at the instant
@@ -56,7 +56,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State(pybind11::module& a
                 const Shared<const Frame>&>(),
             R"doc(
                  Utility constructor for Position/Velocity/Attitude/Angular velocity.
-                 
+
                  Args:
                      instant (Instant): An instant
                      position (Position): The cartesian position at the instant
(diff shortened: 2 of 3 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +499,31 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
 }
 
+bool State::isNear(const State& aState, const std::unordered_map<CoordinateSubset, Real>& aToleranceMap) const
+{
+    if (aToleranceMap.size() != coordinatesBrokerSPtr_->accessSubsets()->getSize())
+    {
+        throw ostk::core::error::runtime::Wrong("Tolerance Map");
+    }
+
+    const State deltaState = aState - *this;
```
apaletta3, line 509: All the other `*this == aState` checks are done during this subtraction already so no need to redo them here

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +499,31 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+
+    bool isNear = true;
+    for (const auto& [subset, tolerance] : aToleranceMap) {
+        if (!coordinatesBrokerSPtr_->hasSubset(subset)) {
```
vishwa2710, line 513: Is this necessary? Won't the subtract method raise an error if the coordinate subsets are not identical?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +499,31 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+            throw ostk::core::error::runtime::Wrong("Tolerance Map Coordinate Subset");
+        }
+
+        const VectorXd deltaCoordinates = deltaState.extractCoordinate(subset);
+
+        if (deltaCoordinates.norm() > tolerance) {
```
vishwa2710, line 519: This makes an inherent assumption that you want to take the norm of each subset. What if you have a composite coordinate subset composed of different units or elements?

vishwa2710: Might be worth having two different methods, one which does the norm of each subset, and another which compares each element.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +499,31 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+            isNear = false;
+            break;
+        }
+    }
+    return isNear;
```
vishwa2710, line 524: Instead of returning a single boolean, I would recommend return an array of booleans, and letting the user decide if they want to do an `all` or `any`

vishwa2710: Better yet, return a map of coordinate subset -> bool

vishwa2710 (changes requested):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
apaletta3: yeah that's not a bad idea. Maybe we can have a flag that determine which mode this works in, norm or per coordinate

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +497,77 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+            throw ostk::core::error::runtime::Wrong("Tolerance Array");
+        }
+
+        Array<bool> isNearArray;
+        isNearArray.reserve(deltaCoordinates.size());
+        for (int i = 0; i < deltaCoordinates.size(); ++i)
+        {
+            isNearArray.add(std::abs(deltaCoordinates[i]) <= toleranceArray[i]);
+        }
+        isNearMap[subsetSPtr] = isNearArray;
```
vishwa2710, line 566:
```suggestion
        Eigen::ArrayXb isNear = deltaCoordinates.abs() <= toleranceArray
```
Can probably simplify by leveraging eigen here.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/State.cpp
```diff
@@ -499,6 +497,77 @@ State State::inFrame(const Shared<const Frame>& aFrameSPtr) const
...
+        if (!coordinatesBrokerSPtr_->hasSubset(subsetSPtr))
+        {
+            throw ostk::core::error::runtime::Wrong("Tolerance Map Coordinate Subset");
+        }
+
+        const VectorXd deltaCoordinates = deltaState.extractCoordinate(subsetSPtr);
+
+        if (toleranceArray.size() != deltaCoordinates.size())
+        {
+            throw ostk::core::error::runtime::Wrong("Tolerance Array Size");
```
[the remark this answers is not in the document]
vishwa2710: Worth incorporating these I believe. The "Wrong" can be really confusing and having more explicit runtime errors is better IMO.

vishwa2710 (approved):
Mostly nits, feel free to resolve the eigen comment if you don't agree but let's improve the error logs.

apaletta3:
Holding on merging as there might be some improvements to the interface we'd like to make, such as:
- Not forcing the user to provider all the coordinate subsets of the state, can be just a subset of it
- ensure that it works with attitude subsets as well (add those to the tests)
- perhaps have a state comparator class or factory method to help with comparing two lists of states to each other
- Also allow for a mode to check whether values are close to 0 (aka checking the norm of a state for various subsets)

PR #542, opened 2025-04-18
fix: within range angular condition, wrapping boundaries

vishwa2710:
This MR addresses an issue in the AngularCondition::WithinRange criterion when the angular range wraps across a boundary.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of angular ranges that cross the 0/360-degree boundary, ensuring accurate evaluation of values within these ranges.

- **Tests**
	- Added new test cases to verify correct behavior for both standard and boundary-crossing angular ranges.


apaletta3 (commented):
great

PR #548, opened 2025-04-29
feat: target velocity profile target type

vishwa2710:
- Non breaking change, add TargetPosition enum type (to replace the Trajectory enum in a future release).
- Add TargetVelocity enum type, which generates a direction vector aligned with the velocity of the target.

## Summary by CodeRabbit

- **New Features**
  - Added support for targeting both position and velocity in trajectory-based pointing profiles, allowing more flexible orientation options.
- **Bug Fixes**
  - Improved clarity and accuracy of documentation strings for trajectory target attributes.
- **Tests**
  - Expanded test coverage for new target types and added validation for yaw compensation in attitude profiles.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
@@ -56,6 +56,8 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
         .value("GeocentricNadir", Profile::TargetType::GeocentricNadir, "Geocentric nadir")
         .value("GeodeticNadir", Profile::TargetType::GeodeticNadir, "Geodetic nadir")
         .value("Trajectory", Profile::TargetType::Trajectory, "Trajectory")
+        .value("TargetPosition", Profile::TargetType::TargetPosition, "Target position")
+        .value("TargetVelocity", Profile::TargetType::TargetVelocity, "Target velocity")
         .value("Sun", Profile::TargetType::Sun, "Sun")
         .value("Moon", Profile::TargetType::Moon, "Moon")
         .value("VelocityECI", Profile::TargetType::VelocityECI, "Velocity in ECI")
@@ -121,10 +123,32 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
             arg("anti_direction") = false
         )
 
+        .def_static(
+            "target_position",
+            &Profile::TrajectoryTarget::TargetPosition,
+            R"doc(
+                Create a target, which produces a vector pointing from the observer to the target position.
+            )doc",
+            arg("trajectory"),
+            arg("axis"),
+            arg("anti_direction") = false
+        )
+
+        .def_static(
+            "target_velocity",
+            &Profile::TrajectoryTarget::TargetVelocity,
+            R"doc(
+                Create a target, which produces a vector pointing along the target velocity.
+            )doc",
+            arg("trajectory"),
+            arg("axis"),
+            arg("anti_direction") = false
+        )
+
         .def_readonly(
             "trajectory",
             &Profile::TrajectoryTarget::trajectory,
-            "The trajectory of the target. Required only if the target type is `Trajectory`."
+            "The trajectory of the target. Used to compute the target position or velocity."
         )
 
         ;
@@ -156,7 +180,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
         .def_readonly(
             "orientation_profile",
             &Profile::OrientationProfileTarget::orientationProfile,
-            "The orientation profile of the target"
+            "The orientation profile of the target."
         )
 
         ;
@@ -188,7 +212,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
         .def_readonly(
             "orientation_generator",
             &Profile::CustomTarget::orientationGenerator,
-            "The orientation generator of the target"
+            "The orientation generator of the target."
         )
 
         ;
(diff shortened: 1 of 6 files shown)


include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -128,9 +129,43 @@ class Profile
         /// @param aTrajectory The trajectory to point towards.
         /// @param anAxis The axis of the target.
         /// @param isAntiDirection Whether the target is in the anti-direction.
-        TrajectoryTarget(const Trajectory& aTrajectory, const Axis& anAxis, const bool& isAntiDirection = false);
+        TrajectoryTarget(
+            const ostk::astrodynamics::Trajectory& aTrajectory, const Axis& anAxis, const bool& isAntiDirection = false
+        );
+
```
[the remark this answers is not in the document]
apaletta3: agree with this suggestion

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile.cpp
```diff
@@ -425,6 +480,28 @@ Vector3d Profile::ComputeTargetDirectionVector(const State& aState, const Trajec
...
+
+    const Vector3d relativePositionDirection =
+        (slidingTargetState_GCRF.getPosition().getCoordinates() - aState.getPosition().getCoordinates()).normalized();
+    const Vector3d relativeVelocityDirection =
+        ITRF_GCRF_transform.applyToVector(slidingTargetState_ITRF.getVelocity().getCoordinates()).normalized();
+
+    const Vector3d relativeNormalDirection = relativePositionDirection.cross(relativeVelocityDirection).normalized();
+
+    return relativeNormalDirection.cross(relativePositionDirection);
+}
```
apaletta3, line 503: now that we properly construct the trajectory velocity in gcrf at trajectory construction time, why can't we just do this computation the same way as the ComputeTargetDirectionVector method? Where you do
`targetVelocityCoordinates - satelliteVelocityCoordinates).normalized()`?

I believe you'd get the same result. Here's youre doing the cross product of the position and velocity and then of the position and normal which gives you back a vector aligned with the original velocity direction

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
```diff
@@ -56,6 +56,8 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
         .value("GeocentricNadir", Profile::TargetType::GeocentricNadir, "Geocentric nadir")
         .value("GeodeticNadir", Profile::TargetType::GeodeticNadir, "Geodetic nadir")
         .value("Trajectory", Profile::TargetType::Trajectory, "Trajectory")
```
apaletta3, line 58: can you add a depracation warning here?

test/OpenSpaceToolkit/Astrodynamics/Flight/Profile.test.cpp
```diff
@@ -883,6 +889,68 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Flight_Profile, OrientationProfileTarget)
...
+    const State compensatedState = compensatedProfile.getStateAt(instant);
+    const State uncompensatedState = uncompensatedProfile.getStateAt(instant);
+
+    // Values taken from Orekit for unit test
+
+    EXPECT_NEAR(
+        compensatedState.getAttitude().angularDifferenceWith(uncompensatedState.getAttitude()).inDegrees(),
+        3.803545407107513,
+        1e-4
+    );
```
apaletta3, line 951: can you assert a positive angular differnce pointing westwards or something like that? Like to physically assert that the clocking axis (X here) has a westwards component?

apaletta3 (approved):
Great stuff! I believe you could simplify the yam compensation computation with my suggestion

test/OpenSpaceToolkit/Astrodynamics/Flight/Profile.test.cpp
vishwa2710: I can just assert the quaternion lines up exactly as Orekit.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
vishwa2710: Not sure how to add it at the enum level, added a comment

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile.cpp
vishwa2710: The position and velocity are not orthogonal necessarily. r x v -> Normal, which is orthogonal to the plane of r & v, and then crossing r x n gets us an orthogonal vector to the r & n plane which is not necessarily aligned with the velocity direction.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
```diff
@@ -55,7 +55,9 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
 
         .value("GeocentricNadir", Profile::TargetType::GeocentricNadir, "Geocentric nadir")
         .value("GeodeticNadir", Profile::TargetType::GeodeticNadir, "Geodetic nadir")
-        .value("Trajectory", Profile::TargetType::Trajectory, "Trajectory")
+        .value("Trajectory", Profile::TargetType::Trajectory, "Trajectory")  // Deprecated in favor of TargetPosition
```
apaletta3, line 58: nit: if there's a way to indicate a depracation in the bindings that would be great, if not no worries

apaletta3 (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
vishwa2710: There isn't without fully overriding the bindings which I feel is a bit excessive.

PR #554, opened 2025-05-14
feat: nadir trajectory model

vishwa2710:
This MR improves the interface for the geodetic nadir ground track trajectory:
- Improved naming (geodeticNadirGroundTrack instead of groundStripGeodeticNadir)
- Improved interface (doesn't require instants or a celestial, just the orbit)
- Improved functionality (works for any instant that is valid for the orbit)
- Adds a trajectory model for this.

Deprecates the `groundStripGeodeticNadir` method.

## Summary by CodeRabbit

- **New Features**
  - Introduced a Nadir trajectory model for geodetic nadir position computation of orbits.
  - Added Python bindings for the Nadir model with detailed functionality.
  - Added method to access the celestial object associated with an orbit.

- **Deprecations**
  - Deprecated the previous ground strip geodetic nadir trajectory method in favor of the Nadir model.

- **Tests**
  - Added extensive C++ and Python tests for the Nadir model covering construction, state calculation, cloning, and accessors.
  - Added tests for celestial object access in orbits.
  - Removed tests related to the deprecated ground strip geodetic nadir method.

- **Style**
  - Reformatted Python test method signatures for better readability.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.hpp
@@ -26,8 +26,6 @@ using ostk::physics::time::Interval;
 using ostk::astrodynamics::trajectory::Model;
 using ostk::astrodynamics::trajectory::State;
 
-#define DEFAULT_NADIR_STEP_SIZE Duration::Seconds(1e-2)
-
 /// @brief Nadir pointing trajectory model to represent the geodetic nadir position of an orbit on the surface of the
 /// earth
 class Nadir : public virtual Model
@@ -43,7 +41,7 @@ class Nadir : public virtual Model
     /// @param anOrbit The orbit of the nadir model
     /// @param aStepSize The step size for the nadir model. Defaults to 1e-2 seconds
     /// @return A nadir model
-    Nadir(const Orbit& anOrbit, const Duration& aStepSize = DEFAULT_NADIR_STEP_SIZE);
+    Nadir(const Orbit& anOrbit, const Duration& aStepSize = Duration::Seconds(1e-2));
 
     /// @brief Clone the nadir model
     ///
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.hpp
@@ -38,17 +38,17 @@ class Nadir : public virtual Model
     ///              Nadir nadirModel(orbit);
     /// @endcode
     ///
-    /// @param aPosition The position of the static model. Must be provided in the ITRF frame.
-    Nadir(const Orbit& orbit);
+    /// @param anOrbit The orbit of the nadir model.
+    Nadir(const Orbit& anOrbit);
 
-    /// @brief Clone the static model
+    /// @brief Clone the nadir model
     ///
     /// @code{.cpp}
     ///              Nadir nadirModel = { ... };
     ///              Nadir* clonedModel = nadirModel.clone();
     /// @endcode
     ///
-    /// @return A pointer to the cloned static model
+    /// @return A pointer to the cloned nadir model
     virtual Nadir* clone() const override;
 
     /// @brief Equality operator
@@ -59,7 +59,7 @@ class Nadir : public virtual Model
     ///              bool isEqual = (nadirModel1 == nadirModel2);
     /// @endcode
     ///
-    /// @param aNadirModel The static model to compare with
+    /// @param aNadirModel The nadir model to compare with
     /// @return True if the models are equal, false otherwise
     bool operator==(const Nadir& aNadirModel) const;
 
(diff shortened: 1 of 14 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    Size i = 0;
+    for (const auto& instant : instants)
+    {
+        times(i) = (instant - anInstant).inSeconds();
+
+        const State state = orbit_.getStateAt(instant);
+
+        const LLA lla =
+            LLA::FromPosition(state.getPosition().inFrame(Frame::ITRF(), instant), orbit_.accessCelestialObject())
```
phc1990, line 104: Celestial does not have an associated body-fixed frame in physics, otherwise we could use it.

We should not assume ITRF (erath-bound), but let the user specify it in the Nadir constructor.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.cpp
```diff
@@ -0,0 +1,161 @@
...
+                .onSurface();
+
+        const physics::coordinate::Position positionAtInstant =
+            physics::coordinate::Position::FromLLA(lla, orbit_.accessCelestialObject()).inFrame(Frame::GCRF(), instant);
```
phc1990, line 108: Do we need to convert to GCRF()

Could we not return the states in the original orbit frame?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.cpp
```diff
@@ -0,0 +1,161 @@
...
+        Interpolator::GenerateInterpolator(Interpolator::Type::CubicSpline, times, positionCoordinates.row(2)),
+    };
+
+    const Vector3d velocityCoordinates = {
+        interpolators[0]->computeDerivative(0.0),
+        interpolators[1]->computeDerivative(0.0),
+        interpolators[2]->computeDerivative(0.0),
+    };
+
+    const Velocity velocity = Velocity::MetersPerSecond(velocityCoordinates, Frame::GCRF());
```
phc1990, line 132: Idem

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    if (!this->isDefined())
+    {
+        throw ostk::core::error::runtime::Undefined("Nadir");
+    }
+
+    // interpolate the velocity using polynomial interpolation
+
+    static const Duration stepSize = Duration::Seconds(1e-2);
```
phc1990, line 81: maybe can be made an optional argument as part of the constructor

or, if we think it's too much exposure maybe a we can place it as a top-level file constant

phc1990 (commented):
Nice stuff, left minor comments.

The more relevant one is on the ITRF frame assumption.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.cpp
phc1990: God I just checked and FromLLA assumes ITRF() 😭 

So this is open to an Orbit pointing to a celestial object other than earth, but returning an earth-bound nadir trajectory (which would not be correct)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Nadir.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+/// @brief Nadir pointing trajectory model to represent the geodetic nadir position of an orbit on the surface of the
+/// earth
+class Nadir : public virtual Model
+{
+   public:
+    /// @brief Constructor
+    ///
+    /// @code{.cpp}
+    ///              Nadir nadir = { ... };
```
alex-liang3, line 37:
```suggestion
    ///              Orbit orbit = { ... };
```

phc1990 (approved):
Thanks for addressing the frame stuff, approved ✅

src/OpenSpaceToolkit/Astrodynamics/Trajectory.cpp
```diff
@@ -244,37 +246,12 @@ Trajectory Trajectory::GroundStrip(
...
-        const physics::coordinate::Position position =
-            physics::coordinate::Position::FromLLA(lla, celestialSPtr).inFrame(Frame::GCRF(), anInstant);
-
-        return position;
-    };
-
-    const Array<State> states = computeStates(positionGenerator, anInstantArray);
-
-    return Trajectory(states);
+    [[deprecated("Use Trajectory::GeodeticNadirGroundTrack instead.")]] return Trajectory(Nadir(anOrbit));
```
alex-liang3, line 254:
```suggestion
    [[deprecated("Use Trajectory::Nadir instead.")]] return Trajectory(Nadir(anOrbit));
```
(though I think `GeodeticNadirGroundTrack` is more explicit, but up to you - as long as it's consistent)

alex-liang3, line 254: I think there's some weird conflicts going on, I can't see `Trajectory::GeodeticNadirGroundTrack` in this PR, but it's in [this one](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/555/files)

alex-liang3 (changes requested):

apaletta3 (commented):
:100:

PR #556, opened 2025-05-15
docs: fix statebuilder docs

alex-liang3:
Some of the python docstrings are wrong and doesn't actually say what the StateBuilder will do

## Summary by CodeRabbit

- **Documentation**
  - Improved docstrings for the constructor and methods of the StateBuilder class to clarify return types and behavior descriptions.
- **Tests**
  - Updated test to prevent automatic display of plots during execution.


vishwa2710 (approved):

apaletta3 (approved):
great - if you want to keep the cpp docstrings in sync as well that would be great

vishwa2710 (approved):

PR #558, opened 2025-05-17
feat: improve tabulated profile model

vishwa2710:
- Add bindings for the trajectory tabulated model
- Improve the profile tabulated model to use a variety if interpolators, while using SLERP for quaternions

## Summary by CodeRabbit

- **New Features**
  - Enhanced the tabulated trajectory/profile model with explicit interpolation type support and refined quaternion interpolation using SLERP.
  - Added Python bindings for the tabulated model under a dedicated submodule, exposing constructors, state calculation, and interpolation type retrieval.
  - Introduced a method to get the interpolation type used by the model.
  - Added a method to retrieve coordinate subset indices for improved subset management.

- **Bug Fixes**
  - Improved error handling for state calculations requested outside the valid time interval.

- **Tests**
  - Expanded C++ and Python test suites to cover new constructors, interpolation types, boundary conditions, and error scenarios.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Model/Tabulated.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Model/Tabulated.cpp
@@ -130,7 +130,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Model_Tabulated(pybind11:
 
         .def(
             "calculate_states_at",
-            &Tabulated::calculateStateAt,
+            &Tabulated::calculateStatesAt,
             R"doc(
                 Calculate the states of the model at the specified instants.
 
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Model/Tabulated.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Model/Tabulated.cpp
@@ -1,8 +1,8 @@
 /// Apache License 2.0
 
-#include <OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp>
 
-inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile_Model_Tabulated(pybind11::module& aModule)
+inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Model_Tabulated(pybind11::module& aModule)
 {
     using namespace pybind11;
 
@@ -10,29 +10,18 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile_Model_Tabulated(pybin
 
     using ostk::mathematics::curvefitting::Interpolator;
 
-    using ostk::astrodynamics::flight::profile::Model;
-    using ostk::astrodynamics::flight::profile::model::Tabulated;
+    using ostk::astrodynamics::trajectory::Model;
+    using ostk::astrodynamics::trajectory::model::Tabulated;
     using ostk::astrodynamics::trajectory::State;
 
     class_<Tabulated, Model>(
         aModule,
         "Tabulated",
         R"doc(
-            A flight profile model defined by a set of states.
+            A trajectory model defined by a set of states.
 
         )doc"
     )
-
-        .def(
-            init<Array<State>&>(),
-            R"doc(
-                Constructor.
-
-                Args:
-                    states (Array[State]): The states of the model.
-             )doc",
-            arg("states")
-        )
 
         .def(
             init<const Array<State>&, const Interpolator::Type&>(),
@@ -92,6 +81,39 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile_Model_Tabulated(pybin
         )
 
         .def(
+            "get_interpolation_type",
+            &Tabulated::getInterpolationType,
+            R"doc(
+                Get the interpolation type of the model.
+
+                Returns:
+                    Interpolator.Type: The interpolation type of the model.
+             )doc"
+        )
+
+        .def(
+            "get_first_state",
+            &Tabulated::getFirstState,
+            R"doc(
+                Get the first state of the model.
+
+                Returns:
+                    State: The first state of the model.
+             )doc"
+        )
+
+        .def(
+            "get_last_state",
+            &Tabulated::getLastState,
+            R"doc(
+                Get the last state of the model.
+
+                Returns:
+                    State: The last state of the model.
+             )doc"
+        )
+
+        .def(
             "calculate_state_at",
             &Tabulated::calculateStateAt,
             R"doc(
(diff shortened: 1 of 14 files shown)


src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -94,53 +110,73 @@ State Tabulated::calculateStateAt(const Instant& anInstant) const
...
+            interval.accessStart().toString(),
+            interval.accessEnd().toString()
+        ));
     }
-    else if (stateRange.first != nullptr)
+
+    VectorXd coordinates(stateBuilder_.getSize());
+
+    // Interpolate the coordinates except for the attitude quaternion
+    Size k = 0;
```
alex-liang3, line 128: (sorry, the deleted line in the diff means I can't give a code suggestion) 

- I don't think this logic is very trivial, so I think i'd rather have more verbose names for `i`, `j`, `k`
- Why have both `k` and `j`? Aren't they incrementing in the same way?

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -255,98 +290,62 @@ bool Tabulated::operator!=(const Model& aModel) const
...
-        Unpack(previousStatePtr, nextStatePtr) = this->accessStateRangeAtIndex(stateIndex_);  // Check index cache
+        return;
+    }
+
+    const State& firstState = aStateArray.accessFirst();
+
+    Array<Shared<const CoordinateSubset>> coordinateSubsets = firstState.getCoordinateSubsets();
+
+    // ensure that the last coordinate subset is the attitude quaternion
+    coordinateSubsets.remove(AttitudeQuaternion::Default());
```
alex-liang3, line 305: is it safe to assume that these states have a quaternion CoordinateSubset to begin with?

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -255,98 +290,62 @@ bool Tabulated::operator!=(const Model& aModel) const
...
-            }
+            return lhs.getInstant() < rhs.getInstant();
         }
-        else if (previousStatePtr != nullptr)
+    );
+
+    stateArray_ = stateArray;
+
+    VectorXd timestamps(stateArray.getSize());
+    MatrixXd coordinates(stateArray.getSize(), firstState.getSize() - 4);  // Exclude quaternion
```
alex-liang3, line 324: could we instead do:
```suggestion
    MatrixXd coordinates(stateArray.getSize(), firstState.getSize() - AttitudeQuaternion::Default().getSize());  // Exclude quaternion
```
?

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.hpp
```diff
@@ -31,54 +35,119 @@ using ostk::core::filesystem::File;
...
+    /// @brief Constructor
+    ///
+    /// @param aStateArray An array of states
+    /// @deprecated Use the constructor with Interpolator::Type instead
+    [[deprecated("Use the constructor with Interpolator::Type instead")]]
     Tabulated(const Array<State>& aStateArray);
 
+    /// @brief Constructor
+    ///
+    /// @param aStateArray An array of states
```
alex-liang3, line 65:
```suggestion
    /// @param aStateArray An array of states
    /// @param anInterpolatorType The type of interpolation used for all but the AttitudeQuaternion subset. Defaults to Linear.
```

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Model/Tabulated.cpp
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+        .def(
+            init<const Array<State>&, const Interpolator::Type&>(),
+            R"doc(
+                Constructor.
+
+                Args:
+                    states (Array[State]): The states of the model.
+                    interpolation_type (Interpolator.Type): The type of interpolation to use. Defaults to Linear.
```
alex-liang3, line 33: worth clarifying that this does not apply to attitude quaternions?

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -94,53 +110,73 @@ State Tabulated::calculateStateAt(const Instant& anInstant) const
...
+            interval.accessEnd().toString()
+        ));
     }
-    else if (stateRange.first != nullptr)
+
+    VectorXd coordinates(stateBuilder_.getSize());
+
+    // Interpolate the coordinates except for the attitude quaternion
+    Size k = 0;
+    for (Size i = 0; i < this->stateBuilder_.getCoordinateSubsets().getSize() - 1; ++i)
```
phc1990, line 129: I like your new approach better, just 1 comment.

I think this sequence might be, both more clear, and not rely so much on which subset might be present:

1. Make a "reduced builder" (with all the same coordinates as this->stateBuilder_, except Quaternion).
2. Loop and get "reduced" states by ReducedBuilder.reduce(original_state_array)
3. Loop and interpolate the "reduced" states 👉 you don't need to do it by subset, just interpolate the whole coordinates vector
4. You can re-use the Reduced Builder to create your newly interpolated states (still no quaternion in them)

Then You can do the same sequence to interpolate the attitude quaternion, where your reduced builder only contains the quaternion subset.

Then you can go back to the stateBuilder_ to get the final interpolated states:

stateBuilder_.expand(interpolatedStateWithNoQuaternion, interpolatedStateOnlyWithQuaternion)

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -94,53 +110,73 @@ State Tabulated::calculateStateAt(const Instant& anInstant) const
...
+        const Quaternion qBefore = stateBefore.getAttitude();
+        const Quaternion qAfter = stateAfter.getAttitude();
+
+        const double ratio = (anInstant - stateBefore.accessInstant()).inSeconds() /
+                             (stateAfter.accessInstant() - stateBefore.accessInstant()).inSeconds();
+        qAtInstant = Quaternion::SLERP(qBefore, qAfter, ratio);
     }
 
-    throw ostk::core::error::RuntimeError("Cannot calculate state at [{}].", anInstant.toString());
+    coordinates.tail<4>() = qAtInstant.toNormalized().toVector(Quaternion::Format::XYZS);
```
phc1990, line 177: So that we don't have to rely on quaternion being at the end of the subsets in the future

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
vishwa2710: Yeah for a Profile I think that is a fair assumption, it will error out at many points before it gets here otherwise.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.hpp
vishwa2710: Unfortunately I can't add a default here, because it would then be identical to the existing constructor, and I want to avoid a breaking change for now.

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -20,24 +23,34 @@ namespace profile
...
-using ostk::physics::coordinate::Velocity;
-using ostk::physics::time::Duration;
+
+using ostk::astrodynamics::trajectory::state::CoordinateSubset;
+using ostk::astrodynamics::trajectory::state::coordinatesubset::AttitudeQuaternion;
+
+Tabulated::Tabulated(const Array<State>& aStateArray, const Interpolator::Type& anInterpolatorType)
+    : Model(),
+      stateArray_(aStateArray),
+      stateBuilder_(StateBuilder::Undefined())
```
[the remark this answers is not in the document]
vishwa2710: Not in the case the state array is empty, therefore it's done after verifying there are atleast 2 states.

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.hpp
alex-liang3: ah sorry I'm not sure why I said `Defaults to Linear` - maybe it was meant to be for the deprecated constructor. I more just meant to add the docs for the Interpolator arg in this signature

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
vishwa2710: Great suggestion! I modified it a bit but it should be much cleaner now!

alex-liang3 (approved):

apaletta3 (commented):
:100:

PR #559, opened 2025-05-19
feat: distinguish between profile target velocity and ground velocity

apaletta3:
Non-breaking change that moves the yaw compensation functionality over to `TargetGroundVelocity`, while the leaving the `TargetVelocity` to be used for non-yaw compensated profiles. This more clearly indicates the intent of each target type, since yaw compensation doesn't align the clocking axis with the target's general velocity, it aligns it with the target's ground velocity.

Additionally, this corrects the computation of the `ComputeTargetGroundVelocityVector` method - because before it was projecting the target's ground velocity onto the plane that was normal to the target's relative direction, but this isn't needed due to how the clocking/alignement logic works. Further, if a user wanted to Align with the TargetGroundVelocity, previously they would have gotten the wrong result.

Things left to do:
- Modify python bindings accordingly
- Add more unit test cases for the yaw compensation and not yaw compensated profiles - including simple geometry cases where we can back out the expected clocking vectors from 2D geometry

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
@@ -90,18 +90,19 @@ class Profile
 
     enum class TargetType
     {
-        GeocentricNadir,     /// Negative of the position vector of the satellite in the ECI frame
-        GeodeticNadir,       /// Negative of the geodetic normal of the satellite in the ECI frame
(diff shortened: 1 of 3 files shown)


apaletta3:
In favor of this [PR](https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/561)

apaletta3:
https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/561

PR #563, opened 2025-05-20
fix: tabulated profile implementation

apaletta3:
Small fix that:
1. Removes an extra copy of the stateArray
2. throw an error when constructing a Tabulated profile with less than 2 states, which no longer allows construction of undefined tabulated models

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling by providing clear feedback when an insufficient number of states is provided.
- **Refactor**
	- Streamlined internal handling of state arrays for improved efficiency.
	- Simplified test setup and removed tests related to undefined or partially constructed profiles.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
@@ -308,7 +308,7 @@ void Tabulated::setMembers(const Array<State>& aStateArray)
 {
     if (aStateArray.getSize() < 2)
     {
-        throw ostk::core::error::RuntimeError("State array must have at least length 2.")
+        throw ostk::core::error::RuntimeError("State array must have at least length 2.");
     }
 
     const State& firstState = aStateArray.accessFirst();
--- a/src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
@@ -39,7 +39,6 @@ using ostk::astrodynamics::trajectory::state::coordinatesubset::AttitudeQuaterni
 
 Tabulated::Tabulated(const Array<State>& aStateArray, const Interpolator::Type& anInterpolatorType)
     : Model(),
-      stateArray_(aStateArray),
       interpolatorType_(anInterpolatorType),
       stateBuilder_(StateBuilder::Undefined()),
       reducedStateBuilder_(StateBuilder::Undefined())
@@ -49,7 +48,6 @@ Tabulated::Tabulated(const Array<State>& aStateArray, const Interpolator::Type& 
 
 Tabulated::Tabulated(const Array<State>& aStateArray)
     : Model(),
-      stateArray_(aStateArray),
       interpolatorType_(Interpolator::Type::Linear),
       stateBuilder_(StateBuilder::Undefined()),
       reducedStateBuilder_(StateBuilder::Undefined())
@@ -310,7 +308,7 @@ void Tabulated::setMembers(const Array<State>& aStateArray)
 {
     if (aStateArray.getSize() < 2)
     {
-        return;
+        throw ostk::core::error::RuntimeError("State array must have at least length 2.")
     }
 
     const State& firstState = aStateArray.accessFirst();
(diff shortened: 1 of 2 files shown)


src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -310,7 +308,7 @@ void Tabulated::setMembers(const Array<State>& aStateArray)
 {
     if (aStateArray.getSize() < 2)
     {
-        return;
+        throw ostk::core::error::RuntimeError("State array must have at least length 2.")
```
apaletta3, line 311:
```suggestion
        throw ostk::core::error::RuntimeError("State array must have at least length 2.");
```

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -310,7 +308,7 @@ void Tabulated::setMembers(const Array<State>& aStateArray)
 {
     if (aStateArray.getSize() < 2)
     {
-        return;
+        throw ostk::core::error::RuntimeError("State array must have at least length 2.");
```
vishwa2710, line 311: This prevents us from creating an undefined tabulated model.

vishwa2710 (approved):
I don't think this will pass tests. The early return was intentional as if there are less than 2 states there's nothing to sort, interpolate or save. We just error out early when calculating states, but we can still create an undefined profile for comparisons.

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
apaletta3: Yeah that was my intention.

I don't really see why we want to keep creating things that are undefined, IMO it just creates more confusion.

The only use case I see for creating things that are undefined is to allow passing of a variable into a callback of some sort, but you can just as easily do it by initializing the variable and passing it

test/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Tabulated.test.cpp
```diff
@@ -82,25 +67,13 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Flight_Profile_Models_Tabulated, EqualToOp
     {
         EXPECT_TRUE(tabulated_ == tabulated_);
     }
-
-    {
-        const Array<State> states = {states_[0]};
-        const Tabulated reducedTabulated = {states};
-        EXPECT_FALSE(tabulated_ == reducedTabulated);
```
alex-liang3, line 89: I think you should create another tabulated with 2 different states, or a 3rd state alongside `states_`, rather than outright removing this test

Same goes for the test below

apaletta3: Fair.
Broader question though, do we actually think it makes to == trajectories?

alex-liang3: ~trajectories~ tabulated flight profile models; I don't think it hurts, and also it's an abstract method in the (flight profile) `Model` base class so it'd be pretty hairy to undo

vishwa2710: Yeah I think if we want to refactor and remove operator == checks that's a separate discussion. I don't think we should be removing tests. Same for below.

apaletta3: yup that's fair, separate discussion, will re-add the missing test coverage

test/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Tabulated.test.cpp
```diff
@@ -95,12 +79,6 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Flight_Profile_Models_Tabulated, NotEqualT
     {
         EXPECT_FALSE(tabulated_ != tabulated_);
     }
-
-    {
-        const Array<State> states = {states_[0]};
-        const Tabulated reducedTabulated = {states};
-        EXPECT_TRUE(tabulated_ != reducedTabulated);
```
alex-liang3, line 102: don't forget to readd this test too pls 🙏

alex-liang3 (changes requested):

PR #565, opened 2025-05-30
feat: extract maneuvers for custom dynamics

vishwa2710:
This allows us to extract maneuvers from segments where the maneuver was a tabulated dynamic. This is rare, but there can be situations where we would want to re-solve the maneuver with a fixed attitude for example.


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -136,25 +138,8 @@ Array<flightManeuver> Segment::Solution::extractManeuvers(const Shared<const Fra
         return {};
     }
 
-    // Loop through dynamics to find Thruster dynamics
```
apaletta3, line 139: I think its worth keeping this logic but generalizing it to any dynamics from the provided argument

apaletta3 (commented):
Nice! I think its worth marking the old method as depracated and redirecting directly to the new method if you accept my suggestion

PR #567, opened 2025-06-14
fix: lazy evaluation for temporal condition solver

vishwa2710:
using std::all_of is quite clean, but requires evaluation of every single condition which can be expensive. Instead we can return early if any of the conditions return False


lucas-bremond:
Nice! Why not use std::any_of?

vishwa2710:
> ~Nice! Why not use std::any_of?~
> 
> Actually, @vishwa2710
> 
> > using std::all_of is quite clean, but requires evaluation of every single condition which can be expensive
> 
> I do believe `std::all_of` ([definition](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of.html)) returns as soon as it finds a falsy element. Am I missing an additional rationale?

Yeah you're right, upon reading further that is indeed the case. Cursor AI let me down when I asked it :'(

PR #571, opened 2025-07-15
feat: add COE getCartesianState frame check

apaletta3:
Doesn't make sense to be able to create a set of COEs in a non-quasi inertial frame

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for undefined states and non-quasi-inertial frames in orbital element to Cartesian state conversions.
  * Enhanced error messages for frame validation to provide clearer and more consistent feedback.

* **Tests**
  * Added and updated tests to verify runtime errors are correctly raised when non-quasi-inertial frames are used in orbital state conversions.


test/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/ModifiedEquinoctial.test.cpp
```diff
@@ -100,7 +100,7 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_Kepler_ModifiedEqui
     {
```
apaletta3, line 100:
```suggestion
    // Non-quasi inertial Frame
    {
```

alex-liang3 (approved):

PR #573, opened 2025-07-18
feat: add eclipse generator

phc1990:


bindings/python/test/eclipse/test_generator.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+
+@pytest.fixture
+def analysis_interval() -> Interval:
+    return Interval.closed(
+        Instant.date_time(DateTime(2000, 1, 1, 14, 0, 0), Scale.UTC),
+        Instant.date_time(DateTime(2000, 1, 1, 17, 0, 0), Scale.UTC),
+    )
+
```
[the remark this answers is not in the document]
phc1990: It's a LEO equatorial orbit, it will contain events

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Eclipse/Generator.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    using ostk::physics::Environment;
+    using ostk::physics::environment::utilities::Eclipse;
+    using ostk::physics::time::Duration;
+    using ostk::physics::time::Interval;
+
+    using ostk::astrodynamics::eclipse::Generator;
+    using ostk::astrodynamics::Trajectory;
+
+    class_<Generator, Shared<Generator>>(
```
vishwa2710, line 21:
```suggestion
    class_<Generator>(
```
Don't think you need to have the Shared<Generator> here as it's a base class and not inherited, not used as a shared pointer across our repo

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Eclipse/Generator.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+        .def(
+            init<const Environment&, const Duration&, const Duration&>(),
+            R"doc(
+                Constructor.
+
+                Args:
+                    environment (Environment): The Environment to use during the search. Defaults to the default environment.
+                    search_step_size (Duration): The step size to use during the search. Defaults to 60 seconds.
+                    search_tolerance (Duration): The tolerance to use during the search. Defaults to 1 millisecond.
```
vishwa2710, line 37:
```suggestion
                    environment (Environment): The Environment to use during the search. Defaults to the Environment.default().
                    search_step_size (Duration): The step size to use during the search. Defaults to Duration.seconds(60.0).
                    search_tolerance (Duration): The tolerance to use during the search. Defaults to Duration.milliseconds(1.0).
```
Just to be consistent.

vishwa2710 (approved):

PR #580, opened 2025-08-06
fix: access generator stepping outside analysis interval

vishwa2710:
Fixes a bug where if a solved coarse access interval is between the analysis interval boundary and the last step, then it steps outside the analysis interval.

For example, if the analysis interval ends at 13:30:25, and the step size is 1 minute, and the analysis interval ends at 13:30:00,
then if the last step satisfies the access condition at 13:29:25, it will try and check for the step beyond by 1 minute to see if it can find a precise crossing, even though it is outside the analysis interval.

To fix this, we only solve for the crossing if within bounds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of access interval boundaries to ensure precise computation at the edges of analysis intervals.

* **Tests**
  * Added a new test to verify that interpolation does not occur outside the specified analysis interval, ensuring correct access detection at interval boundaries.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
@@ -839,7 +839,7 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
 
         Instant intervalStart = anAnalysisInterval.getStart();
 
-        // compute start crossing if it is not the start of the requested analysis interval
+        // Compute start crossing if both bounding instants are within the analysis interval
         if (lowerBoundPreviousInstant >= anAnalysisInterval.getStart())
         {
             const auto startCrossingDurationSeconds = rootSolver.solve(
@@ -854,12 +854,12 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
         }
 
         const Instant upperBoundInstant = interval.getEnd();
-        const Instant upperBoundNextInstant = interval.getEnd() + this->step_;
+        const Instant upperBoundNextInstant = std::min(interval.getEnd() + this->step_, anAnalysisInterval.getEnd());
 
         Instant intervalEnd = anAnalysisInterval.getEnd();
 
-        // compute end crossing if it is not the end of the requested analysis interval
-        if (upperBoundNextInstant <= anAnalysisInterval.getEnd())
+        // Compute end crossing if both bounding instants are within the analysis interval
+        if (upperBoundNextInstant <= anAnalysisInterval.getEnd() && upperBoundNextInstant != upperBoundInstant)
         {
             const auto endCrossingDurationSeconds = rootSolver.solve(
                 [&upperBoundInstant, &condition](double aDurationInSeconds) -> double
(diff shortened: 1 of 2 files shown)


src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -840,7 +840,7 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
         Instant intervalStart = anAnalysisInterval.getStart();
 
         // compute start crossing if it is not the start of the requested analysis interval
```
apaletta3, line 842:
```suggestion
        // Compute start crossing if it minus one step is after the analysis interval lower bound
```

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -859,7 +859,7 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
         Instant intervalEnd = anAnalysisInterval.getEnd();
 
         // compute end crossing if it is not the end of the requested analysis interval
```
apaletta3, line 861:
```suggestion
        // Compute end crossing if it plus one step is before the analysis interval upper bound
```

test/OpenSpaceToolkit/Astrodynamics/Access/Generator.test.cpp
```diff
@@ -1103,6 +1105,68 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, ComputeAccesses_5)
     }
 }
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, ComputeAccesses_6)
+{
+    // Ensure no interpolation is performed outside of the analysis interval
+
+    {
+        const Instant instant = Instant::J2000();
```
apaletta3, line 1113: Nit: clearer for the reader of the unit test to either put this instant in isoformat, or put the below `interval` instants in `instant + Duration` and `instant - Duration` format

test/OpenSpaceToolkit/Astrodynamics/Access/Generator.test.cpp
```diff
@@ -1103,6 +1105,68 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, ComputeAccesses_5)
     }
 }
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, ComputeAccesses_6)
+{
+    // Ensure no interpolation is performed outside of the analysis interval
+
+    {
```
apaletta3, line 1112: This test only tests exercise the upper bound code path? Or does it exercise both the upper and lower?

If its not much work (idk if you have to "know" what the accesses look like ahead of time to trigger the lower bound condition) it would be good to exercise the other code path as well

apaletta3 (approved):

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
vishwa2710: Will update it but I think we can make it a bit clearer.

test/OpenSpaceToolkit/Astrodynamics/Access/Generator.test.cpp
vishwa2710: You want me to instantiate the J2000 epoch as an isoformat for the orbit epoch? It's not even used relative to the access interval, not sure why it matters in this case? But I advocate against instantiating the J2000 as ISOformat for sure, will lead to significant human error.

test/OpenSpaceToolkit/Astrodynamics/Access/Generator.test.cpp
vishwa2710: Hashed this out in person and we agree that only the upper bound edge case is hit.

PR #583, opened 2025-08-22
fix: don't overshoot access gaps at end of analysis span

alex-liang3:
At the end of the analysis span, when an access interval can be less than one "step" long, we risk overshooting/skipping-over an access gap if we directly take one step back. This can lead to a case where, when doing the "fine" solve, both endpoints appear to be in-access, leading the root solver (bracketing) to fail. See test case for representative example.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved access interval boundary handling to avoid missed gaps or overshoots between consecutive accesses and at analysis window edges (e.g., when passing through overhead blind spots). Users will see more accurate start/end times for partial accesses.

- **Tests**
  - Added a test validating blind-spot and boundary alignment, ensuring accesses end precisely at the analysis interval when applicable.


phc1990 (approved):
Amazing.

apaletta3 (approved):
Nice!

PR #589, opened 2025-10-06
feat: add constructBodyFrame method to Profile

vishwa2710:
Renaming `getBodyFrame` to `constructBodyFrame` as the body frame is not a member variable of the profile class. Also adding an overwrite flag for convenience.

## Summary by CodeRabbit

* **New Features**
  * Added ability to construct a body frame from a profile with an optional overwrite flag (defaults to false) to replace an existing frame.

* **Deprecations**
  * get_body_frame/getBodyFrame now emits a deprecation warning; use construct_body_frame/constructBodyFrame instead.

* **Tests**
  * Added tests for construction, name-collision errors, overwrite behavior, undefined-profile errors, and frame persistence after overwrite.

### DIFF
--- a/test/OpenSpaceToolkit/Astrodynamics/Flight/Profile.test.cpp
+++ b/test/OpenSpaceToolkit/Astrodynamics/Flight/Profile.test.cpp
@@ -272,22 +272,23 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Flight_Profile, ConstructBodyFrame)
     {
         const String frameName = "test_construct_body_frame";
 
-        // Test basic construction
+        // basic construction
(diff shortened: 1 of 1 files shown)


bindings/python/test/flight/test_profile.py
apaletta3, line 1: nit: we don't usually test the full functionality in python tests (just that the bindings bind properly) - any reason why you did so here?

apaletta3 (approved):

bindings/python/test/flight/test_profile.py
vishwa2710: I can get rid of the pytest raises, but definitely wanted to test with args. Will update.

alex-liang3 (approved):

PR #595, opened 2025-10-31
feat!: solve Sequences considering maneuvering constraints

phc1990:
!Breaking change!
- Sequence no longer accepts minimum maneuver duration as an input

## What

Add Segment-level Constraints:
- minimum maneuver duration
- minimum maneuver separation
- maximum maneuver duration (with different strategies)

Additionally:
- adds utility Sequence Solution level `extractManeuvers`

## How

A Segment always solves until an event condition is reached, or maximum propagation duration is reached. For coasting segments this is trivial. For maneuvering segments, in the case that maneuvering constraints are provided, maneuvers are solved 1 by 1, skipping short maneuvers, coasting to ensure minimum separation constraints are met, and following the appropriate maximum maneuver duration violation strategy.

### DIFF
--- a/bindings/python/test/trajectory/test_segment.py
+++ b/bindings/python/test/trajectory/test_segment.py
@@ -378,14 +378,14 @@ class TestSegment:
     ):
         assert len(coast_duration_segment.get_dynamics()) == len(dynamics)
 
-    def test_get_coast_dynamics(
+    def test_get_free_dynamics(
         self,
         dynamics: list,
         coast_duration_segment: Segment,
         maneuver_segment: Segment,
     ):
-        assert len(coast_duration_segment.get_coast_dynamics()) == len(dynamics)
-        assert len(maneuver_segment.get_coast_dynamics()) == len(dynamics)
+        assert len(coast_duration_segment.get_free_dynamics()) == len(dynamics)
+        assert len(maneuver_segment.get_free_dynamics()) == len(dynamics)
 
     def test_get_numerical_solver(
         self,


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -219,13 +232,35 @@ class Segment
...
+    ///
+    /// @param aName Optional name for the new segment. If not provided, uses the current segment's name
+    /// @return A new coast segment
+    Segment toCoastSegment(const String& aName = String::Empty()) const;
+
+    /// @brief Build a maneuver segment from the current instance.
+    ///
+    /// @param aThrusterDynamics The thruster dynamics for the new maneuver segment
+    /// @param aName Optional name for the new segment. If not provided, uses the current segment's name
+    /// @return A new maneuver segment
```
apaletta3: Nit
```suggestion
    /// @param aName (optional) name for the new segment. If not provided, uses the current segment's name
    /// @return A new coast segment
    Segment toCoastSegment(const String& aName = String::Empty()) const;

    /// @brief Build a maneuver segment from the current instance.
    ///
    /// @param aThrusterDynamics The thruster dynamics for the new maneuver segment
    /// @param aName (optional) name for the new segment. If not provided, uses the current segment's name
    /// @return A new maneuver segment
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.hpp
```diff
@@ -42,6 +46,17 @@ using ostk::astrodynamics::trajectory::State;
...
    public:
+    /// @brief If the maximum maneuver duration is specified, this strategy determines how to handle maneuvers that
+    /// exceed said constraint.
+    enum class MaximumManeuverDurationStrategy
+    {
+        Fail,   ///< The sequence will fail if a maneuver exceeds the maximum duration.
+        Skip,   ///< The maneuver will be skipped entirely.
+        Slice,  ///< The maneuver will be sliced into one or more maneuvers starting from the beginning and solving
+                ///< iteratively.
+        Center  ///< The maneuver will be shortened and centered around its midpoint.
```
apaletta3:
```suggestion
    enum class MaximumManeuverDurationViolationStrategy
    {
        Fail,   ///< The sequence will fail if a maneuver exceeds the maximum duration.
        Skip,   ///< The maneuver will be skipped entirely.
        Split,  ///< The maneuver will be split into one or more maneuvers that are each within the maximum duration, until the last maneuver which will be equal or shorter than the maximum duration.
        Center  ///< The maneuver will be shortened to the maximum duraiton and centered around its midpoint.
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -663,42 +830,36 @@ Segment::Solution Segment::Solve_(
...
-}
-
-const Shared<Thruster> Segment::FindThrusterDynamics(const Array<Shared<Dynamics>>& aDynamicsArray)
-{
-    // Loop through dynamics to find Thruster dynamics
-    Shared<Thruster> thrusterDynamics = nullptr;
-    bool thrusterDynamicsFound = false;
-    for (const Shared<Dynamics>& dynamic : aDynamicsArray)
+    // Determine the final conditionIsSatisfied value
+    bool finalConditionIsSatisfied = conditionSolution.conditionIsSatisfied;
```
apaletta3: I'm not sure I understand why you may need the condition re-evaluated? Can you leave a comment explaining what situation this happens in?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -93,6 +97,19 @@ Mass Sequence::Solution::computeDeltaMass() const
...
 }
 
+Array<Maneuver> Sequence::Solution::extractManeuvers(const Shared<const Frame>& aFrameSPtr) const
+{
+    Array<Maneuver> maneuvers = Array<Maneuver>::Empty();
+
+    for (const Segment::Solution& segmentSolution : this->segmentSolutions)
+    {
+        const Array<Maneuver> segmentManeuvers = segmentSolution.extractManeuvers(aFrameSPtr);
+        maneuvers.add(segmentManeuvers);
```
apaletta3: just make sure you end up with an `Array<Maneuver>` and not a `Array<Array<Maneuver>>`

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -191,18 +208,23 @@ Sequence::Sequence(
       numericalSolver_(aNumericalSolver),
       dynamics_(aDynamicsArray),
       segmentPropagationDurationLimit_(aMaximumPropagationDuration),
-      minimumManeuverDuration_(aMinimumManeuverDuration)
+      minimumManeuverDuration_(Duration::Undefined()),
+      minimumManeuverSeparation_(Duration::Seconds(30.0)),
```
apaletta3: is there a way to avoid putting in a default value here and make it undefined?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -307,25 +413,16 @@ Sequence::Solution Sequence::solve(const State& aState, const Size& aRepetitionC
 
             BOOST_LOG_TRIVIAL(debug) << "Solving Segment:\n" << segment << std::endl;
 
-            Segment::Solution segmentSolution = segment.solve(initialState, segmentPropagationDurationLimit_);
+            auto [segmentSolution, newLastManeuverInterval] =
+                solveSegment_(initialState, segment, segmentPropagationDurationLimit_, lastManeuverInterval);
```
apaletta3:
```suggestion
                this->solveSegment_(initialState, segment, segmentPropagationDurationLimit_, lastManeuverInterval);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+) const
+{
+    const Shared<Thruster> thrusterDynamics = aSegment.getThrusterDynamics();
+    const Shared<const GuidanceLaw> originalGuidanceLaw = thrusterDynamics->getGuidanceLaw();
+    const Shared<HeterogeneousGuidanceLaw> heterogeneousGuidanceLaw = std::make_shared<HeterogeneousGuidanceLaw>();
+
+    for (const Interval& maneuverInterval : aManeuverIntervals)
+    {
+        heterogeneousGuidanceLaw->addGuidanceLaw(
+            std::const_pointer_cast<GuidanceLaw>(originalGuidanceLaw), maneuverInterval
```
apaletta3: does the pointer need to be cast here?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+                    segmentToSolve = aSegment.toManeuverSegment(
+                        shortenedManeuverThruster, aSegment.getName() + " (Sliced Maneuver)"
+                    );
+                    coastToInstant = candidateManeuver.getInterval().getStart() + maximumManeuverDuration_;
+                    break;
+                }
+
+                case MaximumManeuverDurationStrategy::Center:
+                {
+                    const Shared<Thruster> shortenedManeuverThruster = buildThrusterDynamicsWithManeuverIntervals_(
```
apaletta3:
```suggestion
                    const Shared<Thruster> shortenedManeuverThruster = this->buildThrusterDynamicsWithManeuverIntervals_(
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+        maneuverIntervals.add(candidateManeuver.getInterval());
+        heterogeneousGuidanceLaw->addGuidanceLaw(
+            std::const_pointer_cast<GuidanceLaw>(maneuverSegmentSolution.getThrusterDynamics()->getGuidanceLaw()),
+            candidateManeuver.getInterval()
+        );
+
+        if (conditionIsSatisfied)
+        {
+            break;
+        }
```
apaletta3: It seems that this statement is placed incorrectly in the code? It should go right after somewhere that changes its value, like on line 565

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -199,10 +204,18 @@ class Segment
     /// @return Dynamics
     Array<Shared<Dynamics>> getDynamics() const;
 
+    /// @brief Get coast dynamics array
+    /// @return Coast dynamics array
+    Array<Shared<Dynamics>> getCoastDynamics() const;
```
vishwa2710: I think it's misleading to call these `coastDynamics`. They're dynamics that are applied during a maneuver as well.
If you want to segment it in this way, I would maybe call it `generalDynamics` or `commonDynamics` or `freeDynamics` and `thrusterDynamics` or something like that.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -199,10 +204,18 @@ class Segment
...
+
     /// @brief Get numerical solver
     /// @return Numerical solver
     NumericalSolver getNumericalSolver() const;
 
+    /// @brief Get thruster dynamics
+    /// @return Thruster dynamics
+    Shared<Thruster> getThrusterDynamics() const;
```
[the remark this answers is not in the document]
vishwa2710: Nit, but the class is called Dynamics, I think the singular version of dynamics is still dynamics.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+            const Array<State>& segmentStates =
+                Array<State>(coastSegmentSolution.states.begin() + 1, coastSegmentSolution.states.end());
+            states.add(segmentStates);
+
+            if (conditionIsSatisfied)
+            {
+                break;
+            }
+
+            segmentToSolve = aSegment;
```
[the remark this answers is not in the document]
phc1990: Variables renamed ✅ 

> Or its fine cause you always start from a newly added state and so should always make progress towards the end of the segment?

Correct, you always make progress as the initial state used it's always the last one

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+                            candidateManeuver.getInterval().getStart() + maximumManeuverDuration_
+                        )},
+                        "(Sliced Maneuver)"
+                    );
+
+                    segmentToSolve = aSegment.toManeuverSegment(
+                        shortenedManeuverThruster, aSegment.getName() + " (Sliced Maneuver)"
+                    );
+                    coastToInstant = candidateManeuver.getInterval().getStart() + maximumManeuverDuration_;
+                    break;
```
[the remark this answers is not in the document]
phc1990: this break is at Switch level, not outer while loop

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+                            maximumManeuverDuration_,
+                            Interval::Type::Closed
+                        )},
+                        "(Center Maneuver)"
+                    );
+                    segmentToSolve = aSegment.toManeuverSegment(
+                        shortenedManeuverThruster, aSegment.getName() + " (Centered Maneuver)"
+                    );
+                    coastToInstant = candidateManeuver.getInterval().getEnd();
+                    break;
```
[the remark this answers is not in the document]
phc1990: this break is at Switch level, not outer while loop

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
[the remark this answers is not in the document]
phc1990: I see the rationale, and I had the same doubt, but it looks like we're using Dynamics form for both singular or plural.

(the actual class is already called `Dynamics` as opposed to `Dynamic`)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -622,34 +705,118 @@ Segment Segment::ConstantLocalOrbitalFrameDirectionManeuver(
...
+            const Vector3d velocityCoordinates =
+                state.getVelocity().inUnit(Velocity::Unit::MeterPerSecond).accessCoordinates();
+            const Vector3d thrustAcceleration = guidanceLaw->calculateThrustAccelerationAt(
+                state.accessInstant(), positionCoordinates, velocityCoordinates, 1.0, state.accessFrame()
+            );
+            return thrustAcceleration.norm();
+        };
+
+        const Shared<RealCondition> thrustOffCondition = std::make_shared<RealCondition>(
+            "Thrust Off Condition", RealCondition::Criterion::NegativeCrossing, thrustAccelerationNormEvaluator, 0.5
```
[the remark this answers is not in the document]
vishwa2710: Adding a comment to address this.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.cpp
```diff
@@ -418,6 +503,231 @@ Sequence::Solution Sequence::solveToCondition(
...
+                Array<State>(maneuverSegmentSolution.states.begin() + 1, maneuverSegmentSolution.states.end());
+            states.add(segmentStates);
+
+            break;
+        }
+
+        if (maneuverSegmentSolutionManeuvers.getSize() > 1)
+        {
+            throw ostk::core::error::RuntimeError(
+                "More than one maneuver found when solving maneuver segment with constraints."
```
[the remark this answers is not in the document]
vishwa2710: I believe it is the latter. Just validating the contract that each sub-segment can only have 1 maneuver.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: Added a comment.

vishwa2710 (approved):

PR #600, opened 2025-11-11
refactor!: move core logic to segment

vishwa2710:
Moves the core sub-segment solving logic to the Segment class, changing the interface to `solve` and `solveWithConstraints`.
Removes the need for a lot of the exposed methods like toCoastSegment, toManeuverSegment etc. and simplifies the interface IMO.
Can probably be further improved, let's discuss this a bit.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
@@ -48,13 +48,16 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
         R"doc(
             Strategy to use when a maneuver exceeds the maximum duration constraint.
 
-            
-            Proposed maneuver:                                                                         [--------------------|------------]
-            Fail: Will throw an exception.
-            Skip: Will skip the maneuver.
-            LeadingSlice: Will shorten the maneuver to the maximum duration, sliced from the start.    [--------------------]
-            TrailingSlice: Will shorten the maneuver to the maximum duration, sliced from the end.                  [--------------------]
-            Center: Will take the centered part of the maneuver.                                              [--------------------]
+            Fail: Will throw a RuntimeError if a maneuver exceeds the maximum duration.
+            Skip: The maneuver will be skipped entirely.
+            TruncateEnd: The maneuver will be shortened to the maximum duration, truncating the end segment.
+            TruncateStart: The maneuver will be shortened to the maximum duration, truncating the start segment.
+            Center: The maneuver will be shortened to the maximum duration, truncating the edges, keeping the centered part of the maneuver.
+
+            Proposed maneuver: [--------------------|------------]
+            TruncateEnd:       [--------------------]
+            TruncateStart:                  [--------------------]
+            Center:                     [--------------------]
         )doc"
     )
 
@@ -67,14 +70,14 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
             "Skip", Segment::MaximumManeuverDurationViolationStrategy::Skip, "The maneuver will be skipped entirely."
         )
         .value(
-            "LeadingSlice",
-            Segment::MaximumManeuverDurationViolationStrategy::LeadingSlice,
-            "The maneuver will be shortened to the maximum duration, sliced from the start."
+            "TruncateEnd",
+            Segment::MaximumManeuverDurationViolationStrategy::TruncateEnd,
+            "Will shorten the maneuver to the maximum duration, truncating the end segment."
         )
         .value(
-            "TrailingSlice",
-            Segment::MaximumManeuverDurationViolationStrategy::TrailingSlice,
-            "The maneuver will be shortened to the maximum duration, sliced from the end."
+            "TruncateStart",
+            Segment::MaximumManeuverDurationViolationStrategy::TruncateStart,
+            "Will shorten the maneuver to the maximum duration, truncating the start segment."
         )
         .value(
             "Center",
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
@@ -65,13 +65,22 @@ class Segment
         Maneuver  ///< Maneuver
     };
 
+    /// @brief Maximum maneuver duration violation strategy
+    /// Strategy to use when a maneuver exceeds the maximum duration constraint.
+    ///
+    /// Example:
+    /// Proposed maneuver: [--------------------|------------]
+    /// TruncateEnd:       [--------------------]
+    /// TruncateStart:                  [--------------------]
+    /// Center:                     [--------------------]
     enum class MaximumManeuverDurationViolationStrategy
     {
-        Fail,           ///< The sequence will fail if a maneuver exceeds the maximum duration.
+        Fail,           ///< Will throw a RuntimeError if a maneuver exceeds the maximum duration.
         Skip,           ///< The maneuver will be skipped entirely.
-        LeadingSlice,   ///< The maneuver will be shortened to the maximum duration, sliced from the start.
-        TrailingSlice,  ///< The maneuver will be shortened to the maximum duration, sliced from the end.
-        Center          ///< The maneuver will be shortened to the maximum duration and centered around its midpoint.
+        TruncateEnd,    ///< The maneuver will be shortened to the maximum duration, truncating the end segment.
+        TruncateStart,  ///< The maneuver will be shortened to the maximum duration, truncating the start segment.
+        Center  ///< The maneuver will be shortened to the maximum duration, truncating the edges, keeping the centered
+                ///< part of the maneuver.
     };
 
     struct ManeuverConstraints
@@ -118,10 +127,6 @@ class Segment
         /// @param aManeuverConstraints A maneuver constraints
         /// @return An output stream
         friend std::ostream& operator<<(std::ostream& anOutputStream, const ManeuverConstraints& aManeuverConstraints);
-
-        String MaximumManeuverDurationViolationStrategyToString(
-            const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy
-        ) const;
     };
 
     /// @brief Once a segment is set up with an event condition, it can be solved, resulting in this segment's Solution.
@@ -370,6 +375,14 @@ class Segment
         const Shared<const LocalOrbitalFrameFactory>& aLocalOrbitalFrameFactory,
         const Angle& aMaximumAllowedAngularOffset = Angle::Undefined(),
         const ManeuverConstraints& aManeuverConstraints = ManeuverConstraints()
+    );
+
+    // TBI: Remove this and replace with magic_enum based solution
+    /// @brief String from maximum maneuver duration violation strategy
+    /// @param aMaximumDurationStrategy The maximum maneuver duration violation strategy
+    /// @return String from maximum maneuver duration violation strategy
+    static String StringFromMaximumManeuverDurationViolationStrategy(
+        const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy
     );
 
    private:
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
@@ -85,7 +85,8 @@ Segment::ManeuverConstraints::ManeuverConstraints(
     if (maximumDuration.isDefined() && !minimumSeparation.isDefined())
     {
         throw ostk::core::error::RuntimeError(
-            "Minimum separation of atleast 1 second must be defined if maximum duration is defined. This is to prevent "
+            "Minimum separation of at least 1 second must be defined if maximum duration is defined. This is to "
+            "prevent "
             "aliasing issues which can cause sequential maneuver intervals to overlap by a nanosecond."
         );
     }
@@ -132,7 +133,7 @@ void Segment::ManeuverConstraints::print(std::ostream& anOutputStream, bool disp
         << (this->minimumSeparation.isDefined() ? this->minimumSeparation.toString() : "Undefined");
     ostk::core::utils::Print::Line(anOutputStream)
         << "Maximum Duration Strategy:"
-        << MaximumManeuverDurationViolationStrategyToString(this->maximumDurationStrategy);
+        << Segment::StringFromMaximumManeuverDurationViolationStrategy(this->maximumDurationStrategy);
 
     if (displayDecorator)
     {
@@ -146,9 +147,9 @@ std::ostream& operator<<(std::ostream& anOutputStream, const Segment::ManeuverCo
     return anOutputStream;
 }
 
-String Segment::ManeuverConstraints::MaximumManeuverDurationViolationStrategyToString(
+String Segment::StringFromMaximumManeuverDurationViolationStrategy(
     const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy
-) const
+)
 {
     switch (aMaximumDurationStrategy)
     {
@@ -156,8 +157,10 @@ String Segment::ManeuverConstraints::MaximumManeuverDurationViolationStrategyToS
             return "Fail";
         case MaximumManeuverDurationViolationStrategy::Skip:
             return "Skip";
-        case MaximumManeuverDurationViolationStrategy::LeadingSlice:
-            return "Slice";
+        case MaximumManeuverDurationViolationStrategy::TruncateEnd:
+            return "TruncateEnd";
+        case MaximumManeuverDurationViolationStrategy::TruncateStart:
+            return "TruncateStart";
         case MaximumManeuverDurationViolationStrategy::Center:
             return "Center";
         default:
@@ -726,11 +729,11 @@ Segment::Solution Segment::solve(
 
     Array<State> segmentStates = {aState};
 
-    // Helper lambda to build a thruster dynamics with maneuver intervals
+    // Helper lambda to build a thruster dynamics that only thrusts within the given interval
     const Shared<Thruster> segmentThrusterDynamics = this->getThrusterDynamics();
     const Shared<const GuidanceLaw> originalGuidanceLaw = segmentThrusterDynamics->getGuidanceLaw();
 
-    const auto buildThrusterDynamicsWithManeuverIntervals =
+    const auto buildThrusterDynamicsWithinInterval =
         [&originalGuidanceLaw, &segmentThrusterDynamics](const Interval& aManeuverInterval) -> Shared<Thruster>
     {
         const Shared<HeterogeneousGuidanceLaw> heterogeneousGuidanceLaw = std::make_shared<HeterogeneousGuidanceLaw>();
@@ -745,7 +748,7 @@ Segment::Solution Segment::solve(
 
     // Helper lambda to solve a coast segment and update segmentStates
     const Instant maximumInstant = aState.accessInstant() + maximumPropagationDuration;
-    const auto solveAndAddCoastSegment = [&](const Instant& endInstant) -> bool
+    const auto solveAndAcceptCoast = [&](const Instant& endInstant) -> bool
     {
         const State& lastState = segmentStates.accessLast();
         const Segment::Solution coastSegmentSolution =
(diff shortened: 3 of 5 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -693,10 +966,11 @@ Segment Segment::ConstantLocalOrbitalFrameDirectionManeuver(
     const Array<Shared<Dynamics>>& aDynamicsArray,
     const NumericalSolver& aNumericalSolver,
     const Shared<const LocalOrbitalFrameFactory>& aLocalOrbitalFrameFactory,
-    const Angle& aMaximumAllowedAngularOffset
+    const Angle& aMaximumAllowedAngularOffset,
+    const ManeuverConstraints& aManeuverConstraints
 )
 {
-    return {
+    return Segment {
```
apaletta3: I'm not sure this works?
```suggestion
    return {
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+
+    while (segmentStates.accessLast().accessInstant() < maximumInstant)
+    {
+        // Coast
+        if (nextSubSegmentToSolve.getType() == Segment::Type::Coast)
+        {
+            const Segment::Solution coastSegmentSolution = nextSubSegmentToSolve.solve(
+                segmentStates.accessLast(),
+                std::min(nextInstantToCoastTo, maximumInstant) - segmentStates.accessLast().accessInstant()
+            );
```
apaletta3: I find it a little confusing to be calling the other "same level" public function from this one - for a second I thought it was recursive but it isn't, Might be better to have `solveWithConstraints` and `solve` never calling each other, but instead only calling the private `solve_` function
```suggestion
            const Segment::Solution coastSegmentSolution = nextSubSegmentToSolve.solve_(
                segmentStates.accessLast(),
                std::min(nextInstantToCoastTo, maximumInstant) - segmentStates.accessLast().accessInstant(),
                false,
            );
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -65,6 +65,34 @@ class Segment
         Maneuver  ///< Maneuver
     };
 
+    enum class MaximumManeuverDurationViolationStrategy
+    {
+        Fail,   ///< The sequence will fail if a maneuver exceeds the maximum duration.
+        Skip,   ///< The maneuver will be skipped entirely.
+        Slice,  ///< The maneuver will be split into one or more maneuvers that are each within the maximum duration,
+                ///< until the last maneuver which will be equal or shorter than the maximum duration.
+        Center  ///< The maneuver will be shortened to the maximum duraiton and centered around its midpoint.
```
alex-liang3:
```suggestion
        Center  ///< The maneuver will be shortened to the maximum duration and centered around its midpoint.
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+            subSegmentConditionIsSatisfied = coastSegmentSolution.conditionIsSatisfied;
+
+            segmentStates.add(Array<State>(coastSegmentSolution.states.begin() + 1, coastSegmentSolution.states.end()));
+
+            if (subSegmentConditionIsSatisfied)
+            {
+                break;
+            }
+
+            nextSubSegmentToSolve = *this;
```
apaletta3: Can you put a comment here explaining why we need to resolve this subsgment if the condition isn't satisfied? and why that doesn't lead to an infinite loop? Is it because segmentStates have been added compared to the last time this ran - so you have still made progress towards the end of the segment?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+        {
+            BOOST_LOG_TRIVIAL(debug
+            ) << "Maneuver duration is less than the minimum maneuver duration. Skipping the maneuver."
+              << std::endl;
+
+            nextSubSegmentToSolve =
+                buildCoastSegment("(Skipping Maneuver due to Minimum Maneuver Duration Constraint)");
+
+            nextInstantToCoastTo =
+                updateNextCoastInstant(candidateManeuverInterval.getEnd() + subSegmentManeuverMargin);
```
apaletta3: Is my understand correct that the `nextInstantToCoastTo` is actually more like a "next instant to coast to if the next subsegment to solve is a coast subsegment"? So its only used if the next subsegment turn out to be a subsegment that's either a coast segment from the start or a maneuver segment that violates a constraint and gets converted to coasting?

If both of those statements are correct, could you make that clearer with comments/variable naming?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+            continue;
+        }
+
+        // Maneuver has been accepted
+        BOOST_LOG_TRIVIAL(debug) << "Maneuver accepted." << std::endl;
+        segmentStates.add(
+            Array<State>(maneuverSubsegmentSolution.states.begin() + 1, maneuverSubsegmentSolution.states.end())
+        );
+        segmentManeuverIntervals.add(candidateManeuverInterval);
+        segmentGuidanceLaw->addGuidanceLaw(
```
apaletta3: Make its clearer
```suggestion
        segmentHeterogenousGuidanceLaw->addGuidanceLaw(
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -45,6 +49,33 @@ using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianAcceler
 using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianPosition;
 using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianVelocity;
 using ostk::astrodynamics::trajectory::StateBuilder;
+using flightManeuver = ostk::astrodynamics::flight::Maneuver;
+
+Segment::ManeuverConstraints::ManeuverConstraints()
```
alex-liang3: can't you just set the defaults in the declaration in the hpp, rather than having 2 definitions of the constructor here?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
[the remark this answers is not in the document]
vishwa2710: I gotta update the description for slice. Basically we have 4 options:
```
Original maneuver: x---------------------|-------x
FAIL: raise error
SKIP: []
SLICE:                      x---------------------x
CENTER:                         x---------------------x
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -65,6 +65,34 @@ class Segment
...
+            const Duration& aMinimumDuration,
+            const Duration& aMaximumDuration,
+            const Duration& aMinimumSeparation,
+            const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy
+        );
+
+        Duration minimumDuration;
+        Duration maximumDuration;
+        Duration minimumSeparation;
+        MaximumManeuverDurationViolationStrategy maximumDurationStrategy;
```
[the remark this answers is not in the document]
vishwa2710: Personally I find sorting by `Duration` -> `Separation` clearer than `Minimum` -> `Maximum`

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+    if (lastManeuverInterval.isDefined() && maneuverConstraints_.minimumSeparation.isDefined() &&
+        (aState.accessInstant() - lastManeuverInterval.getEnd()) < maneuverConstraints_.minimumSeparation)
+    {
+        nextInstantToCoastTo =
+            updateNextCoastInstant(lastManeuverInterval.getEnd() + maneuverConstraints_.minimumSeparation);
+        nextSubSegmentToSolve = buildCoastSegment("(Coast - Minimum Maneuver Separation Constraint)");
+    }
+
+    // TBI: is this necessary?
+    const Duration subSegmentManeuverMargin = Duration::Seconds(10.0);
```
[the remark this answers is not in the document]
vishwa2710: It applies a buffer after the maneuver ends before a new sub-segment begins. Not sure why Pau added it I'll check and see if it's truly necessary.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +622,269 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+        // Account for minimum maneuver separation
+        BOOST_LOG_TRIVIAL(debug) << "Accounting for minimum maneuver separation." << std::endl;
+        nextSubSegmentToSolve = buildCoastSegment("(Coast - Minimum Maneuver Separation Constraint)");
+
+        if (maneuverConstraints_.minimumSeparation.isDefined())
+        {
+            nextInstantToCoastTo = updateNextCoastInstant(
+                segmentManeuverIntervals.accessLast().getEnd() + maneuverConstraints_.minimumSeparation
+            );
+        }
```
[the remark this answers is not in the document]
vishwa2710: Good question, let's chat about this tomorrow in person, I have a few questions around this logic that I want to break down as well. I think we need to create a comprehensive set of cases that we want to test and make sure we're hitting all of them, as there's a lot of branches and Pau's existing tests might not cover everything.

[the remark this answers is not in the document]
vishwa2710: Test case ideas:
- with QLaw (a guidance law that can generate multiple maneuvers in a single "solve"
- with ConstantThrust (a guidance law that only fires on)

Then, we should do it with:
- minimum duration
- maximum duration
- minimum separation
- minimum duration, maximum duration
- minimum duration, minimum separation
- maximum duration, minimum separation
- minimum duration, maximum duration, minimum separation

Along with some event conditions like:
- eclipse
- total duration

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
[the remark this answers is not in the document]
vishwa2710: This is applied because if your solver decides to thrust off, you might not want to start IMMEDIATELY, you want like a short gap. I think we should remove this and just have someone define a minimum separation instead.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -45,6 +49,121 @@ using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianAcceler
...
+    switch (aMaximumDurationStrategy)
+    {
+        case MaximumManeuverDurationViolationStrategy::Fail:
+            return "Fail";
+        case MaximumManeuverDurationViolationStrategy::Skip:
+            return "Skip";
+        case MaximumManeuverDurationViolationStrategy::LeadingSlice:
+            return "Slice";
+        case MaximumManeuverDurationViolationStrategy::Center:
+            return "Center";
```
alex-liang3:
```suggestion
        case MaximumManeuverDurationViolationStrategy::LeadingSlice:
            return "Leading Slice";
        case MaximumManeuverDurationViolationStrategy::TrailingSlice:
            return "Trailing Slice";
        case MaximumManeuverDurationViolationStrategy::Center:
            return "Center";
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +710,233 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
-        constantManeuverDirectionLocalOrbitalFrameFactory_,
-        constantManeuverDirectionMaximumAllowedAngularOffset_,
+    if (!maneuverConstraints_.isDefined())
+    {
+        return solveManeuver_(aState, maximumPropagationDuration, this->getThrusterDynamics());
+    }
+
+    Array<State> segmentStates = {aState};
+
+    // Helper lambda to build a thruster dynamics with maneuver intervals
```
apaletta3:
```suggestion
    // Helper lambda to build a thruster dynamics that only thrusts within the given interval
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +710,233 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+
+        segmentStates.add(Array<State>(coastSegmentSolution.states.begin() + 1, coastSegmentSolution.states.end()));
+
+        segmentConditionIsSatisfied = coastSegmentSolution.conditionIsSatisfied;
+
+        return segmentConditionIsSatisfied;
+    };
+
+    // Helper lambda to solve a single maneuver and extract results
+    const auto solveManeuver = [&](const Shared<Thruster>& thrusterDynamics
```
apaletta3:
```suggestion
    const auto solveManeuver = [&segmentStates, &maximumInstant, &aState](const Shared<Thruster>& thrusterDynamics
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +710,233 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+
+            case MaximumManeuverDurationViolationStrategy::TrailingSlice:
+            {
+                const Interval validManeuverInterval = Interval::Closed(
+                    candidateManeuverInterval.getEnd() - maneuverConstraints_.maximumDuration,
+                    candidateManeuverInterval.getEnd()
+                );
+                const Shared<Thruster> slicedThruster =
+                    buildThrusterDynamicsWithManeuverIntervals(validManeuverInterval);
+                const auto [maneuverSolution, maneuvers] = solveManeuver(slicedThruster);
```
apaletta3: Since these are not used, is there a way to just not set them in cpp like we do in python?
```suggestion
                const auto [maneuverSolution, ~] = solveManeuver(slicedThruster);
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -584,38 +710,233 @@ const NumericalSolver& Segment::accessNumericalSolver() const
...
+        previousManeuverInterval = maneuverInterval;
+
+        segmentHeterogenousGuidanceLaw->addGuidanceLaw(
+            maneuverSolution.getThrusterDynamics()->getGuidanceLaw(), maneuverInterval
+        );
+    };
+
+    // Helper lambda to handle maximum duration violation strategies, returning true if the segment condition is
+    // satisfied. It also updates the segment states and the guidance law if the maneuver is accepted.
+    const auto handleMaximumDurationViolation = [&](const Interval& candidateManeuverInterval) -> bool
```
apaletta3: Same here if you can be explicit with what variables you're capturing in the lambda, since these values get potentially mutated

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Segment.cpp
```diff
@@ -29,6 +29,61 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Segment(pybind11::module&
...
+        "MaximumManeuverDurationViolationStrategy",
+        R"doc(
+            Strategy to use when a maneuver exceeds the maximum duration constraint.
+
+            
+            Proposed maneuver:                                                                         [--------------------|------------]
+            Fail: Will throw an exception.
+            Skip: Will skip the maneuver.
+            LeadingSlice: Will shorten the maneuver to the maximum duration, sliced from the start.    [--------------------]
+            TrailingSlice: Will shorten the maneuver to the maximum duration, sliced from the end.                  [--------------------]
```
apaletta3: Can you include these docstrings in the cpp files as well?

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -1940,11 +1738,1119 @@ TEST_P(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment_Solve_Parameterized, Ma
...
+    const Segment::ManeuverConstraints constraints(
+        Duration::Minutes(10.0),  // Minimum duration (large, so short maneuvers will be rejected)
+        Duration::Undefined(),
+        Duration::Undefined(),
+        Segment::MaximumManeuverDurationViolationStrategy::Fail
+    );
+
+    // Event condition that will be satisfied during coast
+    const Shared<InstantCondition> eventCondition = std::make_shared<InstantCondition>(
+        InstantCondition::Criterion::AnyCrossing, initialStateWithMass_.accessInstant() + Duration::Minutes(5.0)
```
apaletta3: Why not use a duration condition like the test above and below?
Also, are you sure that it is coasting when the event condition is met? The defaultThrusterDynamicsSPtr_ looks like they are constantly thrusting

```suggestion
    // Event condition that will be satisfied before the Minimum maneuver duration is reached
    const Shared<RealCondition> durationCondition = std::make_shared<RealCondition>(
        RealCondition::DurationCondition(RealCondition::Criterion::AnyCrossing, Duration::Minutes(5.0))
    );
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -1940,11 +1738,1119 @@ TEST_P(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment_Solve_Parameterized, Ma
...
+    const Array<Interval> guidanceLawIntervals = {
+        Interval::Closed(
+            initialStateWithMass_.accessInstant() + Duration::Minutes(5.0),
+            initialStateWithMass_.accessInstant() + Duration::Minutes(10.0)
+        ),
+        Interval::Closed(
+            initialStateWithMass_.accessInstant() + Duration::Minutes(15.0),
+            initialStateWithMass_.accessInstant() + Duration::Minutes(30.0)
+        ),
+    };
```
apaletta3: The comment and the code don't indicate the same thing

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -65,6 +65,65 @@ class Segment
...
+        /// @param displayDecorator If true, display decorators
+        void print(std::ostream& anOutputStream, bool displayDecorator = true) const;
+
+        /// @brief Output stream operator
+        /// @param anOutputStream An output stream
+        /// @param aManeuverConstraints A maneuver constraints
+        /// @return An output stream
+        friend std::ostream& operator<<(std::ostream& anOutputStream, const ManeuverConstraints& aManeuverConstraints);
+
+        String MaximumManeuverDurationViolationStrategyToString(
```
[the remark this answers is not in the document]
vishwa2710: Ah yeah I forgot we added that, it's not implicit you have to define the macros, but indeed we can leverage the macros.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -65,6 +65,65 @@ class Segment
...
+
+        /// @brief Check if the minimum duration is valid
+        /// @param aManeuverInterval The maneuver interval
+        /// @return True if the minimum duration is valid
+        bool intervalHasValidMinimumDuration(const Interval& aManeuverInterval) const;
+
+        /// @brief Check if the maximum duration is valid
+        /// @param aManeuverInterval The maneuver interval
+        /// @return True if the maximum duration is valid
+        bool intervalHasValidMaximumDuration(const Interval& aManeuverInterval) const;
```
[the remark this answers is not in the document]
vishwa2710: The issue with this is that it reads more like the maneuverConstraints object has a valid minimum duration constraint. How about `validateIntervalForMinimumDuration `?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: Going to keep it as is, cause if we switch to magicenum it will just to a direct string conversion (without the spacing)

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -1645,220 +1569,96 @@ INSTANTIATE_TEST_SUITE_P(
...
-        // Leading edge maneuver - allow multiple
+        // Leading edge maneuver
         SolveTestParams {
             "LeadingEdge_MultipleManeuvers",
             Array<Tuple<Duration, Duration>> {
-                Tuple<Duration, Duration> {Duration::Minutes(-10.0), Duration::Minutes(10.0)}
+                Tuple<Duration, Duration> {Duration::Minutes(-1.0), Duration::Minutes(1.0)}
             },
+            Duration::Minutes(6.0),
             true,
```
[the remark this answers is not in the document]
vishwa2710: These tests can't really test for the failure case as the eventCondition duration is fixed, the test `Solve_LoopExitsDueToMaximumInstant` covers this case.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -1940,11 +1738,1119 @@ TEST_P(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment_Solve_Parameterized, Ma
...
+    );
+
+    const Segment::Solution solution =
+        maneuverSegment.solve(initialStateWithMass_, Duration::Minutes(30.0), previousManeuverInterval);
+
+    EXPECT_TRUE(solution.conditionIsSatisfied);
+    EXPECT_EQ(solution.extractManeuvers(defaultFrameSPtr_).getSize(), 0);
+}
+
+TEST_F(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment, Solve_MinimumSeparationRequired_AlreadyPastTarget)
```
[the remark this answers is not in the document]
vishwa2710: I think AlreadyPastTarget is more specific than NoCoastAdded, because a coast could be not added for a variety of reasons, but in this case we want to specify that it's. because we're already past the previous maneuver end.
Updated the name to PastMinimumSeparationTarget

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
vishwa2710: I can't remember why I started out with Instant Conditions, but they can all be durations, I'll align. 
As for the coasting, it's going to be coasting as it's going to skip the maneuver due to the minimum maneuver duration.

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.test.cpp
```diff
@@ -1940,11 +1738,1119 @@ TEST_P(OpenSpaceToolkit_Astrodynamics_Trajectory_Segment_Solve_Parameterized, Ma
...
+    if (maneuvers.getSize() > 0)
+    {
+        // Centered maneuver should have duration equal to maximum
+        EXPECT_LE(maneuvers[0].getInterval().getDuration(), Duration::Minutes(5.0));
+    }
+}
+
+TEST_F(
+    OpenSpaceToolkit_Astrodynamics_Trajectory_Segment,
+    Solve_ManeuverDurationExceedsMaximum_CenterStrategy_MultipleManeuvers
```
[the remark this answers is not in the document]
vishwa2710: This is a hangover of the AllowMultipleManeuvers flag that was existing before I simplified everything, Will remove the single case.

PR #601, opened 2025-11-11
test!: update close approach generator defaults and add more thorough tests 

apaletta3:
breaking MR to be lumped in with our major OSTk astro release that:
- clarifies the docstrings of how to set the time step of the temporal condition solver
- changes the default value of the step size of the close approach generator to 20 minutes, because there can be up to 4 intervals where two objects are moving apart from each other in LEO, and 20 is smaller than ~85min/4 meaning that it will always catch one of those events
- added another test to clarify the two vs one close approach per orbit geometrical case

## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance for time_step/step selection, clarifying trade-offs, requiring the step be smaller than the smallest expected condition-change interval, and providing examples.

* **Chores**
  * Changed default step size for close approach generation from 15 seconds to 20 minutes and aligned inline docs.

* **Tests**
  * Updated test fixtures and expectations to match the new default step and revised sampling cadence.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Solver/TemporalConditionSolver.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Solver/TemporalConditionSolver.hpp
@@ -40,6 +40,12 @@ class TemporalConditionSolver
     ///                  Duration::Microseconds(1.0) };
     /// @endcode
     ///
+    /// @note Be careful When selecting the time_step. A very small step can lead to higher precision, but increased
+    /// runtime and memory consumption. On the other hand, a step that is too large, can result in missing event windows
+    /// that are shorter than the time_step. For example:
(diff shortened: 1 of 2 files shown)


include/OpenSpaceToolkit/Astrodynamics/Solver/TemporalConditionSolver.hpp
```diff
@@ -40,7 +40,9 @@ class TemporalConditionSolver
     ///                  Duration::Microseconds(1.0) };
     /// @endcode
     ///
-    /// @param aTimeStep A time step used to generate the temporal grid.
+    /// @param aTimeStep A time step used to generate the temporal grid, within which condition switching instants are
```
vishwa2710: Update accordingly if you agreed with my feedback above.

vishwa2710:
Is this a breaking change as we're updating the default values?

PR #605, opened 2025-11-20
docs: clarify mltan input

apaletta3:
LTAN and MLTAN should be computed using the mean RAAN instead of the osculating RAAN, because using the osculating RAAN makes the LTAN/MLTAN value fluctuate within an orbital period. For completeness, the Mean in MLTAN signifies mean solar time, which is the solar time averaged over an entire year.

## Summary by CodeRabbit

* **Documentation**
  * Added notes recommending the use of a mean RAAN variant for more stable LTAN/LTDN calculations.
  * Clarified argument wording to state the reference is the Ascending Node (RAAN) and emphasized that RAAN refers to mean, not osculating, elements.


vishwa2710 (dismissed):
Just the bots comment.

PR #616, opened 2025-12-02
refactor: revert breaking change in orbit geosynchronous constructor

apaletta3:
There was an unintentional break in the API in the previous PR (https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/615), so this is an MR to revert the breaking part of that change. This MR is not marked as breaking because the previous MR has not been released yet

Plus, I think using geosynchronous as a general term is good enough because:
- it is technically more correct than stationary (because providing a non-zero inclination makes it not stationary)
- even though it has "geo" in the name, people do say "Mars geostarionary orbit" or at least would understand what you're talking about (see this https://space.stackexchange.com/questions/40166/name-for-geostationary-orbit-around-another-planet)

## Summary by CodeRabbit

* **Refactor**
  * Renamed the "stationary" geosynchronous orbit factory to "geo_synchronous" for consistent naming across APIs.

* **API**
  * Longitude parameter now expects an Angle-type value instead of a plain numeric value.

* **Documentation**
  * Updated docstrings and descriptions to reflect geo-synchronous terminology and the revised parameter type.

* **Tests**
  * Test suites updated to exercise the renamed factory and cover related constructors and conversions.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit.cpp
@@ -357,7 +357,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Orbit(pybind11::module& a
                     Args:
                         epoch (Instant): The epoch.
                         inclination (Angle): The inclination.
-                        longitude (double): The longitude.
+                        longitude (Angle): The longitude.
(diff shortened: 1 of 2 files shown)


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit.cpp
apaletta3:
```suggestion
                        longitude (Angle): The longitude.
```

PR #623, opened 2025-12-17
feat: support state builder expansion and reduction for any reference frame

vishwa2710:
I believe the state builder reduce method should not raise an error if the frames are different. This is because the reduce method is merely _removing_ coordinates from the provided state, it doesn't leverage the frame in any way. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduction and expansion now tolerate differing input frames: operations return a reparented state in the builder’s frame instead of throwing a frame-mismatch error.

* **Documentation**
  * Clarified method docstrings: reduce/expand explicitly state the returned state is provided in the StateBuilder’s frame.

* **Tests**
  * Tests updated to assert returned states are distinct, use the builder’s frame, preserve instants, and have expected coordinate sizes.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.cpp
@@ -173,16 +173,6 @@ const State StateBuilder::expand(const State& aState, const State& defaultState)
         throw ostk::core::error::runtime::Undefined("Default State");
     }
 
-    if (defaultState.accessFrame() != aState.accessFrame())
-    {
-        throw ostk::core::error::runtime::Wrong(
-            "Default State Frame",
-            String::Format(
-                "Expected: {}, Got: {}", aState.accessFrame()->getName(), defaultState.accessFrame()->getName()
-            )
-        );
-    }
-
     VectorXd coordinates = VectorXd(this->coordinatesBrokerSPtr_->getNumberOfCoordinates());
     Integer nextIndex = 0;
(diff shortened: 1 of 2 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/StateBuilder.cpp
```diff
@@ -157,7 +152,7 @@ const State StateBuilder::reduce(const State& aState) const
         nextIndex += subsetCoordinates.size();
     }
 
-    return this->build(aState.accessInstant(), coordinates);
+    return {aState.accessInstant(), coordinates, aState.accessFrame(), this->coordinatesBrokerSPtr_};
```
phc1990: Re-frame then reduce

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/StateBuilder.cpp
```diff
@@ -157,6 +157,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_StateBuilder(pybind11::mo
             arg("state"),
             R"doc(
                 Reduce a `State` object to the coordinate subsets of the `StateBuilder`.
+                The output state is provided in the Frame of the `StateBuilder`.
```
phc1990: Gonna leave this comment here at the top:

Don't think this is a fix any more, I would rename the PR to something like:

"feat: allow state builder expansion and reduction for any reference frame"

(or something like that)

phc1990 (approved):

PR #627, opened 2025-12-24
feat: use celestial frame instead of ITRF when possible

vishwa2710:
Cleaning up any hardcoded Frame::ITRF(), ideally these are taken from the relevant Celestial's ephemeris body frame, rather than always using the Frame::ITRF() (which is an order of magnitude slower than the SPICE based frame).

  1. Access Generator (Generator.cpp, Generator.hpp)

  - Updated internal methods to accept Shared<const Celestial> parameter
  - Uses celestialSPtr->accessFrame() instead of Frame::ITRF() for position transformations
  - Uses celestial's getEquatorialRadius() and getFlattening() for LLA calculations
  - Uses Celestial::FrameType::NED instead of Earth::FrameType::NED
  - AccessTarget::FromLLA now uses aCelestialSPtr->accessFrame() for positions
  - Added warning if environment lacks a central celestial object

  2. Line of Sight Visibility Criterion (LineOfSight.cpp)

  - Uses environment.accessCentralCelestialObject()->accessFrame() instead of hardcoded ITRF
  - Falls back to Frame::ITRF() if no central celestial object exists (will eventually be deprecated)
  - Added warning for missing central celestial object

  3. Atmospheric Drag Dynamics (AtmosphericDrag.cpp)

  - Uses celestialObjectSPtr_->accessFrame() for computing Earth angular velocity

  4. Flight Profile (Profile.hpp, Profile.cpp, bindings)

  - New overload: AlignAndConstrain(alignmentTarget, clockingTarget, celestialSPtr, angularOffset)
  - Deprecated: Original AlignAndConstrain without celestial parameter
  - CustomPointing now uses anOrbit.accessCelestialObject() to get the celestial
  - Updated helper functions:
    - ComputeGeodeticNadirDirectionVector(state, celestialSPtr) - uses celestial's frame and shape parameters
    - ComputeTargetSlidingGroundVelocityVector(state, trajectory, celestialFrameSPtr) - uses provided frame
  - Python bindings updated with deprecation warning for old signature

  5. Trajectory Components

  - Static.cpp: Minor frame-related updates
  - Orbit.cpp: Uses celestial frame where appropriate
  - COE.cpp: Minor updates for celestial frame usage

  6. Tests (Profile.test.cpp)

  - Added AlignAndConstrain_Celestial test verifying:
    - GeodeticNadir with Earth celestial matches original behavior
    - TargetSlidingGroundVelocity with Earth celestial matches original
    - Angular offset works with celestial parameter

  Migration Path

  Replace:
```
  Profile::AlignAndConstrain(alignmentTarget, clockingTarget, angularOffset);
```
  With:
```
  Profile::AlignAndConstrain(alignmentTarget, clockingTarget, celestialSPtr, angularOffset);
```

When instantiating `Generator`, or `LineOfSight` Visibility Conditions, use an environment that has a central celestial body (the default Environment does).

## Summary by CodeRabbit

* **New Features**
  * Alignment, access, trajectory, dynamics, and visibility calculations now support explicit celestial contexts and reference frames for bodies beyond Earth.
  * Access generation uses the environment’s central celestial body and its shape parameters when available.
* **Deprecations**
  * The legacy alignment overload is deprecated; use the celestial-aware variant instead.
* **Backward Compatibility**
  * The legacy overload remains available, emits a warning, and defaults to Earth.
* **Tests**
  * Updated coverage validates celestial-aware alignment and Earth-context results.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
@@ -16,6 +16,7 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
     using ostk::mathematics::geometry::d3::transformation::rotation::Quaternion;
 
     using ostk::physics::coordinate::Frame;
+    using ostk::physics::environment::object::Celestial;
 
     using ostk::astrodynamics::flight::Profile;
     using ostk::astrodynamics::flight::profile::Model;
@@ -503,7 +504,19 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
 
         .def_static(
             "align_and_constrain",
-            &Profile::AlignAndConstrain,
+            [](const Shared<const Profile::Target>& anAlignmentTargetSPtr,
+               const Shared<const Profile::Target>& aClockingTargetSPtr,
+               const Angle& anAngularOffset) -> std::function<Quaternion(const State&)>
+            {
+                PyErr_WarnEx(
+                    PyExc_DeprecationWarning,
+                    "Use AlignAndConstrain(const Shared<const Target>& anAlignmentTargetSPtr, const Shared<const "
+                    "Target>& aClockingTargetSPtr, const Shared<const Celestial>& aCelestialSPtr, const Angle& "
+                    "anAngularOffset = Angle::Zero()) instead.",
+                    1
+                );
+                return Profile::AlignAndConstrain(anAlignmentTargetSPtr, aClockingTargetSPtr, anAngularOffset);
+            },
             R"doc(
                 Generate a function that provides a quaternion that aligns to the `alignment_target` and constrains to the `clocking_target` for a given state.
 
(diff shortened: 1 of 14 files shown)


alex-liang3 (approved):

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Flight/Profile.cpp
```diff
@@ -521,6 +534,33 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Flight_Profile(pybind11::module& aMo
...
+        .def_static(
+            "align_and_constrain",
+            overload_cast<
+                const Shared<const Profile::Target>&,
+                const Shared<const Profile::Target>&,
+                const Shared<const Celestial>&,
+                const Angle&>(&Profile::AlignAndConstrain),
+            R"doc(
+                Generate a function that provides a quaternion that aligns to the `alignment_target` and constrains to the `clocking_target` for a given state.
+                Uses the celestial object's frame instead of ITRF for geodetic nadir and sliding ground velocity calculations.
```
apaletta3: can you clarify this docstring? It makes it sound like the "celestial object's frame" and ITRF are two different things, when really one is just a faster version of the other

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -406,10 +406,28 @@ class Profile
...
+        const Angle& anAngularOffset = Angle::Zero()
+    );
 
+    /// @brief Generate a function that provides a quaternion that aligns and constrains for a given state.
+    ///
+    /// @param anAlignmentTargetSPtr An alignment target
+    /// @param aClockingTargetSPtr A clocking target
+    /// @param aCelestialSPtr A celestial object whose frame will be used instead of ITRF for geodetic nadir and
+    /// sliding ground velocity calculations
+    /// @param anAngularOffset An angular offset applied to the clocking axis
```
apaletta3: same comment as above on clarifying docstring

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
```diff
@@ -624,9 +624,10 @@ Shared<const Frame> Orbit::getOrbitalFrame(const Orbit::FrameType& aFrameType) c
                     std::sin(latitude_radians),
                 };
 
-                const Vector3d xAxis = -(Frame::ITRF()->getTransformTo(Frame::GCRF(), aState.accessInstant()))
-                                            .applyToVector(nadir_ITRF)
-                                            .normalized();
+                const Vector3d xAxis =
+                    -(this->celestialObjectSPtr_->accessFrame()->getTransformTo(Frame::GCRF(), aState.accessInstant()))
```
apaletta3: This technically doesn't generalize to any celestial body because GCRF doesn't generalize to any celestial body, even though your ITRF equivalent now does. Worth raising an error?

apaletta3 (approved):
Great stuff! Some small comments to apply.

It would be great to also clarify (by either renaming variables or improving docstrings) that the celestialObjectSPtr->accessFrame() is indeed return the celestial object fixed frame, and not the inertial (or any other) frame centered at the body. It wasn't immediately obvious to me that that was what was happening under the hood, but I'm guessing that the `Ephemeris` class has the assumption baked into it (that its the celestial object fixed and not inertial frame), we've just not used that class explicitly.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit.cpp
vishwa2710: What are checking for here though? If the celestial body is not earth? I think for now this is a benign enough change, don't want to make a breaking change where the inertial frame also needs to be provided.

PR #628, opened 2026-01-01
feat: generalize condition based solving

vishwa2710:
Leverage the provided numerical solver to compute the Event Crossings. Previously, only dense steppers are supported. Now, instead of leveraging the dense interpolant, integration is performed between the previous step and the current step (post crossing step), and a CubicSpline interpolant is used to solve for the event crossing efficiently.
This unlocks using RK78 for various applications (such as within the Segment solver).

The segment benchmark with this reads:
| Benchmark                                                             | Time    | duration_h | final_sma_difference | maneuvers | max_step_s | min_step_s |
|-----------------------------------------------------------------------|---------|------------|----------------------|-----------|------------|------------|
| BM_Segment_ConstantThrust_Intrack_550_to_580/iterations:1             | 1.46 |  16.0027    | 4.65661n            | 20        | 94.7937    | 0.129444    |
| BM_Segment_QLaw_Analytical_SMA_550_to_580/iterations:1                | 5.29 |  30.8055    | 931.323p            | 20        | 94.9888    | 0.0183992   |
| BM_Segment_QLaw_FiniteDifference_SMA_550_to_580/iterations:1          | 11.7 |  30.8053    | 2.79397n            | 20        | 94.9887    | 0.765134    |
| BM_Segment_QLaw_Analytical_Frozen_550_to_580/iterations:1             | 8.06 |  40.6546    | 3.72529n            | 45        | 94.6473    | 1.33437m    |
| BM_Segment_ConstantThrust_Intrack_DutyCycle_550_to_580/iterations:1   | 11.4 |  48         | -23.8683k           | 4         | 94.2705    | 5           |

A huge speed up compared to main:
| Benchmark                                                             | Time    | duration_h | final_sma_difference | maneuvers | max_step_s | min_step_s |
|-----------------------------------------------------------------------|---------|------------|----------------------|-----------|------------|------------|
| BM_Segment_ConstantThrust_Intrack_550_to_580/iterations:1             | 5.04 |  16.0027    | 5.58794n            | 20        | 13.6413    | 0.315719   |
| BM_Segment_QLaw_Analytical_SMA_550_to_580/iterations:1                | 19.7 |  30.8056    | 4.65661n            | 20        | 13.6803    | 671.847u   |
| BM_Segment_QLaw_FiniteDifference_SMA_550_to_580/iterations:1          | 41.7 |  30.8056    | 931.323p            | 20        | 13.6804    | 303.349u   |
| BM_Segment_QLaw_Analytical_Frozen_550_to_580/iterations:1             | 142  |  40.6558    | 1.86265n            | 45        | 13.6921    | 20.212u    |
| BM_Segment_ConstantThrust_Intrack_DutyCycle_550_to_580/iterations:1   | 46.9 |  48         | -23.8683k           | 4         | 13.6573    | 0.469988   |

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
@@ -366,7 +366,15 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
             )
             .def_static(
                 "default_conditional",
-                &NumericalSolver::DefaultConditional,
+                +[](const std::function<void(const State&)>& stateLogger = nullptr) -> NumericalSolver
+                {
+                    PyErr_WarnEx(
+                        PyExc_DeprecationWarning,
+                        "NumericalSolver.default_conditional is deprecated. Use NumericalSolver.default instead.",
+                        1
+                    );
+                    return NumericalSolver::DefaultConditional(stateLogger);
+                },
                 R"doc(
                     Return the default conditional numerical solver.
 
@@ -380,7 +388,20 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
             )
             .def_static(
                 "conditional",
-                &NumericalSolver::Conditional,
+                +[](const Real& aTimeStep,
+                    const Real& aRelativeTolerance,
+                    const Real& anAbsoluteTolerance,
+                    const std::function<void(const State&)>& stateLogger = nullptr) -> NumericalSolver
+                {
+                    PyErr_WarnEx(
+                        PyExc_DeprecationWarning,
+                        "NumericalSolver.conditional is deprecated. Use NumericalSolver constructor instead.",
+                        1
+                    );
+                    return NumericalSolver::Conditional(
+                        aTimeStep, aRelativeTolerance, anAbsoluteTolerance, stateLogger
+                    );
+                },
                 R"doc(
                     Return a conditional numerical solver.
 
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
@@ -51,6 +51,11 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
             "Re-integrate with smaller steps during root finding"
         )
         .value(
+            "CubicInterpolation",
+            NumericalSolver::RootFindingStrategy::CubicInterpolation,
+            "Cubic interpolation between step endpoints"
+        )
+        .value(
             "Skip",
             NumericalSolver::RootFindingStrategy::Skip,
             "Skip the root finding and return the first step boundary where condition is satisfied"
@@ -228,9 +233,8 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
                         pybind11::cast<pythonSystemOfEquationsSignature>(aSystemOfEquationsObject);
 
                     const NumericalSolver::SystemOfEquationsWrapper& systemOfEquations =
-                        [&](const NumericalSolver::StateVector& x,
-                            NumericalSolver::StateVector& dxdt,
-                            const double t) -> void
+                        [&](const NumericalSolver::StateVector& x, NumericalSolver::StateVector& dxdt, const double t
+                        ) -> void
                     {
                         dxdt = pythonDynamicsEquation(x, dxdt, t);
                     };
@@ -265,9 +269,8 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
                         pybind11::cast<pythonSystemOfEquationsSignature>(aSystemOfEquationsObject);
 
                     const NumericalSolver::SystemOfEquationsWrapper& systemOfEquations =
-                        [&](const NumericalSolver::StateVector& x,
-                            NumericalSolver::StateVector& dxdt,
-                            const double t) -> void
+                        [&](const NumericalSolver::StateVector& x, NumericalSolver::StateVector& dxdt, const double t
+                        ) -> void
                     {
                         dxdt = pythonDynamicsEquation(x, dxdt, t);
                     };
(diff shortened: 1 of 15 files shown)


apaletta3 (commented):
Took an initial quite look, very cool stuff!
got a couple questions that I'd like to discuss in a huddle

phc1990 (commented):
(unfinished review, will continue later)

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
```diff
@@ -33,6 +33,34 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
...
+            The strategy for finding the exact event crossing time during conditional integration.
+        )doc"
+    )
+        .value(
+            "LinearInterpolation",
+            NumericalSolver::RootFindingStrategy::LinearInterpolation,
+            "Linear interpolation between step endpoints"
+        )
+        .value(
+            "Propagated",
```
phc1990, line 49: Nit: I think I might have brought this up already, would go for `Propagation` if we go for `LinearInterpolation` or `Propagated` if we go for `LinearlyPropagated`, etc. just for consistency.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/State/NumericalSolver.cpp
```diff
@@ -33,6 +33,34 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_State_NumericalSolver(pyb
...
+            NumericalSolver::RootFindingStrategy::Propagated,
+            "Re-integrate with smaller steps during root finding"
+        )
+        .value(
+            "CubicInterpolation",
+            NumericalSolver::RootFindingStrategy::CubicInterpolation,
+            "Cubic interpolation between step endpoints"
+        )
+        .value(
+            "Skip",
```
phc1990, line 59: Nit: maybe "First"? 

So there is no chance for confusion, e.g. "does this mean it _skips_ the event?"

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -1528,10 +1530,14 @@ Segment::Solution Segment::solve(
...
-            constructLOFCompliantManeuverSolution(maneuverSubSegmentSolution, subsegmentManeuver.value());
-        acceptManeuver(maneuverLOFCompliantSolution, subsegmentManeuver.value());
+        else
+        {
+            // Candidate maneuver passed all constraints - accept it
+            const Segment::Solution maneuverLOFCompliantSolution =
+                constructLOFCompliantManeuverSolution(maneuverSubSegmentSolution, subsegmentManeuver.value());
+
+            acceptManeuver(maneuverLOFCompliantSolution, subsegmentManeuver.value());
+        }
```
phc1990, line 1540: This is just personal opinion, so I leave it up to you: we didn't have an if/else here because that if (the cone that checks max man duration constraint above) is not the only one.

It's the 3rd one (after the minimum duration, maximum duty cycle), of a series of N (if / continue) blocks.

With the `// Candidate maneuver passed all constraints - accept it` being the last step after going through all the "N" blocks.

The code behaviour does not change, but it might give the impresion that that last `// Candidate maneuver passed all constraints - accept it` block is in opposition to the max duration block only.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -1804,7 +1828,25 @@ Segment::Solution Segment::solveUntilThrusterOn_(
         Array<Shared<EventCondition>> {eventCondition_, thrusterToggleCondition}
     );
 
-    Segment::Solution solution = solveWithDynamics_(aState, anEndInstant, freeDynamicsArray_, combinedCondition);
```
phc1990, line 1807: would it make sense to expand `solveWithDynamics_` private function signature with an optional `maxStepSize` argument? 

Then we can keep it re-using in both thurster on/off condition searches

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -1780,7 +1786,25 @@ Segment::Solution Segment::solveUntilThrusterOff_(
         Array<Shared<EventCondition>> {eventCondition_, thrusterToggleCondition}
     );
 
-    Segment::Solution solution = solveWithDynamics_(aState, anEndInstant, dynamicsArray, combinedCondition);
+    // Use a solver with a max step size to ensure the integrator doesn't overshoot the thrust-off
+    // boundary. The thrust toggle condition is a step function that can be missed entirely if the
+    // adaptive stepper takes a step larger than the remaining thrust window.
+    NumericalSolver stepLimitedSolver = numericalSolver_;
+    stepLimitedSolver.setMaxStepSize((anEndInstant - aState.accessInstant()).inSeconds() / 10.0);
```
phc1990, line 1793: Should we set the `numericalSolver_` step size back to the original after the search?

Would it be better to remove the `setMaxStepSize` and just create a copy of the numerical solver to use?

(see other comment suggesting to expand private function `solveWithDynamics_` to keep this logic centralized_

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -1804,7 +1828,25 @@ Segment::Solution Segment::solveUntilThrusterOn_(
         Array<Shared<EventCondition>> {eventCondition_, thrusterToggleCondition}
     );
 
-    Segment::Solution solution = solveWithDynamics_(aState, anEndInstant, freeDynamicsArray_, combinedCondition);
+    // Use a solver with a max step size to ensure the integrator doesn't overshoot the thrust-on
+    // boundary. The thrust toggle condition is a step function that can be missed entirely if the
+    // adaptive stepper takes a step larger than the thrust window.
+    NumericalSolver stepLimitedSolver = numericalSolver_;
+    stepLimitedSolver.setMaxStepSize((anEndInstant - aState.accessInstant()).inSeconds() / 10.0);
```
phc1990, line 1835: duration between start and end instant could be arbitrarily big

would it make sense to have this on/off thruster search step size as a static variable that we consider safe?

phc1990 (commented):
Nice, left some comments, lmk what you think 👍

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
vishwa2710: Just created a new one.

phc1990 (approved):
Great perf. bump!

PR #637, opened 2026-01-15
fix: numerical solver crossings

vishwa2710:
Bunched together a few small fixes as I was trying to fix the Numerical Solver crossings. IMO the NumericalSolver condition based solution should solve until just _after_ crossing the root (or the exact root in the case it exists)

- Ensure condition solution crosses the boundary: Modified NumericalSolver::integrateTime to use solution.upperBound or solution.lowerBound (based on integration direction) instead of solution.root to guarantee the returned state has actually crossed the condition threshold
- Add RootSolver bounds: Extended RootSolver::Solution to include lowerBound and upperBound fields for precise boundary handling
- Fix LogicalCondition for root-finding: Added evaluate() override to LogicalCondition that returns:
    - max of sub-conditions for OR (triggers when any condition crosses)
    - min of sub-conditions for AND (triggers when all conditions cross)
- Fix Instant condition so that it's not referencing J2000 and accumulating floating point precision issues.
- Refactor integrateTime tests for performance (just reduced durations)

## Summary by CodeRabbit

* **New Features**
  * Event conditions gain evaluate() and evaluate_negative_when_satisfied() to get numeric condition values and satisfaction polarity.
  * Angular conditions provide normalized angular-distance evaluation and related helpers.
  * Root solver solutions now include lower and upper convergence bounds alongside the reported root.

* **API Changes**
  * InstantCondition constructor argument order changed: criterion now comes before instant.

### DIFF
--- a/test/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.test.cpp
+++ b/test/OpenSpaceToolkit/Astrodynamics/Trajectory/Sequence.test.cpp
@@ -16,7 +16,6 @@
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/Thruster.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/EventCondition/COECondition.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/EventCondition/InstantCondition.hpp>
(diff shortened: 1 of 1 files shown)


vishwa2710:
Closing in favor of https://github.com/open-space-collective/open-space-toolkit-astrodynamics/pull/638

PR #647, opened 2026-02-12
feat: add maximum maneuver duty cycle constraint

phc1990:
Adds maximum maneuver duty cycle constraint. Expressed as "time over time", e.g. {40 min, 98min} meaning as much as 40 min of burn for any 98 min interval).

## Summary by CodeRabbit

* **New Features**
  * Added duty-cycle constraint support for maneuver planning (numerator/denominator pairs) with validation and reporting.
  * Solver now accounts for duty-cycle limits when scheduling maneuvers, can recommend or enforce coast times, and uses a numeric fallback for tightly constrained cases.
  * Maneuver constraint display updated to include duty-cycle state.

* **Tests**
  * Expanded coverage for duty-cycle scenarios, interval validation, boundary/edge cases, and invalid-interval error paths.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
@@ -49,112 +49,6 @@ using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianVelocit
 using ostk::astrodynamics::trajectory::StateBuilder;
 using FlightManeuver = ostk::astrodynamics::flight::Maneuver;
 
-bool Segment::ManeuverConstraints::isDefined() const
-{
-    return minimumDuration.isDefined() || maximumDuration.isDefined() || minimumSeparation.isDefined() ||
-           (maximumDutyCycle.first.isDefined() && maximumDutyCycle.second.isDefined());
-}
-
-bool Segment::ManeuverConstraints::intervalHasValidMinimumDuration(const Interval& aManeuverInterval) const
-{
-    if (!minimumDuration.isDefined())
-    {
-        return true;
-    }
-
-    return aManeuverInterval.getDuration() >= minimumDuration;
-}
-
-bool Segment::ManeuverConstraints::intervalHasValidMaximumDuration(const Interval& aManeuverInterval) const
-{
-    if (!maximumDuration.isDefined())
-    {
-        return true;
-    }
-
-    return aManeuverInterval.getDuration() <= maximumDuration;
-}
-
-bool Segment::ManeuverConstraints::intervalHasValidMaximumDutyCycle(
-    const Interval& aManeuverInterval, const Array<Interval>& aPreviousManeuverIntervals
-) const
-{
-    // If there are no maximum duty cycle constraints, return valid
-    if (!maximumDutyCycle.first.isDefined() || !maximumDutyCycle.second.isDefined())
-    {
-        return true;
-    }
-
-    for (const auto& previousManeuverInterval : aPreviousManeuverIntervals)
-    {
-        if (previousManeuverInterval.getEnd() >= aManeuverInterval.getStart())
-        {
-            throw ostk::core::error::RuntimeError(
-                "Previous maneuver intervals must be before the candidate maneuver interval."
-            );
-        }
-    }
-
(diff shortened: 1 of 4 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -101,12 +104,41 @@ class Segment
...
+        ///
+        /// @param aMaximumDutyCycle The maximum duty cycle as numerator and denumerator. For example,
+        /// {Duration::Minutes(40.0), Duration::Minutes(98.0)} represents a maximum maneuvering time of 40 minutes over
+        /// any 98 minutes interval.
+        /// @param aMinimumDuration The minimum duration for a maneuver. Defaults to Undefined.
+        /// @param aMaximumDuration The maximum duration for a maneuver. Defaults to Undefined.
+        /// @param aMinimumSeparation The minimum separation between maneuvers. Defaults to Undefined.
+        /// @param aMaximumDurationStrategy The strategy when maximum duration is violated. Defaults to Fail.
+        /// @return ManeuverConstraints instance
+        static ManeuverConstraints ManeuverDutyCycle(
```
apaletta3: Nit: Perhaps we should wait decide on how best we want to do this breaking change factory method pattern (i.e. naming convention) before introducing it? We can simply make another regular constructor with the 5th argument for the time being, right?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -105,12 +104,39 @@ class Segment
             const Duration& aMinimumSeparation = Duration::Undefined(),
             const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy =
                 MaximumManeuverDurationViolationStrategy::Fail
+        )
+            : ManeuverConstraints(
```
vishwa2710: Let's define this in the cpp rather the hpp.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.cpp
```diff
@@ -868,12 +982,210 @@ Segment::Solution Segment::solve(
...
+
+                        return solveAndAcceptCoast(candidateManeuverInterval.getEnd());
+                    }
+
+                    newDuration -= stepSize;
+                }
+
+                // Could not find a valid maneuver interval, here we can't skip to the end of the maneuver,
+                // the fact that we could not allocate a chunk at the beginning of the maneuver interval does
+                // not mean that we can't allocate a chunk later.
```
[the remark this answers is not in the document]
phc1990: changed docs, should be correct and clearer now (it was not correct before)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Segment.hpp
```diff
@@ -107,10 +110,28 @@ class Segment
...
+        /// @param aMaximumDurationStrategy The strategy when maximum duration is violated.
+        /// @param aMaximumDutyCycle The maximum duty cycle as numerator and denominator. For example,
+        ///     {Duration::Minutes(40.0), Duration::Minutes(98.0)} represents a maximum maneuvering time of 40 minutes
+        ///     over any 98 minutes interval.
+        ManeuverConstraints(
+            const Duration& aMinimumDuration,
+            const Duration& aMaximumDuration,
+            const Duration& aMinimumSeparation,
+            const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy,
+            const Pair<Duration, Duration>& aMaximumDutyCycle
```
apaletta3: Shouldn't we have done this to make the user's life easier?
```suggestion
            const Duration& aMinimumDuration = Duration::Undefined(),
            const Duration& aMaximumDuration = Duration::Undefined(),
            const Duration& aMinimumSeparation = Duration::Undefined(),
            const MaximumManeuverDurationViolationStrategy& aMaximumDurationStrategy =
                MaximumManeuverDurationViolationStrategy::Fail,
            const Pair<Duration, Duration>& aMaximumDutyCycle = {Duration::Undefined(),  Duration::Undefined()}
```

PR #664, opened 2026-04-02
chore: add warning for viewer

### LINKED ISSUE #660: [fix] Viewer fails to render any profile without `cesium_asset_id` provided
**Describe the bug**
When using the Viewer, plotting an Orbit works correctly out of the box. For example, this works:
```python
    viewer = Viewer(
        interval=sim_interval,
        cesium_token="<my token>"
        zoom_to_entity=True,
        track_entity=True,
    )
    viewer.add_orbit(orbit, step=Duration.seconds(5.0), color="blue")

    with open("solar_sim.html", "w") as f:
        f.write(viewer.render())
```

The orbital track is visualized and we zoom into the dummy satellite.

However, visualizing a profile does NOT work out of the box:
```
 profile = Profile.custom_pointing(
      orbit=orbit,
      orientation_generator=lambda state: Quaternion.unit(),
 )
viewer.add_profile(
      profile=profile,
      step=Duration.seconds(5.0),
      show_orbital_track=True,
      show_xyz_axes=True,
  )
```

The above show an earth but no hint of any satellite data.

In the [tutorial for this](https://open-space-collective.github.io/open-space-toolkit-astrodynamics/_build/html/_notebooks/Visualize%20Satellite%20Trajectory%20%28Cesium%29.html), the only working example includes a cesium asset ID that is not a common object:
```python
viewer.add_profile(
    profile=profile,
    step=Duration.seconds(5.0),
    show_orbital_track=True,
    cesium_asset_id=669199,
    sensors=[
        ConicSensor(
            name="Payload",
            direction=(0.0, 0.0, 1.0),
            half_angle=Angle.degrees(8.0),
            length=Length.meters(1.0),
            color="red",
        )
    ],
)
```

I suspect that 669199 is a loft specific asset :). 

**Expected behavior**
Either:
- `cesium_asset_id` is a required arg
- a Profile renders correctly even with no cesium asset provided

Probably the second

vishwa2710:
To keep things backwards compatible for now, I've added a deprecation warning. Can refactor this in a future release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a runtime warning when adding a profile without a Cesium asset identifier; rendering and model display remain unchanged for now, but users are alerted that a valid asset ID will be required in a future update.

### DIFF
--- a/bindings/python/tools/python/ostk/astrodynamics/viewer.py
+++ b/bindings/python/tools/python/ostk/astrodynamics/viewer.py
@@ -201,7 +201,7 @@ class Viewer:
 
         if cesium_asset_id is None:
             warnings.warn(
-                "Cesium asset ID is required to render a profile. This will be a required argument in the future."
+                "Cesium asset ID is required to render a profile. This will be a required argument in the future.",
                 FutureWarning,
                 stacklevel=2,
             )
--- a/bindings/python/tools/python/ostk/astrodynamics/viewer.py
+++ b/bindings/python/tools/python/ostk/astrodynamics/viewer.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import functools
 import operator
 from dataclasses import dataclass
+import warnings
 
 import numpy as np
 
@@ -197,6 +198,9 @@ class Viewer:
         Returns:
             Viewer: The Viewer.
         """
+
+        if cesium_asset_id is None:
+            warnings.warn("Cesium asset ID is required to render a profile. This will be a required argument in the future.")
 
         instants: list[Instant] = self._interval.generate_grid(step)
         states: list[State] = profile.get_states_at(instants)
--- a/bindings/python/tools/python/ostk/astrodynamics/viewer.py
+++ b/bindings/python/tools/python/ostk/astrodynamics/viewer.py
@@ -202,6 +202,8 @@ class Viewer:
         if cesium_asset_id is None:
             warnings.warn(
                 "Cesium asset ID is required to render a profile. This will be a required argument in the future."
+                FutureWarning,
+                stacklevel=2,
             )
 
         instants: list[Instant] = self._interval.generate_grid(step)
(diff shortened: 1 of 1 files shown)


bindings/python/tools/python/ostk/astrodynamics/viewer.py
```diff
@@ -198,6 +199,9 @@ def add_profile(
             Viewer: The Viewer.
         """
 
+        if cesium_asset_id is None:
+            warnings.warn("Cesium asset ID is required to render a profile. This will be a required argument in the future.")
+
```
vishwa2710, line 208:
```suggestion
        if cesium_asset_id is None:
            warnings.warn(
                "Cesium asset ID is required to render a profile. This will be a required argument in the future."
            )

```

bindings/python/tools/python/ostk/astrodynamics/viewer.py
```diff
@@ -198,6 +199,11 @@ def add_profile(
             Viewer: The Viewer.
         """
 
+        if cesium_asset_id is None:
+            warnings.warn(
+                "Cesium asset ID is required to render a profile. This will be a required argument in the future."
+            )
```
vishwa2710, line 207:
```suggestion
            warnings.warn(
                "Cesium asset ID is required to render a profile. This will be a required argument in the future."
                FutureWarning,
                stacklevel=2,
            )
```

bindings/python/tools/python/ostk/astrodynamics/viewer.py
```diff
@@ -198,6 +199,13 @@ def add_profile(
             Viewer: The Viewer.
         """
 
+        if cesium_asset_id is None:
+            warnings.warn(
+                "Cesium asset ID is required to render a profile. This will be a required argument in the future."
```
vishwa2710:
```suggestion
                "Cesium asset ID is required to render a profile. This will be a required argument in the future.",
```

apaletta3 (approved):

PR #668, opened 2026-04-23
fix: access is defined behavior

apaletta3:
I realized that the isDefined behavior for accesses didn't really make sense in its current implementation (it didn't correctly reflect the validity of `Partial` accesses, so I fixed it by:
- Deprecating the `Undefined` access type (to be removed in a future version) - we don't need this enum, we can simply improved the logic is the `isDefined` function to understand whether an access is defined or not
- Improved the logic in the `isDefined` function to:
	- mark a `Complete` access as defined when its AOS, LOS, TCA, and maxEl are all defined
	- mark a `Partial` access as defined only when its AOS and LOS are defined, it doesn't need a maxEl or TCA, because our access generation logic can produce these kinds of accesses (those that are clipped by the access analysis timespan)
- Split the check in the access contrustor to:
	- Raise for `Partial` Access' whose LOS is before AOS
	- Only raise for `Complete` Access' whose LOS is before TCA or TCA is before AOS

## Summary by CodeRabbit

* **Deprecations**
  * The "Undefined" access type is deprecated; use Complete or Partial instead.

* **Behavior Changes**
  * Validation is now type-specific: Complete requires all time fields and max elevation; Partial requires only acquisition and loss times.
  * Undefined cases now emit a deprecation warning and are treated as Partial.

* **Documentation**
  * Updated docstrings to reflect deprecation and allowed values.

* **Tests**
  * Expanded tests to assert type-specific validation and clearer error messages.

### DIFF
--- a/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Access.cpp
+++ b/bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Access.cpp
@@ -35,7 +35,9 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Access(pybind11::module& aModule)
         )
 
             .value(
-                "Undefined", Access::Type::Undefined, "Undefined (Deprecated, Accesses can only be Complete or Partial.)"
+                "Undefined",
+                Access::Type::Undefined,
+                "Undefined (Deprecated, Accesses can only be Complete or Partial.)"
             )
             .value("Complete", Access::Type::Complete, "Complete")
             .value("Partial", Access::Type::Partial, "Partial")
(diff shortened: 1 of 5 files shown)


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Access.cpp
```diff
@@ -34,7 +34,9 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Access(pybind11::module& aModule)
             )doc"
         )
 
-            .value("Undefined", Access::Type::Undefined, "Undefined")
+            .value(
+                "Undefined", Access::Type::Undefined, "Undefined (Deprecated) Accesses can only be Complete or Partial."
```
vishwa2710:
```suggestion
                "Undefined", Access::Type::Undefined, "Undefined (Deprecated, Accesses can only be Complete or Partial.)"
```
Ultra nit:

vishwa2710 (approved):

PR #670, opened 2026-04-29
fix: analytical gradient calculations in QLaw

vishwa2710:
Fix avoid catastrophic cancellation issues that occur at lower eccentricities when calculating the analytical gradient.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced numerical stability in guidance law computations by addressing catastrophic cancellation errors that previously occurred at low eccentricity values, significantly improving calculation accuracy and overall system reliability.
  * Improved the precision of analytical derivative calculations through refined numerical formulations to prevent floating-point precision loss in critical orbital mechanics edge cases.


apaletta3 (approved):
Approving since @vishwa2710  needs this change to get around solving issues he's facing - but he will add validation tests (instead of unit tests) for this solving scenario in a follow up MR

PR #673, opened 2026-04-30
fix: revolution number computation

apaletta3:
Fixes the off-by-one bug in the `PropagatedModel.calculateRevolutionNumberAt()` method.

Bug: The break check uses strict inequalities (>, <), but it is performed before revolutionNumber += durationSign and before currentInstant += durationSign * orbitalPeriod for the current iteration. As a result, when the loop reaches a currentInstant that is exactly equal to requesting instant, the predicate is still false, so the code increments the revolution counter one extra time and only breaks on the next iteration after overshooting by a full orbital period.
Result: 
- When asking for a rev number before epoch, you would get the true rev number minus one
- When asking for a rev number before epoch, you would get the true rev number plus one

New logic:
- Start from the epoch state
- compute the orbital period, jump back or forwards by one revolution
- break with the current rev number of we are past the requested instant
- otherwise increment/decrement the rev number

Note:
- Additionally, I also avoid repropagating all the way from the epoch state to the currentState each time the loop runs, because that is computationally slower than just propagating from the previously propagated state. 
- I also generalized the logic in the Kepler model to use the `getRevlutionNumberAtEpoch()` instead of assuming it is always 1, even though it is hardcoded to that

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed revolution-number origin and epoch handling so orbital revolution counts are consistent across models and propagation methods, including edge cases near node crossings.
  * Updated propagation logic to compute periods from the current orbit state for more reliable stepping.

* **Tests**
  * Expanded and refactored tests to use period-based validation, cover edge cases, and assert valid integer results.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler.cpp
@@ -128,7 +128,7 @@ Integer Kepler::getRevolutionNumberAtEpoch() const
         throw ostk::core::error::runtime::Undefined("Kepler");
     }
 
-    return 1;  // [TBI] With param
+    return 1;
 }
 
 Derived Kepler::getGravitationalParameter() const
@@ -233,17 +233,20 @@ Integer Kepler::calculateRevolutionNumberAt(const Instant& anInstant) const
     switch (perturbationType_)
     {
         case Kepler::PerturbationType::None:
-            return Kepler::CalculateNoneRevolutionNumberAt(coe_, epoch_, gravitationalParameter_, anInstant);
+            return Kepler::CalculateNoneRevolutionNumberAt(coe_, epoch_, gravitationalParameter_, anInstant) +
+                   this->getRevolutionNumberAtEpoch();
 
         case Kepler::PerturbationType::J2:
             return Kepler::CalculateJ2RevolutionNumberAt(
-                coe_, epoch_, gravitationalParameter_, anInstant, equatorialRadius_, j2_
-            );
+                       coe_, epoch_, gravitationalParameter_, anInstant, equatorialRadius_, j2_
+                   ) +
+                   this->getRevolutionNumberAtEpoch();
 
         case Kepler::PerturbationType::J4:
             return Kepler::CalculateJ4RevolutionNumberAt(
-                coe_, epoch_, gravitationalParameter_, anInstant, equatorialRadius_, j2_, j4_
-            );
+                       coe_, epoch_, gravitationalParameter_, anInstant, equatorialRadius_, j2_, j4_
+                   ) +
+                   this->getRevolutionNumberAtEpoch();
 
         default:
             throw ostk::core::error::runtime::Wrong("Perturbation type", StringFromPerturbationType(perturbationType_));
@@ -421,7 +424,7 @@ Integer Kepler::CalculateNoneRevolutionNumberAt(
 
     const Duration durationFromEpoch = Duration::Between(anEpoch, anInstant);
 
-    return (durationFromEpoch.inSeconds() / orbitalPeriod.inSeconds()).floor() + 1;
+    return (durationFromEpoch.inSeconds() / orbitalPeriod.inSeconds()).floor();
 }
 
 State Kepler::CalculateJ2StateAt(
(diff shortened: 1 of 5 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler.cpp
```diff
@@ -233,17 +233,20 @@ Integer Kepler::calculateRevolutionNumberAt(const Instant& anInstant) const
     switch (perturbationType_)
     {
         case Kepler::PerturbationType::None:
-            return Kepler::CalculateNoneRevolutionNumberAt(coe_, epoch_, gravitationalParameter_, anInstant);
+            return Kepler::CalculateNoneRevolutionNumberAt(coe_, epoch_, gravitationalParameter_, anInstant) +
```
apaletta3, line 236: `Kepler::CalculateNoneRevolutionNumberAt` is a static method and so can't access `this->getRevolutionNumberAtEpoch()`, hence why I add the value here

Also `Kepler::CalculateNoneRevolutionNumberAt` is a private method, so behavior doesn't change for the user here

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
```diff
@@ -319,16 +310,6 @@ Integer Propagated::calculateRevolutionNumberAt(const Instant& anInstant) const
...
-
-        // Propagate for duration of this orbital period
-        const State currentState = propagator_.calculateStateAt(
-            cachedStateArray_[0], cachedStateArray_[0].accessInstant() + (durationSign * orbitalPeriod)
-        );
-
-        // Update the current instant position and velocity coordinates
-        currentPositionCoordinates = currentState.getPosition().getCoordinates();
-        currentVelocityCoordinates = currentState.getVelocity().getCoordinates();
-        currentInstant += durationSign * orbitalPeriod;
```
apaletta3, line 331: the issue with the previous logic was that `currentInstant` wasn't updated before the `revolutionNumber` was incremented/decremented

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
```diff
@@ -272,44 +275,32 @@ Integer Propagated::calculateRevolutionNumberAt(const Instant& anInstant) const
...
     // Determine whether to count revolution numbers in forwards or backwards time and return function if duration is 0
-    Instant currentInstant = cachedStateArray_[0].getInstant();
-    const double durationInSecs = (anInstant - currentInstant).inSeconds();
-    if (durationInSecs == 0.0)
-    {
-        return 1;
-    }
-    Integer durationSign = (durationInSecs > 0) - (durationInSecs < 0);
+    const State stateAtEpoch = this->calculateStateAt(this->getEpoch());
+    const Integer durationSign = (anInstant > this->getEpoch()) ? 1 : -1;
```
alex-liang3, line 300: nit: why not use the `Sign` type?

alex-liang3 (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
```diff
@@ -272,44 +275,32 @@ Integer Propagated::calculateRevolutionNumberAt(const Instant& anInstant) const
...
-            -gravitationalParameter_SI * currentPositionCoordinates.norm() /
-            (currentPositionCoordinates.norm() * std::pow(currentVelocityCoordinates.norm(), 2) -
-             2.0 * gravitationalParameter_SI);
-        const Duration orbitalPeriod =
-            Duration::Seconds(Real::TwoPi() * std::sqrt(std::pow(semiMajorAxis, 3) / gravitationalParameter_SI));
+        // Calculate orbital period at current state
+        const COE coe = COE::Cartesian(
+            {currentState.getPosition(), currentState.getVelocity()}, Earth::Spherical.gravitationalParameter_
+        );  // (Spherical earth has the most modern value which is the correct one)
+        const Duration orbitalPeriod = coe.getOrbitalPeriod(Earth::Spherical.gravitationalParameter_);
```
vishwa2710:
```suggestion
        TBI: allow user to provide gravitational parameter
        const Duration orbitalPeriod = coe.getOrbitalPeriod(Earth::Spherical.gravitationalParameter_);
```

vishwa2710 (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
apaletta3: not sure I'm able to increment smoothly with it

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
apaletta3: I'll do one better - get the grav param from the propagator_

PR #674, opened 2026-04-30
fix: propagated revolution number computation

apaletta3:
The previous "fix" that I put out still suffers from the same problem as before: namely that it was propagating one whole orbital period before checking if it had encountered a new pass (aka crossed the ascending node) or not. 

Since the propagated model's revolution number computation relies on numerical integration, the only way to properly do it, is to check if you have crossed the next ascending node before incrementing the revolution number counter, using a condition propagator.

I have buffed up the test to actually query the revolution number in non-orbital period increments starting from the epoch to ensure that we don't run into the previous issue again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved revolution-number calculations for propagated orbital models by advancing directly to relevant node crossings, improving accuracy and robustness around edge cases.
  * Updated Kepler-family revolution-number logic to use signed node-crossing counts instead of simple period-based flooring.
* **Tests**
  * Refreshed revolution-number expectations with finer-grained timing offsets and revised node-related edge-case scenarios to better validate transitions.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
@@ -12,6 +12,7 @@
 #include <OpenSpaceToolkit/Physics/Unit/Time.hpp>
 
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/CentralBodyGravity.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/EventCondition/RealCondition.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp>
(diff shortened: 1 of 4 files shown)


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.cpp
```diff
@@ -282,48 +264,48 @@ Integer Propagated::calculateRevolutionNumberAt(const Instant& anInstant) const
...
-            break;
-        if (durationSign.isNegative() && currentInstant < anInstant)
+        // anInstant has been reached before the next ascending node was found, return the current revolution number
+        if (!solution.conditionIsSatisfied)
+        {
             break;
+        }
+
+        // Given that the condition is satisfied, the root solver should have converged
+        if (!solution.rootSolverHasConverged)
```
apaletta3, line 300: @vishwa2710 can you double check my understanding? If the `solution.conditionIsSatisfied == true` (only way to get to this point in the loop), the `solution.rootSolverHasConverged == false` would indicate something having gone horribly wrong, right?

PR #683, opened 2026-06-03
feat: interpolators per coordinate subset for tabulated models

vishwa2710:
- Tabulated trajectory, orbit, and profile models now support per-coordinate-subset interpolation configuration 
- Default static constructor using sensible per-subset interpolation choices

## Release Notes

* **New Features**
  * Added support for per-coordinate-subset interpolation type configuration in tabulated trajectory, orbit, and flight profile models, enabling distinct interpolation methods for position, velocity, and other coordinate subsets.
  * Introduced `Default()` static factory methods for convenient model construction using optimized interpolation defaults.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
@@ -4,6 +4,7 @@
 #define __OpenSpaceToolkit_Astrodynamics_Trajectory_Model_Tabulated__
 
 #include <OpenSpaceToolkit/Core/Container/Array.hpp>
+#include <OpenSpaceToolkit/Core/Container/Map.hpp>
 #include <OpenSpaceToolkit/Core/Container/Pair.hpp>
 #include <OpenSpaceToolkit/Core/FileSystem/File.hpp>
 #include <OpenSpaceToolkit/Core/Type/Index.hpp>
@@ -18,6 +19,7 @@
 
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/Model.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset.hpp>
 
 namespace ostk
 {
@@ -29,6 +31,7 @@ namespace model
 {
 
 using ostk::core::container::Array;
+using ostk::core::container::Map;
 using ostk::core::container::Pair;
 using ostk::core::filesystem::File;
 using ostk::core::type::Index;
@@ -45,6 +48,7 @@ using ostk::physics::time::Scale;
 
 using ostk::astrodynamics::trajectory::Model;
 using ostk::astrodynamics::trajectory::State;
+using ostk::astrodynamics::trajectory::state::CoordinateSubset;
 
 #define DEFAULT_TABULATED_TRAJECTORY_INTERPOLATION_TYPE Interpolator::Type::Linear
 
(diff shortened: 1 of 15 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
```diff
@@ -134,6 +161,9 @@ class Tabulated : public virtual Model
     ///     Interpolator::Type type = tabulated.getInterpolationType();
     /// @endcode
     ///
+    /// @note When the model was constructed with per-coordinate-subset interpolation types, this returns the
+    /// interpolation type of the first coordinate.
```
alex-liang3:
```suggestion
    /// interpolation type of the first coordinate subset.
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.cpp
```diff
@@ -229,6 +288,36 @@ void Tabulated::print(std::ostream& anOutputStream, bool displayDecorator) const
...
+    using ostk::astrodynamics::trajectory::state::coordinatesubset::AttitudeQuaternion;
+    using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianAcceleration;
+    using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianPosition;
+    using ostk::astrodynamics::trajectory::state::coordinatesubset::CartesianVelocity;
+
+    static const Map<Shared<const CoordinateSubset>, Interpolator::Type> defaultInterpolationTypes = {
+        {CartesianPosition::Default(), Interpolator::Type::BarycentricRational},
+        {CartesianVelocity::Default(), Interpolator::Type::BarycentricRational},
+        {CartesianAcceleration::Default(), Interpolator::Type::BarycentricRational},
+        {AttitudeQuaternion::Default(), Interpolator::Type::BarycentricRational},
```
alex-liang3: i worry this could make things more confusing:
- in reality, only the `FlightProfile` Tabulated model uses SLERP, since it's done in its `calculateStateAt()`.
- now, not only are we suggesting that the other Tabulated forms "know" how to interpolate AttitudeQuaternions, but they would still use the wrong form (and there's no indication of this in the `TabulatedOrbit` and `TabulatedTrajectory` docs)

should we just create an `Interpolator::Type::SLERP`?

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
```diff
@@ -283,6 +295,11 @@ void Tabulated::print(std::ostream& anOutputStream, bool displayDecorator) const
     displayDecorator ? ostk::core::utils::Print::Footer(anOutputStream) : void();
 }
 
+Tabulated Tabulated::Default(const Array<State>& aStateArray)
+{
+    return Tabulated(aStateArray, TabulatedTrajectory::DefaultInterpolationTypes());
```
alex-liang3: although `TabulatedTrajectory` was already included, it seemed to be completely unused previously. Are we ok with introducing this relation, given `Trajectory` and `Flight/Profile` fall under unrelated namespaces?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.cpp
vishwa2710: Yeah I agree. I think for now, I can remove this from the map.
Separate MR, we can add a SLERP interpolator type.
And then update the Flight Profile to leverage that, and probably condense all the logic into the Tabulated (Trajectory) model only.
What do you think?

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile/Model/Tabulated.cpp
vishwa2710: Hmmmmm fair point, I think it's better to just specify the desired coordinate subsets here explicitly.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.cpp
alex-liang3: yep that sounds good!

alex-liang3 (approved):

PR #685, opened 2026-06-09
feat: add utility methods to COE

vishwa2710:
Adds some utility methods to COE (that existed in the getters but not in the static compute methods).
Also adds the AoL angular rate computation method.

## Summary by CodeRabbit

* **New Features**
  * Added static calculation methods for orbital characteristics including periapsis and apoapsis radius, mean motion, orbital period, nodal precession rate, and argument-of-latitude angular rate with J2 perturbation effects.

* **Refactor**
  * Refactored existing orbital calculation methods to delegate to new centralized static helper functions, improving consistency and maintainability.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
@@ -221,6 +221,16 @@ class COE
         const Derived& aGravitationalParameter, const Length& anEquatorialRadius, const Real& aJ2Parameter
     ) const;
 
+    /// @brief Get the along-track angular rate (argument of latitude angular rate) including J2 perturbations
+    ///
+    /// @param aGravitationalParameter A gravitational parameter
+    /// @param anEquatorialRadius An equatorial radius
+    /// @param aJ2Parameter A J2 parameter
+    /// @return Argument of latitude angular rate
+    Derived getArgumentOfLatitudeAngularRate(
+        const Derived& aGravitationalParameter, const Length& anEquatorialRadius, const Real& aJ2Parameter
+    ) const;
+
     /// @brief Get Orbital period
     ///
     /// @param aGravitationalParameter A gravitational parameter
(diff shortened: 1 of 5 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
```diff
@@ -442,6 +452,72 @@ class COE
...
+    /// @param anEccentricity Eccentricity of the orbit.
+    /// @return Apoapsis radius in meters.
+    static Real ComputeApoapsisRadius(const Real& aSemiMajorAxis, const Real& anEccentricity);
+
+    /// @brief Compute the mean motion of the orbit.
+    ///
+    /// @param aSemiMajorAxis Semi-major axis of the orbit in meters.
+    /// @param aGravitationalParameter Gravitational parameter.
+    /// @return Mean motion in radians per second.
+    static Real ComputeMeanMotion(const Real& aSemiMajorAxis, const Derived& aGravitationalParameter);
```
apaletta3, line 474: are you sure we don't have this already somewhere?

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
```diff
@@ -442,6 +452,72 @@ class COE
...
+    /// Sums the unperturbed mean motion with the secular J2 rates of the mean anomaly and the argument of periapsis.
+    ///
+    /// @param aSemiMajorAxis Semi-major axis of the orbit in meters.
+    /// @param anEccentricity Eccentricity of the orbit.
+    /// @param anInclination Inclination of the orbit in radians.
+    /// @param aGravitationalParameter Gravitational parameter.
+    /// @param anEquatorialRadius Equatorial radius of the central body in meters.
+    /// @param aJ2Parameter Second zonal harmonic coefficient.
+    /// @return Argument of latitude angular rate in radians per second.
+    static Real ComputeArgumentOfLatitudeAngularRate(
```
apaletta3, line 512: I think this may be more generic
```suggestion
    static Real ComputeAlongTrackRate(
```

apaletta3 (commented):
Suggestion for better naming - but otherwise ✅

apaletta3 (approved):

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
vishwa2710: Since the along track angle is a generic term, and this computes the actual argument of latitude rate (which differs from mean/true anomaly). I think it's better to name it this.

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.hpp
vishwa2710: We have the `instance` method `coe.getMeanMotion`, but I think it's worth having static methods for all of these so users don't have to create a COE class with bogus values just to get a specific value.

PR #687, opened 2026-06-18
feat: add Propagated to Tabulated method

alex-liang3:
Add a helper to convert a Propagated orbit model into a Tabulated one, so that it can be "reused" more easily for a given timespan. 

## Release Notes

* **New Features**
  * Added `to_tabulated()` method to convert Propagated orbit models to Tabulated orbit models by sampling at specified time instants.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.hpp
@@ -10,6 +10,7 @@
 #include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
 
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model.hpp>
+#include <OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Tabulated.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/State.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/State/NumericalSolver.hpp>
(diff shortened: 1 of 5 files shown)


apaletta3 (approved):

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.hpp
```diff
@@ -177,6 +181,12 @@ class Propagated : public ostk::astrodynamics::trajectory::orbit::Model
     /// @param (optional) displayDecorators If true, display decorators
     virtual void print(std::ostream& anOutputStream, bool displayDecorator = true) const override;
 
+    /// @brief Compute the states at the given instants and construct a Tabulated orbit model from them.
+    ///
+    /// @param anInstantArray An array of at least 4 instants to compute the states at.
+    /// @return The Tabulated orbit model
+    Tabulated toTabulated(const Array<Instant>& anInstantArray) const;
```
vishwa2710, line 188: For flexibility, would it be worth allowing user specified map of coordinate subsets to interpolator types?
```
    Tabulated toTabulated(const Array<Instant>& anInstantArray, const Map<CoordinateSubset, InterpolatorType>& interpolatorTypeMapping = {}) const;
```
If it's empty, we just use the default, otherwise we apply the user specified mappings?

In the case that user specifies a coordinate subset that is not a part of the states, we raise an error.
But for a reduced list, we only extract those coordinate subsets post propagating and created a reduced set?

vishwa2710 (commented):
Just 1 comment on improving the interface

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Propagated.hpp
alex-liang3: idk, this MR was just supposed to be a one-liner convenience function

At the very least, the behaviour you've specified here is different to your new Tabulated constructor - i think this should be consistent. 

If we really want the ability to specify an interpolator map at this interface, then I think it should just be directly passed down to the Tabulated constructor and validated there

vishwa2710: Just so I understand correctly, you mean to keep the behaviour consistent with the constructor, which requires a coordinate subset per coordinate that is within the state yeah?

I think it's fair then that you would want to be consistent with the behaviour of the Tabulated constructor.

PR #688, opened 2026-06-24
feat: add output frame option for Tabulated models

vishwa2710:
- minor performance improvements to the access generation 
- output frame option for tabulated models, defaults to GCRF for backwards compatibility
- when using ITRF for tabulated orbits + access generator, greatly increases performance

## Summary by CodeRabbit

* **New Features**
  * Tabulated trajectories now support choosing an output reference frame when constructing or creating default models.
  * Added a way to view the model’s configured output frame.
  * Access calculations and benchmarks now support more frame-aware trajectory handling.

* **Bug Fixes**
  * Interpolated states now consistently return in the selected output frame.
  * Improved accuracy for access, elevation, and closest-approach calculations by using the correct instant-specific frame transformations.
  * Models with different output frames are now treated as different.


benchmark/OpenSpaceToolkit/Astrodynamics/Access.benchmark.cpp
```diff
@@ -91,5 +118,149 @@ static void benchmark001(benchmark::State& state)
...
+    const LLA groundStationLla = {Angle::Degrees(-45.0), Angle::Degrees(-170.0), Length::Meters(5.0)};
+    const Position groundStationPosition = Position::Meters(
+        groundStationLla.toCartesian(Earth::EGM2008.equatorialRadius_, Earth::EGM2008.flattening_), Frame::ITRF()
+    );
+
+    return AccessTarget::FromPosition(visibilityCriterion, groundStationPosition);
+}
+
+// A spread of ground stations across the globe, each with an above-the-horizon elevation criterion.
+static Array<AccessTarget> MakeElevationTargets(const Index aTargetCount)
```
apaletta3, line 177: Align the naming here between MakeGroundStationTarget and MakeElevationTargets since they are both making AccessTargets

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
apaletta3, line 210: I think you should add an equivalent method to get the interpolation type map `Map<Shared<const CoordinateSubset>, Interpolator::Type>` (can be done in a separate MR)

I know you're not storing that variable, so are you even able to extract it later on?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.cpp
```diff
@@ -132,7 +162,12 @@ bool Tabulated::operator==(const Tabulated& aTabulatedModel) const
         }
     }
 
-    return firstState_ == aTabulatedModel.getFirstState() && lastState_ == aTabulatedModel.getLastState();
+    const bool framesEqual = (outputFrameSPtr_ == aTabulatedModel.outputFrameSPtr_) ||
+                             ((outputFrameSPtr_ != nullptr) && (aTabulatedModel.outputFrameSPtr_ != nullptr) &&
+                              ((*outputFrameSPtr_) == (*aTabulatedModel.outputFrameSPtr_)));
```
apaletta3: why are we comparing if both the pointer or the frame are equal? Shouldn't we only worry about if the frames are equal?

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -472,7 +474,10 @@ Array<Array<Access>> Generator::computeAccessesForFixedTargets(
     {
         const MatrixXd dx = (-fromPositionCoordinates_ITRF).colwise() + aToPositionCoordinates_ITRF;
 
-        // TBI: See if this can be vectorized
+        // Apply each target's 3x3 SEZ rotation to its dx column. A vectorized variant (Eigen strided slicing over
+        // targets) was benchmarked and is not worthwhile: the per-target cost is dominated by the azimuth/elevation
+        // trig below, not this matmul, and the strided gather offsets any SIMD gain (slightly slower at low target
+        // counts, a wash at high counts). The contiguous-block loop is kept.
```
apaletta3: Nit: Can you clarify this docstring? Not sure what half these words mean 😂

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -648,26 +653,29 @@ Array<Array<Access>> Generator::computeAccessesForFixedTargets(
...
 
-        if (this->getStateFilter())
+        if (stateFilter)
         {
             for (Index i = 0; i < targetCount; ++i)
             {
                 const State fromState = someAccessTargets[i].accessTrajectory().getStateAt(instant);
 
-                inAccess(i) = inAccess(i) && this->getStateFilter()(fromState, toTrajectoryState);
+                inAccess(i) = inAccess(i) && stateFilter(fromState, toTrajectoryState);
```
apaletta3: Does this actually make a difference? The overhead of isDefined() seems pretty small, no?

Can you do this with the same effect (still accessing by reference, not value, as desired)
```suggestion
for (Index index = 0; index < instants.getSize(); ++index)
{
const Instant& instant = instants[index];

const State toTrajectoryState = aToTrajectory.getStateAt(instant);

        // calculate target to satellite vector in ITRF (transform only the position, not the whole state)
const Vector3d toPositionCoordinates_ITRF =
            toTrajectoryState.getPosition().inFrame(Frame::ITRF(), instant).getCoordinates();

// check if satellite is in access
auto inAccess = visibilityCriterionFilter(fromPositionCoordinates_ITRF, toPositionCoordinates_ITRF, instant);

        if (this->stateFilter_)
{
for (Index i = 0; i < targetCount; ++i)
{
const State fromState = someAccessTargets[i].accessTrajectory().getStateAt(instant);

                inAccess(i) = inAccess(i) && this->stateFilter_(fromState, toTrajectoryState);
```

apaletta3 (approved):
Nice! left some small comments

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
vishwa2710: Yeah that one will be deprecated, and in the future you can just get the interpolators and get the type from that rather than storing the type.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.cpp
vishwa2710: Nice catch!

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
vishwa2710: It's super minor but just better practice to not do repeated actions. That should be equivalent though. Updated!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model/Tabulated.hpp
vishwa2710: Will put out a separate MR for this.

PR #692, opened 2026-07-02
feat: add orbital element condition

vishwa2710:
Instead of having COECondition, BLMShortCondition, BLMLongCondition etc, we can have one set of static constructors that create the appropriate Real/Angular conditions. This supercedes Pau's BLMShortCondition MR if you agree.

## Summary by CodeRabbit

* **New Features**
  * Added a unified orbital-element-based event condition in both C++ and Python, with selectable orbital theories and support for real and angular elements, including range-based checks for angular targets.
* **Deprecation**
  * Marked older orbital-element condition APIs as deprecated in both C++ and Python, directing users to the new orbital-element condition with the equivalent theory.
* **Tests**
  * Added C++ and Python test coverage for the new condition across supported theories and both single-value and range-based angular target constructions.


phc1990 (approved):
Great 👍 

This is def. more scalable

Only added a commit on top that (for now) removes the agnostic `Construct` methods, as I don't think there is a use-case for them (yet), and they added complexity around the allowed combinations.

Can always add them if we need them.

PR #693, opened 2026-07-08
feat: add Alpha-5 support for TLE satellite numbers

maxnqt:
Space-Track now issues TLEs with [Alpha-5](https://www.space-track.org/documentation#tle-alpha5) satellite numbers for catalog IDs above 99,999. The first character of the 5-digit field becomes a letter (e.g. A5544 -> 105544). We were rejecting those on parse and couldn't round-trip them on construct.

This PR adds Alpha-5 encode/decode for the full range (0-339,999), exposes `getSatelliteNumberString()` for the raw TLE field, and wires it through Python. SGP4 propagation is also fixed with a workaround: `libsgp4` raises on the letter in the NORAD field but doesn't use it in the math/propagation, so we substitute a numeric placeholder only when handing lines to `libsgp4`.

**Testing**

- C++ TLE tests: parse, construct, set, checksum with Alpha-5 fields
- C++ SGP4 test: Alpha-5 TLE propagates identically to the numeric equivalent

## Summary by CodeRabbit

- **New Features**
  - Added support for Alpha-5 encoded satellite numbers in TLE parsing and creation (up to 339999).
  - Exposed access to both decoded satellite numbers and the original raw Alpha-5 field (including `A5544`/`Z9999`).
  - Enhanced SGP4 propagation to work with Alpha-5 encoded TLEs.

- **Bug Fixes**
  - Improved handling of malformed, negative, and out-of-range satellite-number fields, with clearer error behavior.
  - Updated TLE line checksum handling to keep generated lines consistent after satellite-number changes.

- **Documentation**
  - Clarified the difference between decoded vs raw satellite-number values in user-facing API docs.


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.hpp
```diff
@@ -154,6 +154,14 @@ class TLE
     /// @return Satellite number
     Integer getSatelliteNumber() const;
 
+    /// @brief Get satellite number field as it appears in the TLE
+    ///
+    ///        For satellite numbers above 99999 this returns the raw Alpha-5 encoded field
+    ///        (e.g. "A5544"), rather than the decoded integer.
+    ///
+    /// @return Satellite number field
+    String getSatelliteNumberString() const;
```
apaletta3: perhaps it should be something like this for clarity?
```suggestion
    /// @return Satellite number field
    String getSatelliteAlpha5Number() const;
```

apaletta3 (commented):
comment on the naming - i think we should just pre-pend Alpha5 to SatelliteNumber to indicate that this is an Alpha5 version of the number

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.hpp
maxnqt: I agree with Alex's sentiment that `getSatelliteNumberString()` is already pretty "good." Here are some other ideas though:

1. `getSatelliteNumberRaw()`: "raw" is idiomatic in parsing contexts for "as it appears in the source"
2. `getSatelliteNumberEncoded()`: Alpha-5 is technically an encoding; but not all of them are actually encoded

maxnqt:
```suggestion
    /// @return Satellite number field
    String getSatelliteNumberRaw() const;
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.hpp
```diff
@@ -151,9 +151,22 @@ class TLE
 
     /// @brief Get satellite number
     ///
+    ///        For satellite numbers above 99999 this returns the decoded integer and not
+    ///        the raw Alpha-5 encoded field (e.g. 105544 instead of "A5544"). If you need
+    ///        the raw Alpha-5 encoded field or the string representation that matches the
+    ///        content of the TLE, use getSatelliteNumberString() instead.
```
alex-liang3: add this to the python binding docstring too

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +855,74 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
     return Real::Parse(string);
 }
 
+String TLE::SatelliteNumberToAlpha5(const Integer& aSatelliteNumber)
+{
+    static const String alpha5Letters = "ABCDEFGHJKLMNPQRSTUVWXYZ";
```
alex-liang3, line 866: nit: add a link to the spec here and mention that `I` and `O` are skipped

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -598,7 +603,7 @@ TLE TLE::Construct(
     using ostk::physics::unit::Derived;
     using ostk::physics::unit::Time;
 
-    if (aSatelliteNumber > 99999)
+    if (aSatelliteNumber > 399999)
```
alex-liang3: shouldn't it be 
```suggestion
    if (aSatelliteNumber > 339999)
```
?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +855,74 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
     return Real::Parse(string);
 }
 
+String TLE::SatelliteNumberToAlpha5(const Integer& aSatelliteNumber)
+{
+    static const String alpha5Letters = "ABCDEFGHJKLMNPQRSTUVWXYZ";
+
+    if (aSatelliteNumber < 0 || aSatelliteNumber > 339999)
```
alex-liang3: maybe best to define this limit as a const somewhere

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +855,74 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
...
+    }
+
+    if (aSatelliteNumber <= 99999)
+    {
+        // Standard: right-justify in 5 characters
+        const String digits = aSatelliteNumber.toString();
+        return String::Replicate(" ", 5 - digits.getLength()) + digits;
+    }
+
+    // Alpha-5: leading letter encodes the first one or two digits
```
alex-liang3:
```suggestion
    // Alpha-5: leading letter encodes the first two digits
```
maps `10xxxx` to `33xxxx` no?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +855,74 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
...
+    const auto letterIndex = alpha5Letters.find(firstChar);
+
+    if (letterIndex == std::string::npos)
+    {
+        throw ostk::core::error::runtime::Wrong(
+            String::Format("Invalid Alpha-5 character '{}' in TLE satellite number field", firstChar)
+        );
+    }
+
+    if (field.getLength() < 5)
```
alex-liang3: i'd probably be stricter here
```suggestion
    if (field.getLength() != 5)
```

test/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.test.cpp
```diff
@@ -142,6 +150,67 @@ TEST(OpenSpaceToolkit_Astrodynamics_Trajectory_Orbit_Model_SGP4_TLE, GetSatellit
...
+    // Alpha-5 satellite number is returned as the raw field
+    {
+        const String firstLine = "1 A5544U 98067A   08264.51782528 -.00002182  00000-0 -11606-4 0  2925";
+        const String secondLine = "2 A5544  51.6416 247.4627 0006703 130.5360 325.0288 15.72125391563535";
+
+        const TLE tle(firstLine, secondLine);
+
+        EXPECT_EQ("A5544", tle.getSatelliteNumberString());
+        EXPECT_EQ(105544, tle.getSatelliteNumber());
+    }
```
alex-liang3, line 223: nit: add test for `Zxxxx` to confirm `I` and `O` skipped in the count

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.hpp
alex-liang3: +1 for `getSatelliteNumberRaw()` (or `getRawSatelliteNumber()`)

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.hpp
```diff
@@ -453,12 +466,17 @@ class TLE
     /// @return Checksum
     static Integer GenerateChecksum(const String& aLine);
 
+    /// @brief Maximum TLE satellite number (Alpha-5 encoding limit)
+    static const Integer MaximumSatelliteNumber;
```
apaletta3: not sure this needs to be public and exposed to the user?

bindings/python/test/trajectory/orbit/models/sgp4/test_tle.py
```diff
@@ -30,6 +30,38 @@ def tle() -> TLE:
...
+        second_line="2        51.6447  64.7824 0005971  73.1467  36.4366 15.53848234128316",
+    )
+
+
+@pytest.fixture
+def tle_with_short_alpha5_satellite_number_field() -> TLE:
+    return TLE(
+        first_line="1 A554 U 98067A   18231.17878740  .00000187  00000-0  10196-4 0  9998",
+        second_line="2 A554   51.6447  64.7824 0005971  73.1467  36.4366 15.53848234128310",
+    )
```
apaletta3, line 56: these don't raise on instantiation of the fixture itself?

bindings/python/test/trajectory/orbit/models/sgp4/test_tle.py
```diff
@@ -319,6 +385,42 @@ def test_construct_without_satellite_name(self, tle: TLE):
         assert constructed_tle.get_first_line() == tle.get_first_line()
         assert constructed_tle.get_second_line() == tle.get_second_line()
 
+    def test_construct_with_alpha5_satellite_number(
```
apaletta3: we usually don't test cpp logic in the python tests (only test it in the cpp tests). Here we usually only test that the bindings themselves work and return the correct types. So I don't think `test_construct_with_alpha5_satellite_number` and `test_generate_checksum_with_alpha5_satellite_number` tests are necessary

bindings/python/test/trajectory/orbit/models/sgp4/test_tle.py
```diff
@@ -73,6 +105,33 @@ def test_get_second_line(self, tle: TLE):
...
+            tle_with_short_alpha5_satellite_number_field.get_satellite_number()
+
+    def test_get_raw_satellite_number(self, tle: TLE):
+        assert tle.get_raw_satellite_number() == "25544"
+
+    def test_get_raw_satellite_number_alpha5(self, tle: TLE):
+        tle.set_satellite_number(105544)
+
+        assert tle.get_raw_satellite_number() == "A5544"
+        assert tle.get_satellite_number() == 105544
```
apaletta3, line 131: ensure the return types are correct
```suggestion
        assert isinstance(tle.get_raw_satellite_number(), str)
        assert tle.get_raw_satellite_number() == "25544"

    def test_get_raw_satellite_number_alpha5(self, tle: TLE):
        tle.set_satellite_number(105544)

        assert isinstance(tle.get_raw_satellite_number(), str)
        assert tle.get_raw_satellite_number() == "A5544"
        assert isinstance(tle.get_raw_satellite_number(), int)
        assert tle.get_satellite_number() == 105544
```

bindings/python/test/trajectory/orbit/models/test_sgp4.py
```diff
@@ -189,3 +189,28 @@ def test_calculate_states_at(
         ]
         states = sgp4.calculate_states_at(instants)
         assert states is not None
+
+    def test_alpha5(
```
apaletta3: same comment here, don't think this test is necessary cause nothing at the SGP4 interface has changed

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -339,29 +352,25 @@ void TLE::setSatelliteNumber(const Integer& aSatelliteNumber)
...
     const Integer firstLineNewChecksum = TLE::GenerateChecksum(newIntermediateFirstLine);
     const Integer secondLineNewChecksum = TLE::GenerateChecksum(newIntermediateSecondLine);
 
     firstLine_ = newIntermediateFirstLine.getSubstring(0, 68) + firstLineNewChecksum.toString();
     secondLine_ = newIntermediateSecondLine.getSubstring(0, 68) + secondLineNewChecksum.toString();
-
-    if (!this->isDefined())
-    {
-        throw ostk::core::error::runtime::Undefined("TLE");
-    }
```
apaletta3, line 364: Why remove this check? You can put it at the top of the function if you'd like

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +859,80 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
...
+    }
+
+    // Alpha-5: leading letter encodes the first two digits
+    const Integer leadingDigits = aSatelliteNumber / 10000;  // e.g. 182931 -> 18
+    const Integer remainder = aSatelliteNumber % 10000;      // e.g. 182931 -> 2931
+
+    const String remainderString = remainder.toString();
+    const String paddedRemainder = String::Replicate("0", 4 - remainderString.getLength()) + remainderString;
+
+    const char letter = alpha5Letters[static_cast<int>(leadingDigits) - 10];  // 10 -> 'A', 11 -> 'B', ...
```
apaletta3: why use a char directly? We don't usually do that and prefer to use a safer String

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
```diff
@@ -850,6 +859,80 @@ Real TLE::ParseReal(const String& aString, bool isDecimalPointAssumed)
...
+
+    String field = aField;  // mutable copy
+    field.trim();           // mutates in place
+
+    if (field.isEmpty())
+    {
+        throw ostk::core::error::runtime::Wrong("TLE satellite number field is empty");
+    }
+
+    const char firstChar = field[0];
```
apaletta3, line 909: same comment as above

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4.cpp
```diff
@@ -33,6 +34,38 @@ using ostk::physics::time::Interval;
...
+    // value so libsgp4's "satellite numbers must match" check passes. Line length (69) and the
+    // leading '1'/'2' are preserved. libsgp4 does not validate the checksum, so the trailing
+    // checksum digit is left untouched.
+    static const String placeholder = "00000";
+
+    return aLine.getSubstring(0, 2) + placeholder + aLine.getSubstring(7, aLine.getLength() - 7);
+}
+
+}  // namespace
+
```
apaletta3: I think you may have accidentally defined this function in the global namespace. The common pattern in ostk is to make functions static methods of whatever class they relate to. In this case, it should be a private static method, as it does not need to be exposed to the user

apaletta3 (commented):
Nice stuff - a couple small comments to address and then we should be good

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
maxnqt: It's already at the top of the function 👍

bindings/python/test/trajectory/orbit/models/sgp4/test_tle.py
maxnqt: `CanParse` only checks:

- Both lines are exactly 69 characters
- The last character of each line is a valid integer
- The checksums match

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
maxnqt: I'm used to defining single letters as chars. Changed it though 👍

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/TLE.cpp
maxnqt: - `std::isdigit(...)` operates on a single character code; there's no String overload, so a 1-char String would just get unpacked back to a char to call it
- `alpha5Letters.find(...)` behaves identically with a char or a 1-char String, so no safety difference
- field[0] is also already safe here since `field.isEmpty()` is checked and throws just above

apaletta3 (approved):

PR #707, opened 2026-08-11
perf: simplify third body gravity calculation

vishwa2710:
In the case of a spherical third body gravity model, the field does not depend on the body's orientation, and so we can simplify the computation a bit by avoiding converting from GCRF -> Body frame -> GCRF.

This improves performance by about 5-10%. Nothing crazy.

## Summary by CodeRabbit

* **Performance**
  * Improved third-body gravity calculations for point-mass celestial bodies.
  * Preserved existing calculation behavior for non-point-mass bodies.

* **Bug Fixes**
  * Improved consistency and reliability of third-body acceleration results across supported gravity models.

* **Chores**
  * Updated the Python package compatibility constraint for the physics toolkit.

### DIFF
--- a/src/OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.cpp
+++ b/src/OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.cpp
@@ -1,7 +1,15 @@
 /// Apache License 2.0
 
+#include <cmath>
+
 #include <OpenSpaceToolkit/Core/Error.hpp>
 #include <OpenSpaceToolkit/Core/Utility.hpp>
+
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Earth.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Model.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Moon.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Spherical.hpp>
+#include <OpenSpaceToolkit/Physics/Environment/Gravitational/Sun.hpp>
 
 #include <OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.hpp>
 #include <OpenSpaceToolkit/Astrodynamics/Trajectory/State/CoordinateSubset/CartesianPosition.hpp>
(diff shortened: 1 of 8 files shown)


src/OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.cpp
```diff
@@ -118,6 +163,40 @@ void ThirdBodyGravity::print(std::ostream& anOutputStream, bool displayDecorator
     displayDecorator ? ostk::core::utils::Print::Footer(anOutputStream) : void();
 }
 
+bool ThirdBodyGravity::IsPointMassGravitationalModel(const Celestial& aCelestial)
```
phc1990: would it make sense for this to be a function in Celestial? 

or perhaps in the Gravitational/Model.hpp ? (👈 I personally think this is the right place)

this way the Earth/Sun/Moon/X or any model specific logic is "self contained"

we can still keep the private property in this ForceModel so that the function is only called once upon instantiation of it

but I feel like this logic belongs in physics

phc1990 (commented):
left comment on function placement

src/OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.cpp
```diff
@@ -118,6 +163,40 @@ void ThirdBodyGravity::print(std::ostream& anOutputStream, bool displayDecorator
...
+
+    // Earth supports higher-fidelity (spherical harmonics) models, so only its `Spherical` type qualifies.
+    // Earth cannot actually reach this class as a third body (see the constructor), but this is kept for
+    // completeness/defensiveness should that restriction ever be lifted.
+    if (const Earth* earthGravitationalModelPtr = dynamic_cast<const Earth*>(gravitationalModelSPtr.get()))
+    {
+        return earthGravitationalModelPtr->getType() == Earth::Type::Spherical;
+    }
+
+    return false;
```
apaletta3: Why is the Earth being checked twice?

apaletta3 (commented):
Is this change worth it? Given that the 5-10% performance improvement only applies to numerically propagating an Orbit with a spherical Earth model. If someone really wants propagation to return quickly at the spherical Earth gravity level of fidelity, they would probably use a Keplerian Model which would run much faster than a Propagated model. 
So the only use case where this would make a difference is for someone benchmarking accuracy/performance when using a numerical propagator with spherical earth gravity AND 3rd body perturbs, which is not a very common force model combination.
Thoughts?

vishwa2710:
> Is this change worth it? Given that the 5-10% performance improvement only applies to numerically propagating an Orbit with a spherical Earth model. If someone really wants propagation to return quickly at the spherical Earth gravity level of fidelity, they would probably use a Keplerian Model which would run much faster than a Propagated model. So the only use case where this would make a difference is for someone benchmarking accuracy/performance when using a numerical propagator with spherical earth gravity AND 3rd body perturbs, which is not a very common force model combination. Thoughts?

Fair question! The ~10% speed up applies to spherical **third body** models, (which is what the only existing Sun and Moon models are).

src/OpenSpaceToolkit/Astrodynamics/Dynamics/ThirdBodyGravity.cpp
apaletta3: @vishwa2710 one last comment from me ^

vishwa2710: I'm not sure what you mean? It's just being checked once?

apaletta3 (approved):

PR #708, opened 2026-08-14
feat: add option to observe dynamic contributions

vishwa2710:
Adds the ability to record the dynamics contributions when propagating. Unfortunately there's no way to save each individual dynamic without re-evaluating it (as we use Boost ODEInt and it evaluates the entire system of equations which is the sum of the dynamic contributions).
This is the next best thing. Open to suggestions on naming.

## Summary by CodeRabbit

* **New Features**
  * Added optional recording of per-dynamics contributions during propagation.
  * Added access to recorded integration instants and contribution data.
  * Added controls to enable, disable, and check contribution observation.
  * Added support for calculating individual dynamics contributions independently.

* **Bug Fixes**
  * Improved portability of reference data paths.

* **Tests**
  * Added comprehensive coverage for contribution recording, propagation directions, resets, and edge cases.

### DIFF
--- a/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
+++ b/include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
@@ -4,10 +4,12 @@
 #define __OpenSpaceToolkit_Astrodynamics_Trajectory_Propagator__
 
 #include <OpenSpaceToolkit/Core/Container/Array.hpp>
+#include <OpenSpaceToolkit/Core/Container/Map.hpp>
 #include <OpenSpaceToolkit/Core/Type/Shared.hpp>
 #include <OpenSpaceToolkit/Core/Type/Size.hpp>
 
 #include <OpenSpaceToolkit/Mathematics/CurveFitting/Interpolator.hpp>
+#include <OpenSpaceToolkit/Mathematics/Object/Matrix.hpp>
 
 #include <OpenSpaceToolkit/Physics/Environment.hpp>
 #include <OpenSpaceToolkit/Physics/Time/Instant.hpp>
@@ -28,10 +30,12 @@ namespace trajectory
 {
 
 using ostk::core::container::Array;
+using ostk::core::container::Map;
 using ostk::core::type::Shared;
 using ostk::core::type::Size;
 
 using ostk::mathematics::curvefitting::Interpolator;
+using ostk::mathematics::object::MatrixXd;
 
 using ostk::physics::Environment;
 using ostk::physics::time::Instant;
@@ -49,6 +53,15 @@ using ostk::astrodynamics::trajectory::state::NumericalSolver;
 class Propagator
 {
    public:
+    /// @brief Per-dynamics contributions recorded at each solver step accepted during a propagation.
+    struct StepContributions
+    {
+        Array<Instant> instants = Array<Instant>::Empty();   ///< One instant per recorded step, in integration order
+        Map<Shared<Dynamics>, MatrixXd> contributions = {};  ///< Per dynamics: rows align with instants, columns
+                                                             ///< follow the dynamics' write coordinate subsets,
+                                                             ///< expressed in GCRF
+    };
+
     /// @brief Default integrator frame
     static const Shared<const Frame> IntegrationFrameSPtr;
 
@@ -117,6 +130,11 @@ class Propagator
     /// @return True if propagator is defined
     bool isDefined() const;
 
+    /// @brief Check if per-dynamics step contribution recording is enabled
+    ///
+    /// @return True if per-dynamics step contribution recording is enabled
+    bool isContributionObservationEnabled() const;
+
     /// @brief Access the coordinate broker
     ///
     /// @return The coordinate broker
@@ -126,6 +144,15 @@ class Propagator
     ///
     /// @return The numerical solver
     const NumericalSolver& accessNumericalSolver() const;
+
+    /// @brief Access the step contributions recorded during the most recent calculate* call.
+    ///        Empty unless recording is enabled. For calculateStateAt / calculateStateToCondition the
+    ///        recorded steps are the solver's accepted steps (initial and final states included); for
+    ///        calculateStatesAt they are the requested instants (each an accepted-step endpoint —
+    ///        intermediate accepted steps are not recorded).
+    ///
+    /// @return The recorded step contributions
+    const StepContributions& accessRecordedStepContributions() const;
 
     /// @brief Get the number of propagated coordinates
     ///
@@ -138,6 +165,13 @@ class Propagator
     /// @endcode
     /// @return An array of dynamics shared pointers
     Array<Shared<Dynamics>> getDynamics() const;
+
+    /// @brief Get the step contributions recorded during the most recent calculate* call
+    /// @code{.cpp}
+    ///              Propagator::StepContributions recorded = propagator.getRecordedStepContributions();
+    /// @endcode
+    /// @return The recorded step contributions
+    StepContributions getRecordedStepContributions() const;
 
     /// @brief Set the dynamics array
     /// @code{.cpp}
@@ -169,6 +203,16 @@ class Propagator
         const Maneuver& aManeuver,
         const Interpolator::Type& anInterpolationType = DEFAULT_MANEUVER_PROPAGATION_INTERPOLATION_TYPE
     );
+
+    /// @brief Enable or disable per-dynamics step contribution recording (disabled by default).
+    ///        When enabled, each calculate* call additionally evaluates every dynamics once per recorded
+    ///        step (one extra RHS-equivalent evaluation per step) and stores the results. Memory grows
+    ///        linearly with the number of recorded steps.
+    /// @code{.cpp}
+    ///              propagator.setContributionObservationEnabled(true);
+    /// @endcode
+    /// @param aContributionObservationEnabled True to enable recording, false to disable it
+    void setContributionObservationEnabled(const bool& aContributionObservationEnabled);
 
     /// @brief Calculate the state at an instant, given initial state
     /// @code{.cpp}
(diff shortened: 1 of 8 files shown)


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -170,6 +204,16 @@ class Propagator
...
 
+    /// @brief Enable or disable per-dynamics step contribution recording (disabled by default).
+    ///        When enabled, each calculate* call additionally evaluates every dynamics once per recorded
+    ///        step (one extra RHS-equivalent evaluation per step) and stores the results. Memory grows
+    ///        linearly with the number of recorded steps.
+    /// @code{.cpp}
+    ///              propagator.setContributionObservationEnabled(true);
+    /// @endcode
+    /// @param aContributionObservationEnabled True to enable recording, false to disable it
+    void setContributionObservationEnabled(const bool& aContributionObservationEnabled);
```
phc1990, line 215: nit on the naming: it might be confusing to both provide a boolean and also have "Enabled" in the name, case then you can have something like

`setContributionObservationEnabled(False)`

maybe `setContributionObservation(bool)` or the pair `enableContributionObservation` and `disableContributionObservation`

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
```diff
@@ -170,6 +204,16 @@ class Propagator
         const Interpolator::Type& anInterpolationType = DEFAULT_MANEUVER_PROPAGATION_INTERPOLATION_TYPE
     );
 
+    /// @brief Enable or disable per-dynamics step contribution recording (disabled by default).
```
phc1990, line 207: some other food for thought, contribution computation might be expensive, but we might only be interested in a particular dynamic contribution

say we have 70x70,sun,moon,drag,thruster but might only be interested in the thruster contribution

maybe we could

```propagator.observeDynamicContribution(myThruster)```

it's compatible with the bulk enable/disable imo, rather than a single `contributionObservationEnabled_` flat we could have a list of dynamics for which to observe the contribution

if one sets the flag to true the list is set to all dynamics, otherwise the list is emptied

File: include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp [hunk 3]
```diff
@@ -127,6 +145,15 @@ class Propagator
...
 
+    /// @brief Access the step contributions recorded during the most recent calculate* call.
+    ///        Empty unless recording is enabled. For calculateStateAt / calculateStateToCondition the
+    ///        recorded steps are the solver's accepted steps (initial and final states included); for
+    ///        calculateStatesAt they are the requested instants (each an accepted-step endpoint —
+    ///        intermediate accepted steps are not recorded).
+    ///
+    /// @return The recorded step contributions
+    const StepContributions& accessRecordedStepContributions() const;
+
```
phc1990, line 156: Maybe controversial take, but I think this will age better:

I would not decouple states and contributions.
States can be expanded at will, and I think this is a good use-case of it.

Putting this + the other thoughts together:

```
propagator = Propagator(...)

dyn_1_contribution_subset = propagator.observeContribution(dynamics_1)
dyn_2_contribution_subset = propagator.observeContribution(dynamics_2)

states = propagator.calculate_states_at(...)

for state in states:
    dyn_1_contribution = state.extract_coordinate(dyn_1_contribution_subset)
    dyn_2_contribution = state.extract_coordinate(dyn_2_contribution_subset)

```

With `propagator.observeContribution(Dynamics)` internatlly creating the 3-D cartesian acceleration corodinate subset under the hood and returning it as a handle to the user.

With `propagator.calculate_states_at(...)` first propagating, then computing the desried contributions (logic you already have) and finally expanding the states with said contributions before returning them.

------

This way we don't store in the propagator info from a previous round, and we use the additional state coordinates.

This way a contribution cand also be rotated as part of the in_frame function, etc.

phc1990 (commented):
Nice stuff left some ideas, lmk

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
vishwa2710: I see where you're getting at. I think the complexity is that each dynamic can update more than one coordinate subset.

For example, a maneuver dynamics will output cartesian acceleration and mass flow rate. So I don't think it we would want to create a new coordinate subset for each dynamic.

Have to think more about this...

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
vishwa2710: Good idea!

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Propagator.hpp
phc1990: Good point, thought a bit more about an option that allows us to have "it all":
- easy usage
- contributions returned as part of the state (for rotation purposes, etc.)
- no storing of contributions in the propagator (compute and forget)

This is the best I coudl come up with:

**1. Create a helper class that handles internal mapping for you: `DynamicsContributionInterpreter/Info/Helper/Broker/...`**

(gonna use 'Broker' for this example)

```python
class DynamicsContributionBroker:
    self._private_map = {}

    def extract_contribution_states(
        dynamics: Dynamics,
        states: list[State],
    ) -> list[State]:
        ...

    def extract_contributions(
        dynamics: Dynamics,
        states: list[State],
        coordinate_subsets: list[CoordinateSubset]
    ) -> list[VectorXd]:
        ...
```

**2. One can simply register dynamics for contribution observation, and then obtain the helper class:**

```python

propagator = Propagator(...)

propagator.observe_contribution(atmospheric_drag)
propagator.observe_contribution(thruster)

contributions_broker = propagator.get_contributions_broker()
``` 

☝️ For each `propagator.observe_contribution(...)` the propagator has internally created the auxiliary coordinate subsets, and linked them to their corresponding derivative:

- After `propagator = Propagator(...)`, map si empty:

```
_private_map = {}
```

- After `propagator.observe_contribution(atmospheric_drag)`

```
_private_map = {
    atmospheric_drag: {
        CartesianVelocity.default(): CarteianAcceleration("ATMOSPHERIC_DRAG_CONTRIBUTION"),
    },
}
```

- After `propagator.observe_contribution(thruster)`

```
_private_map = {
    atmospheric_drag: {
        CartesianVelocity.default(): CoordinateSubset("ATMOSPHERIC_DRAG_CONTRIBUTION"),
    },
    thruster: {
        CartesianVelocity.default(): CoordinateSubset("THRUSTER_VELOCITY_CONTRIBUTION"),
        CoordinateSubset.mass(): CoordinateSubset("THRUSTER_MASS_CONTRIBUTION"),
    },
}
```

Now:
- This mapping is all hidden from the user, they don't need to know about the custom contribution coordinate subsets 
- The names (and sizes) of the contribution coordinate subsets can be created programatically, e.g. `dynamics.name + subset.name + '_CONTRIBUTION'` and the size is that of the subset they point to (e.g. 3 for velocity, 1 for mass)
- Today one can create a CoordinateSubset just by `CoordinateSubset(name, size)` so we could use that. However it does not support in frame rotation. Don't know if this is something we want to support (as a pure rotation). This is something we have to think about, rotation of contributions, and what happens if we don't support it.

**3. Accessing the contributions**

```python
broker = propagator.get_contributions_broker()

states = propagator.calculate_states_at(...)
``` 

Two ways to obtain a contribution (this is an implementation detail, we can do one, the other, or both):

```python
thrust_contribution_states = broker.extract_contribution_states(thruster, states)

# These are now 4-dimensional derivative states (velocity, mass):

for thruster_contribution_state in thrust_contribution_states:
    acceleration = thruster_contribution_state.extract_coordinate(
        CartesianVelocity.default()
    )

    mass_flow_rate = thruster_contribution_state.extract_coordinate(
        CoordinateSubset.mass()
    )
```

...or bypassing the state, and simply getting the contributions:


```python
for state in states:
    acceleration = broker.extract_contributions(
        thruster,
        state,
        CartesianVelocity.default()
    )

    mass_flow_rate = broker.extract_contributions(
        thruster,
        state,
        CoordinateSubset.mass()
    )
```

👉  I personally prefere the first one, since once you have a state, you can rotate, extract one or more subsets, expand it with other subsets, etc.

👉 Maybe `extractDerivativeStates` is a better name.

----

I know I made some mistakes in the pseudocode with pluras/singulars, naming, etc, but the overall idea is:

1. Helper class that handles internal mapping
2. One can obtain derivative (contribution) states

----

EDIT (rotation drama):

One thing we have to consider if we're expanding the propagated state to carry the desired contributions is the impact it may have later down the line, i.e.

```python
propagator.observe_contribution(some_dynamic)

states = propagator.calculate_states_at(...)

# Some more logic

for state in states:
    state.in_frame(some_other_frame)  # RAISES AN ERROR
```

Raises an error as it doesn't know how to rotate a contribution. So: if we add contributions, we should state rotation rules not to hamper the entire state.

That is something we should think about ☝️ 

Which brings me to the following rectification:

> ~I personally prefere the first one, since once you have a state, you can rotate, extract one or more subsets, expand it with other subsets, etc.~

I'm now against this. Sure it's nice to have a derivative `State` that contains the derivatives of the subsets.

But a subset is not only a pointer, it's also instructions on how to rotate it.

So if we have a derivate `State` containing `CartesianVelocity.default()` as it subsets and its **derivative** as its coordinates, we now have:
- rotation instruction of a cartesian velocity
- but accelerations as the numbers to rotate

So it would be an inconsistent State.

At that point, its probably best to go for the pattern that bypasses the derivative `State` and gets the contributions straight away:

```python
for state in states:
    acceleration = broker.extract_contributions(
        thruster,
        state,
        CartesianVelocity.default()
    )

    mass_flow_rate = broker.extract_contributions(
        thruster,
        state,
        CoordinateSubset.mass()
    )
```

PR #709, opened 2026-08-18
feat: propagate SGP4 from mean elements, on Vallado's reference implementation

vishwa2710:
## Why

A TLE is a lossy serialization of an SGP4 mean element set: writing one rounds the eccentricity to `1e-7` and the angles to `1e-4` deg.

That is harmless for propagation but fatal for differentiation. On a near-circular orbit a finite-difference step in eccentricity can be smaller than the quantum, so the eccentricity written into the string does not change at all, the propagator does not respond, and the `ex`/`ey` Jacobian columns collapse onto a single direction. `TLESolver` then drives the orbit onto a hyperbola and throws `"Algorithm error."`

## What

**`sgp4::SGP4FullPrecision`** — the eight elements at full precision, with `calculateStateAt` / `calculateStatesAt`. `TLESolver` now fits against this and writes a TLE once, from the converged state, instead of writing and re-parsing one on every iteration.

**libsgp4 → Vallado.** Reaching a propagator that takes elements rather than 69 columns of text meant replacing [dnwrnr/sgp4](https://github.com/dnwrnr/sgp4), installed as a system library, with Vallado's reference SGP4 — the companion code to AIAA 2006-6753, whose `sgp4init()` takes the elements as doubles. It is vendored rather than installed: two self-contained files depending on nothing outside the standard library, so the Dockerfile install block, `FindSGP4.cmake` and the link entry are all gone. Provenance and the exact delta from upstream are in [`thirdparty/vallado-sgp4/README.md`](thirdparty/vallado-sgp4/README.md); the code is byte-identical to the archive but for line endings, three degree signs in comments, and internal linkage on two debug globals.

**One propagator.** `orbit::model::SGP4` now decodes its TLE into a `SGP4FullPrecision` and propagates that. The Alpha-5 satellite-number laundering that libsgp4's parser required goes with it — nothing downstream sees the NORAD field any more.

## The numerical step this costs

Measured against the outgoing libsgp4, 201 samples spanning each window:

| Orbit | Window | Max Δposition | Max Δvelocity |
|---|---|---|---|
| ISS, near-Earth | ±1 d | 0.28 µm | 3.2e-10 m/s |
| ISS, near-Earth | ±7 d | 2.0 µm | 2.3e-09 m/s |
| Sun-synchronous LEO, near-circular | ±3 d | 6.5 µm | 4.8e-09 m/s |
| Geosynchronous, deep space | ±1 d | 0.16 µm | 1.2e-11 m/s |
| Molniya (e = 0.75), deep space resonance | ±1 d | 52 µm | 3.3e-09 m/s |

Two implementations of the same paper, agreeing to well under a millimetre. The reference-value tests in `SGP4.test.cpp` predate the switch and pass unchanged against Vallado.

## Behaviour changes

- Propagation failures now raise `ostk::core::error::RuntimeError` with a decoded reason ("satellite has decayed", "mean motion below zero", …) instead of libsgp4's `SatelliteException` / `DecayedException`. Anyone catching those types specifically will need to widen the catch.
- A TLE that cannot be propagated still fails at `SGP4` construction, not at first `calculateStateAt` — `SGP4::Impl` forces the propagator's initialisation in its constructor to keep that.

## Testing

Added unit tests based on flight data that throw errors.
Two tests that compared the two propagators are now assertions that the TLE path *is* the element path, and assert exact agreement rather than a tolerance.


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/MeanElements.hpp
```diff
@@ -0,0 +1,215 @@
...
+        const;
+
+    /// @brief Calculate the states at given instants.
+    ///
+    /// @param anInstantArray An array of instants.
+    /// @param anOutputFrameSPtr An output frame. Defaults to TEME.
+    /// @return The states at the given instants.
+    Array<State> calculateStatesAt(
+        const Array<Instant>& anInstantArray, const Shared<const Frame>& anOutputFrameSPtr = Frame::TEME()
+    ) const;
```
phc1990: I think this class might fall under the category of a model. I.e. its initialize with information and is capable of providing timestamped states.

Do you think we can/would be useful to make this comply/inherit with trajectory Model?

It might be worth (for future exchangeability)?


```suggestion
    /// @brief Calculate the state at a given instant.
    ///
    /// @param anInstant An instant.
    /// @return The state at the given instant.
    State calculateStateAt(const Instant& anInstant)
        const;

    /// @brief Calculate the states at given instants.
    ///
    /// @param anInstantArray An array of instants.
    /// @return The states at the given instants.
    Array<State> calculateStatesAt(
        const Array<Instant>& anInstantArray
    ) const;
```

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -217,27 +217,29 @@ TLESolver::Analysis TLESolver::estimate(
...
-        Array<State> states;
-        states.reserve(anInstantArray.getSize());
-
-        for (const Instant& instant : anInstantArray)
-        {
-            states.add(sgp4.calculateStateAt(instant));
-        }
-
-        return states;
+        return TLEStateToMeanElements(aState).calculateStatesAt(anInstantArray, tleStateBuilder_.getFrame());
```
phc1990: Then here, don't need to pass the frame to the calculate function (in case this is what was forcing that signature)

```suggestion
        return TLEStateToMeanElements(aState).calculateStatesAt(anInstantArray).inFrame(tleStateBuilder_.getFrame());
```

include/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/MeanElements.hpp
```diff
@@ -0,0 +1,215 @@
...
+    /// @param aBStarDragTerm A B* drag term. Defaults to 0.0.
+    MeanElements(
+        const Instant& anEpoch,
+        const Angle& anInclination,
+        const Angle& aRaan,
+        const Real& anEccentricity,
+        const Angle& anAop,
+        const Angle& aMeanAnomaly,
+        const Derived& aMeanMotion,
+        const Real& aBStarDragTerm = 0.0
```
phc1990: Just checking whether we want to make this optional

Seeing that this is an expanded mean element set in the context of SGP4 theory, wouldn't it make sense to make b* mandatory?

This way it's even more clear that this is part of that theory (since the underlying calculate states at will be using it)

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4.cpp
```diff
@@ -66,47 +49,23 @@ class SGP4::Impl
...
-          SGP4::SanitizeLineForLibsgp4(tle_.getSecondLine())
-      )),
-      temeFrameOfEpochSPtr_(Frame::TEME())
+    : meanElements_(sgp4::MeanElements::FromTLE(aTle)),
+      outputFrameSPtr_(anOutputFrameSPtr)
 {
+    // MeanElements builds its propagator lazily, and the propagator is what rejects an
+    // element set it cannot fly. Force that now: a TLE this model cannot propagate has always
+    // been an error at construction, not one that waits for a caller to ask for a state.
+    this->meanElements_.calculateStateAt(aTle.getEpoch(), Frame::TEME());
```
phc1990: In line with the other comment, `meanElements_` have been obtained from `aTLE` so the `aTle.getEpoch()` would be redundant (as that would be th ecpoh in the `meanElements_` right?)


```suggestion
    this->meanElements_.calculateStateAt();
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4/SGP4FullPrecision.cpp
```diff
@@ -0,0 +1,408 @@
...
+            throw ostk::core::error::RuntimeError(
+                "Cannot initialize SGP4 from the provided mean elements: {}.",
+                Impl::ErrorMessage(this->satelliteRecord_.error)
+            );
+        }
+    }
+
+    State calculateStateAt(
+        const Instant& anEpoch, const Instant& anInstant, const Shared<const Frame>& anOutputFrameSPtr
+    ) const
```
phc1990, line 105: Ok I think I understand the purpose of this function:

so we have:
- calculateStateAt function (this one, actually calling the underlying implementation)

then 2 other functions:
- calculateStateAt (checking sanity/undefined, etc. before calling the underlying implementation one)
- calculateStatesAt (same as above but for an array)

In that case, I'd jus re-suggest (sorry for any repetition) my point of:
1. inheriting from model
2. have 2 public `calculateStateAt` and `calculateStatesAt` functions (only requiring Instant or Array of Instant)
3. having the actual function calling Vallado's imp private (maybe `calculateStateAt(...)`

Private one can have any args it needs to comply with 3rd party ipml (anEpoch, anInstant, anOuptutFrameSPTr, etc.)

phc1990 (commented):
nice stuff, I think this is a pretty clean way to solve it

left only a few comments on structure

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit/Model/SGP4/MeanElements.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    using ostk::astrodynamics::trajectory::orbit::model::sgp4::MeanElements;
+    using ostk::astrodynamics::trajectory::orbit::model::sgp4::TLE;
+    using ostk::astrodynamics::trajectory::State;
+
+    class_<MeanElements>(
+        aModule,
+        "MeanElements",
+        R"doc(
+            SGP4 mean element set, held at full precision, and the propagator that flies it.
```
alex-liang3: this is important information, let's make it more succinct please.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4.cpp
```diff
@@ -34,26 +32,11 @@ using ostk::physics::time::Interval;
...
-    static const String placeholder = "00000";
-
-    return aLine.getSubstring(0, 2) + placeholder + aLine.getSubstring(7, aLine.getLength() - 7);
-}
-
+/// @brief Propagates one TLE.
+///
+/// @details The TLE is decoded into its mean elements once, here, and everything downstream
+/// runs on MeanElements. Both entry points into SGP4 therefore share a single propagator,
+/// and a TLE is never re-parsed from its text.
```
alex-liang3: why would it ever be re-parsed, in the context of this class?

test/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.test.cpp
```diff
@@ -541,3 +541,157 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Estimation_TLESolver, Estimate_Eccentricit
...
+            )
+        );
+
+        ASSERT_NO_THROW(analysis = solver.estimate(observations[0], observations))
+            << "eccentricity " << eccentricity.toString();
+
+        EXPECT_EQ(analysis.solverAnalysis.terminationCriteria, "RMS Update Threshold")
+            << "eccentricity " << eccentricity.toString();
+
+        // The eccentricity must be recovered to within the TLE's own quantum.
```
alex-liang3: > // The eccentricity must be recovered to within the TLE's own ~quantum~.

prefer "(numerical) precision" or even "quantization"; I've never heard of "quantum" used in this context

apply elsewhere too

test/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.test.cpp
```diff
@@ -541,3 +541,157 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Estimation_TLESolver, Estimate_Eccentricit
...
+// Regression, on flight data, for what the synthetic sweep above isolates: 10.4 h of 20 s
+// GNSS-derived states for a 533 km SSO spacecraft, fitted eccentricity 5.9e-5 -- an order of
+// magnitude under the 1e-4 threshold below which a TLE-in-the-loop estimator loses the ex/ey
+// Jacobian block to the text format's 1e-7 eccentricity quantum. Fitting this arc through a
+// written TLE throws "Algorithm error." out of COE::EccentricAnomalyFromTrueAnomaly, the
+// Gauss-Newton step having run off to a hyperbola; through MeanElements it converges in four
+// iterations to the SGP4 floor of the arc.
+//
+// The 20 s sampling is load-bearing -- decimating the arc to 60 s happens to skirt the
+// degeneracy (it converged, but in nine iterations rather than four).
```
alex-liang3: not important imo
```suggestion
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4.cpp
vishwa2710: I think was more in the context of the how the SGP4 model is used within the TLESolver. Will update.

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit/Model/SGP4/SGP4FullPrecision.cpp
```diff
@@ -0,0 +1,286 @@
...
+            A TLE is a lossy serialization of this set: writing one rounds the eccentricity to 1e-7,
+            the angles to 1e-4 deg, the mean motion to 1e-8 rev/day and the epoch to 1e-8 day.
+            That rounding is harmless for propagation but fatal for differentiation — on a near-circular
+            orbit a finite-difference step in eccentricity can be smaller than 1e-7, so the
+            propagator does not respond at all and the Jacobian silently loses rank.
+
+            Estimators should iterate on `SGP4FullPrecision` and serialize a TLE once, at the end, which costs a few
+            tens of millimetres.
+
+            To propagate a TLE that was published, received, or read from a file, use the `SGP4` class.
```
alex-liang3: I think something like this is all we need
```suggestion
            An Orbit Model that holds SGP4 mean elements at full precision.

            This uses the SGP4 propagator, but with the orbital elements stored at machine-precision rather than the truncated precision of the TLE format. 
```

apply to the other docs if you agree

bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Trajectory/Orbit/Model/SGP4.cpp
```diff
@@ -30,9 +31,23 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Trajectory_Orbit_Model_SGP4(pybind11
...
+                them: the eccentricity to 1e-7, the angles to 1e-4 deg, the mean motion to
+                1e-8 rev/day, and the epoch to 1e-8 day (about 0.9 ms). That is the precision the
+                element set was published at, so it costs nothing here — but it makes the model a
+                staircase rather than a smooth function of its inputs. Anything that differentiates
+                the propagator, an estimator above all, wants `SGP4FullPrecision` instead, which takes the
+                same elements as continuous values.
+
+                Underneath there is one propagator: a TLE is decoded into an `SGP4FullPrecision` once, at
+                construction, and propagated from there. Fed the same values the two agree exactly;
+                they differ only in what precision can reach them.
```
alex-liang3:
```suggestion
                SGP4 Orbit Model, propagating a Two-Line Element set.

                Simplified General Perturbations 4. Takes one TLE, or several with validity
                intervals.
                
                To use SGP4 with orbital elements at a higher numerical precision that what can be encoded in a TLE, use `SGP4FullPrecision` instead.

```

include/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.hpp
```diff
@@ -229,11 +231,25 @@ class TLESolver
 
     /// @brief Convert TLE state to TLE
     ///
+    /// @details This is the lossy step: the TLE text format rounds the eccentricity to 1e-7
+    /// and the angles to 1e-4 degrees. It runs once, on the converged state — never inside
+    /// the iteration, where that rounding would flatten the Jacobian.
+    ///
```
alex-liang3:
```suggestion
    /// @details Warning: this truncates orbital elements to the precision of the TLE format. This is only run once, on the converged state.
```

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -217,27 +217,29 @@ TLESolver::Analysis TLESolver::estimate(
         initialGuessTLEState = CartesianStateAndBStarToTLEState(*state);
     }
 
+    // Snap the estimation epoch onto the grid the TLE epoch field can represent (1e-8 day).
+    // The epoch is held fixed for the whole fit, so this costs nothing in conditioning; it
+    // just means the TLE written at the end has nothing left to round in its epoch field,
+    // which would otherwise be worth a few metres of along-track error on its own.
```
alex-liang3:
```suggestion
    // Snap the estimation epoch to the nearest epoch that can be represented by a TLE. 
    // This avoids accuracy loss compared to if the epoch was truncated to TLE precision at the end of the solve.
```

src/OpenSpaceToolkit/Astrodynamics/Estimator/TLESolver.cpp
```diff
@@ -217,27 +217,29 @@ TLESolver::Analysis TLESolver::estimate(
...
             return aState.inFrame(tleStateBuilder_.getFrame());
         }
     );
 
+    // Propagate trial states from the elements themselves, not from a TLE written and
+    // re-parsed on every call. The text format rounds the eccentricity to 1e-7 and the
+    // angles to 1e-4 deg; on a near-circular orbit the finite-difference step in the
+    // eccentricity components falls below that quantization, the propagator does not respond at
+    // all, and the ex/ey Jacobian columns collapse onto one direction. A TLE is written
+    // once, at the end, from the converged state.
```
alex-liang3:
```suggestion
    // Propagate trial states using full-precision elements, rather than TLE text field precision. 
    // This can prevent Jacobian conditioning issues, especially when eccentricity is small.
```

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/SGP4.cpp
```diff
@@ -66,47 +48,20 @@ class SGP4::Impl
...
-          tle_.getSatelliteName(),
-          SGP4::SanitizeLineForLibsgp4(tle_.getFirstLine()),
-          SGP4::SanitizeLineForLibsgp4(tle_.getSecondLine())
-      )),
-      temeFrameOfEpochSPtr_(Frame::TEME())
+    : sgp4FullPrecision_(sgp4::SGP4FullPrecision::FromTLE(aTle, anOutputFrameSPtr))
 {
+    // Constructing the SGP4FullPrecision above builds the propagator, and the propagator is what
+    // rejects an element set it cannot fly, so a TLE this model cannot propagate is already an
+    // error here rather than one that waits for a caller to ask for a state.
```
alex-liang3:
```suggestion
    // Validation is done by the propagator's constructor
```

phc1990 (approved):
NIce stuff 👍

PR #712, opened 2026-09-17
feat: enrich Close Approach with Covariance Matrices

phc1990:
### Context

Up until today, close approaches only contained primary and secondary state information.

`CloseApproach` is already a useful class defining an encounter geometry. But as of today, it lacks covariance information.

### This MR

- Enriches the `CloseApproach` class with optional primary and secondary `CovarianceMatrices`
- Provides utility methods geared towards probability of collision related computation (e.g. compute relative velocity, or scaling mechanisms)

## Summary by CodeRabbit

- **New Features**
  - Added support for associating covariance matrices with each object in a close approach.
  - Added covariance retrieval, updates, and scaling, with optional per-object scale factors.
  - Added relative velocity reporting in meters per second.
  - Added encounter-frame generation with GCRF as the default reference frame.
  - Added the ability to swap the two objects and their associated data.
  - Exposed all new capabilities through the Python interface.

- **Bug Fixes**
  - Added validation to ensure covariance data matches the corresponding state time.


bindings/python/src/OpenSpaceToolkitAstrodynamicsPy/Conjunction/CloseApproach.cpp
```diff
@@ -105,6 +125,62 @@ inline void OpenSpaceToolkitAstrodynamicsPy_Conjunction_CloseApproach(pybind11::
...
+                scaling only one covariance or when a covariance is undefined.
+
+                Args:
+                    scale_factor_1 (float, optional): The scale factor for Object 1 covariance. Defaults to Real.undefined().
+                    scale_factor_2 (float, optional): The scale factor for Object 2 covariance. Defaults to Real.undefined().
+
+                Returns:
+                    CloseApproach: A new Close Approach with scaled covariance matrices.
+            )doc",
+            arg_v("scale_factor_1", Real::Undefined(), "Real.undefined()"),
```
vishwa2710: Implementation detail, let's use `std::optional` instead of ::Undefined when wanting to provide optional arguments.

PR #717, opened 2026-09-19
perf: solve closest approach as a root

vishwa2710:
## What

`FindTimeOfClosestApproach` minimized range with NLopt's COBYLA — a general-purpose derivative-free simplex method — at a relative x tolerance of 1 µs. A range minimum is locally flat, so that is a badly conditioned thing to search for. Profiling put the call at **93% of a 20-target elevation run**, at ~30 objective evaluations per access, each propagating and frame-transforming *two* full states.

Range is smooth across a pass and has a single minimum, so the derivative of its square changes sign exactly once. This solves for that sign change with the existing bracketed `RootSolver` instead.

The residual is `dx · dv` — half the derivative of the squared range. It shares its sign and its root with the range rate but needs no division, so it stays finite when the two trajectories coincide and the range is zero (which `ComputeAccesses_4` exercises).

Two cases a bracketed solve must handle explicitly, where a minimization did not:

- An access clipped by the analysis interval can begin with the satellite already receding, or end with it still approaching. There is then no sign change, and the closest approach sits on the corresponding endpoint.
- An access captured by a single coarse sample has a zero-length interval and nowhere to search, so that sample is the answer. This incidentally fixes `coarse = true` raising `Cannot find TCA (solution did not converge)` on such an access.

Also threads the fixed target's celestial-frame coordinates down from `computeAccessesForFixedTargets`. A fixed target does not move in that frame — constant position, zero velocity — so supplying them removes the second state propagation and transform from every iteration of the search, and from the max-elevation evaluation.

NLopt is no longer used anywhere in the library.

## Measured

Benchmark suite, same build before and after the change:

| Scenario | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| 100 targets, 1 week, Elevation | 9323 ms | 500 ms | **18.7×** |
| 100 targets, 1 day, Elevation | 1334 ms | 78 ms | **17.1×** |
| 10 targets, 1 day, Elevation | 137 ms | 10 ms | 13.4× |
| 10 targets, 1 day, AER | 143 ms | 11 ms | 13.2× |
| 1 target, 1 day, Elevation | 19.7 ms | 3.9 ms | 5.0× |
| 10 targets, 1 day, LineOfSight | 799 ms | 656 ms | 1.2× |
| 1 target, 2 weeks, GCRF out | 2494 ms | 2355 ms | 1.1× |

Line-of-sight and GCRF move least because they are dominated by crossing refinement and by the coarse scan respectively, not by TCA. The speedup exceeds what the evaluation count alone predicts, because the search now visits far fewer distinct instants, which also relieves the frame-transform cache.


test/OpenSpaceToolkit/Astrodynamics/Access/Generator.test.cpp
```diff
@@ -732,6 +732,66 @@ TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, ComputeAccesses_3)
     }
 }
 
+TEST_F(OpenSpaceToolkit_Astrodynamics_Access_Generator, TimeOfClosestApproachIsARangeMinimum)
+{
+    // The time of closest approach is solved for as the sign change of the range derivative rather than by
+    // minimizing the range. Assert the property that defines it: the range at the reported instant is smaller
+    // than the range just before and just after it.
```
vishwa2710:
```suggestion
    // minimizing the range. The range at the reported instant is smaller
    // than the range just before and just after it.
```

phc1990 (approved):

include/OpenSpaceToolkit/Astrodynamics/Access/Generator.hpp
```diff
@@ -441,11 +447,14 @@ class Generator
         const bool& coarse = false
     ) const;
 
+    /// @param aFixedFromPositionCoordinates The target's coordinates in the central celestial object's frame, when
```
apaletta3, line 450: nit: add the rest of the params docstring if you're going to to add one

include/OpenSpaceToolkit/Astrodynamics/Access/Generator.hpp
```diff
@@ -465,22 +474,30 @@ class Generator
...
-        const Shared<const Celestial>& aCelestialSPtr
+        const Shared<const Celestial>& aCelestialSPtr,
+        const std::optional<Vector3d>& aFixedFromPositionCoordinates = std::nullopt
     );
 
+    /// @brief Find the time of closest approach within an access interval.
+    ///
+    /// @details Range is smooth across a pass and has a single minimum, so its time derivative changes sign exactly
+    /// once. This solves for that sign change rather than minimizing the range itself: a range minimum is locally
+    /// flat, which makes it expensive to localize, whereas the derivative crosses zero transversally.
```
apaletta3, line 485: nit: transversally? not sure what that means in this context, can you clarify

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -1035,114 +1050,95 @@ Instant Generator::FindTimeOfClosestApproach(
...
+    const Shared<const Celestial>& aCelestialSPtr,
+    const std::optional<Vector3d>& aFixedFromPositionCoordinates
 )
 {
-    struct Context
+    const Shared<const Frame>& celestialFrameSPtr = aCelestialSPtr->accessFrame();
+
+    const Instant startInstant = anAccessInterval.getStart();
+
+    // Half the derivative of the squared range, d(|dx|^2)/dt / 2 = dx . dv, at `aDurationFromStart` seconds into
```
apaletta3, line 1061: nit: Why do you mention half here? I don't see a factor of half in the equations below.
Also can you clarify the docstring to read more like: this is what we're using as the root finding evaluation function and this is why

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -1035,114 +1050,95 @@ Instant Generator::FindTimeOfClosestApproach(
...
-    {
-        (void)aGradient;
-        if (aDataContext == nullptr)
-        {
-            throw ostk::core::error::runtime::Wrong("Data context", "nullptr");
-        }
+        const Instant instant = startInstant + Duration::Seconds(aDurationFromStart);
 
-        const Context* contextPtr = static_cast<const Context*>(aDataContext);
+        const State toState = aToTrajectory.getStateAt(instant).inFrame(celestialFrameSPtr);
```
apaletta3, line 1075: this might have performance implications for satellite to satellite accesses that we're going to generate in the near future: we may want also provide an argument for this `FindTimeOfClosestApproach` function to allow the range rate to be computed in ITRF or GCRF. Here, ITRF is clearly advantageous if one or both targets are earth fixed, but if both targets are inertial then it doesn't make sense anymore.

Food for thought, don't have to implement this just yet

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
apaletta3: also can you fix the notation: range should be rho

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -1035,114 +1050,95 @@ Instant Generator::FindTimeOfClosestApproach(
...
-
-    optimizer.set_lower_bounds(lowerBound);
-    optimizer.set_upper_bounds(upperBound);
+    const double accessDuration_s = anAccessInterval.getDuration().inSeconds();
 
-    optimizer.set_min_objective(calculateRange, &context);
-
-    optimizer.set_xtol_rel(aTolerance.inSeconds());
+    // An access captured by a single coarse sample has a zero-length interval and nowhere to search.
+    if (accessDuration_s <= 0.0)
```
apaletta3, line 1098: Shouldn't it be this
```suggestion
    if (accessDuration_s == 0.0)
```
or checking within a tolerance? How can the interval be negative?

apaletta3 (approved):
Nice! A couple small comments

include/OpenSpaceToolkit/Astrodynamics/Access/Generator.hpp
vishwa2710: It means it crosses it only once, will update the docstrings.

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
vishwa2710: Yeah I agree. Added a TBI to improve. I think in the future we should:
1. Make output frame a mandatory field for a trajectory model
2. Based on the output frame, we can decide what frame we want to solve in
3. If both trajectories have different output frames, we can just give a warning and use ITRF.

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
vishwa2710: It can't it's mostly just for floating point safety since even zero can have a sign https://en.wikipedia.org/wiki/IEEE_754

PR #718, opened 2026-09-19
perf: refine crossings on a continuous residual, not a boolean

vishwa2710:
## What

`computePreciseCrossings` fed `RootSolver` a function returning `+1.0 : -1.0`. TOMS748's interpolation steps are worthless on a step function, so it degenerates to bisection — measured at about **29 criterion evaluations per crossing** to reach the 1 µs tolerance.

This replaces the boolean condition with a **signed residual**: positive while the criterion is satisfied, negative while it is not. A crossing is still a sign change, but the solver now sees the geometry instead of a step.

Only the elevation-interval criterion gets a genuine margin: `min(elevation - lower, upper - elevation)` — the distance in radians to whichever bound is nearer, which handles a crossing of either. AER interval, AER mask and line-of-sight keep the boolean expressed as ±1, so their behaviour is unchanged. Line-of-sight in particular is an intersection test with no natural margin; giving it one needs a separate surrogate, such as the segment's perpendicular distance to the ellipsoid.

## Measured

**On its own this is worth almost nothing**

| Scenario | main | this PR | Speedup |
| --- | ---: | ---: | ---: |
| 100 targets, 1 week, Elevation | 9323 ms | 9017 ms | 1.03× |
| 10 targets, 1 day, Elevation | 137 ms | 136 ms | 1.01× |
| 1 target, 1 day, Elevation | 19.7 ms | 12.9 ms | 1.52× |
| everything else | | | 1.00× |

Crossing refinement is only ~5% of an elevation run, while closest approach is 93% of it, so the improvement is invisible underneath.

It becomes worth having once that 93% is addressed. Stacked on top of #717:

| Scenario | main | #717 alone | #717 + this | this PR's contribution |
| --- | ---: | ---: | ---: | ---: |
| 100 targets, 1 week, Elevation | 9323 ms | 500 ms (18.7×) | **231 ms (40.4×)** | 2.17× |
| 100 targets, 1 day, Elevation | 1334 ms | 78 ms (17.1×) | **34 ms (39.4×)** | 2.30× |
| 10 targets, 1 day, Elevation | 137 ms | 10.2 ms (13.4×) | **5.5 ms (24.8×)** | 1.85× |
| 10 targets, 1 day, AER | 143 ms | 10.8 ms (13.2×) | 10.4 ms (13.7×) | 1.04× |
| 10 targets, 1 day, LineOfSight | 799 ms | 656 ms (1.2×) | 639 ms (1.25×) | 1.03× |

AER and line-of-sight barely move, as expected — they keep the boolean.

The two PRs are independent and touch different functions; they merge cleanly and the combined build passes all 75 tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the precision of visibility crossing calculations for access intervals.
  - Elevation-based access boundaries now account continuously for both lower and upper limits, producing more accurate start and end times.
  - Updated visibility checks for angular, masking, and line-of-sight constraints to improve boundary detection.


src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -849,25 +853,32 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
...
             anAccessTarget.accessVisibilityCriterion().as<VisibilityCriterion::ElevationInterval>().value();
 
-        condition = [&fromPositionCoordinate_ITRF, &aToTrajectory, &aCelestialSPtr, visibilityCriterion](
-                        const Instant& instant
-                    ) -> bool
+        const double lowerBound_rad = visibilityCriterion.elevation.accessLowerBound();
+        const double upperBound_rad = visibilityCriterion.elevation.accessUpperBound();
+
+        // Distance to whichever elevation bound is nearer, in radians. Positive between the bounds, negative
+        // outside them, and zero exactly on a crossing - so the solver sees the geometry rather than a step.
```
apaletta3:
```suggestion
        // outside them, and zero exactly on a crossing - so the solver sees a continuous function and can converge on its value as well as its derivate.
```
is this a more accurate description of what's happening?

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -796,7 +796,11 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
 
     const Matrix3d SEZRotation = anAccessTarget.computeR_SEZ_ECEF(aCelestialSPtr);
 
-    std::function<bool(const Instant&)> condition;
+    // A signed residual: strictly positive while the criterion is satisfied, negative while it is not, so that a
+    // crossing is a sign change. Where the criterion admits a continuous margin the residual carries it, and the
+    // root solve converges on the real function; otherwise it degrades to the boolean +/-1 step, on which the
```
apaletta3: I don't know what this means, can you clarify?
"Where the criterion admits a continuous margin the residual carries it"

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
apaletta3: feel free to add that to the comment at the function delcaration instead

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
```diff
@@ -878,7 +889,7 @@ Array<physics::time::Interval> Generator::computePreciseCrossings(
 
             const double elevation_rad = std::asin(dx.dot(fromPositionCoordinate_ITRF.normalized()) / dx.norm());
```
apaletta3: could be nice to add a quick diagram to show what the value of the function is over time as you get close to an access, for building intuition

apaletta3 (approved):
Nice! All nits

src/OpenSpaceToolkit/Astrodynamics/Access/Generator.cpp
vishwa2710: Going to replace this with a derivative based method in a follow up MR so will skip out on the diagram for now. I've updated the comments appropriately.

PR #723, opened 2026-09-20
perf: hoist the invariant gradient out of the Q-Law effectivity sweep

vishwa2710:
When an effectivity threshold is configured, every Q-Law evaluation sweeps 50 true anomalies and builds the full thrust vector at each one, so the current thrust direction can be scored against the best and worst directions available over the orbit.

`computeThrustVector` is the product of two factors:

- **∂Q/∂oe** — depends only on the five non-anomalistic elements
- **∂oe/∂F** (the Gauss variational matrix) — depends on the true anomaly

The sweep holds the five elements fixed and varies only the true anomaly. So `∂Q/∂oe` was being recomputed **identically 50 times per call**. With `GradientStrategy::FiniteDifference`, each of those recomputations is itself 10 evaluations of Q.

## Change

Evaluate it once before the loop and multiply by the per-anomaly Gauss matrix inside. The arithmetic and its order are unchanged, so results are bit-identical; the NaN guard moves with the computation.

## Measurement

On the Q-Law orbit raise from #721 (best of 3, analytical gradient):

| | main | this branch |
| --- | --- | --- |
| `Segment::solve` | 8.53 s | **7.27 s** |
| `QLaw::calculateThrustAccelerationAt` | 77.9 µs | **64.9 µs** |
| `extractManeuvers` | 86.6 ms | **59.5 ms** |

## Interaction with #724

That PR adds a `double` overload of `Compute_dOE_dF`. This branch deliberately calls the existing `Derived` overload so it stands alone. Once both land, the call site introduced here can adopt the cheaper overload — a one-line follow-up, not a conflict; the two touch different functions and merge cleanly in either order.

## Summary by CodeRabbit

- **Performance**
  - Improved the efficiency of guidance-effectiveness calculations by avoiding repeated gradient evaluation during anomaly sweeps.
  - Preserved existing thrust-vector behavior while adding validation for invalid numerical results before calculation proceeds.


src/OpenSpaceToolkit/Astrodynamics/GuidanceLaw/QLaw.cpp
```diff
@@ -864,14 +864,24 @@ Tuple<double, double> QLaw::computeEffectivity_(
     Vector6d coeVector = aCOEVector;
     VectorXd dQ_dt(trueAnomalyAngles.size());
 
+    // ∂Q/∂oe only depends on the five non-anomalistic elements, which are held fixed over the
+    // sweep below. Evaluate it once here instead of once per true anomaly: only the Gauss
+    // variational matrix ∂oe/∂F depends on the true anomaly.
```
vishwa2710:
```suggestion
    // sweep below.
```

apaletta3 (approved):

PR #724, opened 2026-09-20
perf: add a double overload of QLaw::Compute_dOE_dF

vishwa2710:
`Compute_dOE_dF` is the Gauss variational matrix. With an effectivity threshold configured it runs **51 times per Q-Law evaluation** — once for the current true anomaly and once for each point of the 50-point sweep.

Each call converted the gravitational parameter out of its `Derived` unit, then went through `COE::ComputeSemiLatusRectum`, `COE::ComputeAngularMomentum` and `COE::ComputeRadialDistance`, whose arguments and results are `ostk::core::type::Real`.

`Real` is not a zero-cost wrapper over `double`: its operators are defined in `libopen-space-toolkit-core.so`, so they cannot inline across the shared-library boundary, and each one branches on undefined and infinite operands and calls further out-of-line predicates (`isZero`, `isStrictlyPositive`, `getSign`). For three one-line formulas evaluated 51 times per guidance law call, that overhead dominates the formulas themselves.

## Change

Add an overload taking the gravitational parameter as a `double` and inline the three helper expressions on plain doubles. `QLaw` already stores `mu_` in SI units, so the hot path passes it straight through.

The existing `Derived` overload is kept and simply forwards, so **no caller or binding has to change** — including the Python bindings and the existing `Compute_dOE_dF` unit test.

## Measurement

This change is small enough that it needed care: the container drifts by more than 10% across tens of minutes, which is the same order as the effect. Both builds were therefore measured **adjacently and with repetitions**, rather than against an earlier reading.

| `QLaw::calculateThrustAccelerationAt`, 15 reps | merge base | this branch |
| --- | --- | --- |
| mean | 92.8 µs | **85.5 µs** |
| median | 93.4 µs | **88.3 µs** |

`Segment::solve` came out at 8.0 s against 8.5–9.2 s for the merge base.

## Relationship to the other Q-Law PR

#723 hoists `∂Q/∂oe` out of the same sweep. The two are independent — they touch different functions and merge cleanly in either order — and they compose: this one makes each of the 51 Gauss-matrix calls cheaper, that one removes 50 redundant gradient evaluations.


Voyagerroc-Lab (commented):
Reviewed the double overload, direct orbital-element formulas, forwarding from Derived, and the pybind overload disambiguation. The hot path now avoids repeated unit/Real conversions while retaining the existing Derived API and binding, and the documented numerical semantics are unchanged. The measured improvement is consistent with the targeted allocation/conversion overhead. I found no blocking issue in the current diff.

include/OpenSpaceToolkit/Astrodynamics/GuidanceLaw/QLaw.hpp
apaletta3, line 1: why are we still dragging around an identical copy of `mu_` and `gravitationalParameter_` on line 292 in this file? Shoudn't we remove the redundany or its more performant to keep both around so you can use each one when it is needed?

apaletta3 (approved):
LGTM, but one question for you

include/OpenSpaceToolkit/Astrodynamics/GuidanceLaw/QLaw.hpp
vishwa2710: So one is the SI units, and one is the Derived one. We're maintaining the Derived (public) facing interface for now, so have to maintain both values internally for performance so that we're not constantly converting back and forth.

PR #726, opened 2026-09-20
perf: solve Kepler's equation on doubles instead of Real

vishwa2710:
`COE::EccentricAnomalyFromMeanAnomaly` runs a Newton-type refinement of up to 1000 iterations, and its two helper lambdas — `keplerstart3` and `eps3` — took and returned `ostk::core::type::Real` **by value**, about 15 boxed operations per iteration.

## Change

Declare the lambdas and the loop variables as `double`. The arithmetic is identical, so results are bit-identical. The `Real::Undefined()` sentinel for `E` becomes quiet NaN and is still assigned before use on every path. The public signature `(Angle, Real, Real)` is untouched.

## Measurement

Stacked on #725, where the profile says it should matter, it is unambiguous (15 repetitions each, built back to back):

| `QLaw::calculateThrustAccelerationAt` | #725 alone | #725 + this |
| --- | --- | --- |
| mean | 70.6 µs | **51.6 µs** (−27%) |
| median | 70.5 µs | **51.9 µs** |
| standard deviation | 0.8 µs | 1.1 µs |


src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -983,43 +985,47 @@ Angle COE::EccentricAnomalyFromMeanAnomaly(
...
+        const double t35 = e * t34;
+        const double t33 = std::cos(M);
 
         return M + (-0.5 * t35 + e + (t34 + 1.5 * t33 * t35) * t33) * std::sin(M);
     };
 
     // An iteration (correction) method to solve Kepler's equation
 
-    auto eps3 = [](Real e, Real M, Real x) -> Real
+    auto eps3 = [](const double& e, const double& M, const double& x) -> double
```
maxnqt, line 1002: Isn't `double e, double M, double x` more idiomatic?

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
```diff
@@ -983,43 +985,47 @@ Angle COE::EccentricAnomalyFromMeanAnomaly(
...
+    const double eccentricity = anEccentricity;
+    const double tolerance = aTolerance;
 
-    const Real M = meanAnomaly_rad;
-    const Real Mnorm = std::fmod(M, 2.0 * M_PI);
+    const double M = aMeanAnomaly.inRadians();
+    const double Mnorm = std::fmod(M, 2.0 * M_PI);
 
-    Real E = Real::Undefined();
+    double E = std::numeric_limits<double>::quiet_NaN();
```
maxnqt, line 1020: `double E = std::numeric_limits<double>::quiet_NaN()` isn't strictly equivalent to `Real::Undefined()`. Maybe `std::optional<double>` is closer to `Real::Undefined()` semantically?

maxnqt (approved):

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
vishwa2710: Technically both `const double&` and `double` consume the same amount of memory so I don't think it makes much difference, but you're right it's more idiomatic.

src/OpenSpaceToolkit/Astrodynamics/Trajectory/Orbit/Model/Kepler/COE.cpp
vishwa2710: Maybe but it's much more grotesque to write a `value_or` everywhere. Mostly we just want to set it to some default value as it's going to get immediately overwritten.

PR #727, opened 2026-09-22
feat: compute the angular velocity of custom pointing profiles by finite difference

vishwa2710:
## Summary

The angular velocity of `Profile::CustomPointing` profiles was hardcoded to zero (`TBM: Artificially set to 0 for now`). It is now computed by finite difference of the generated orientation:

- **Central difference** with a 0.1 s step everywhere the orientation is defined on both sides.
- **Forward difference** at the start and **backward difference** at the end of the time range over which the orbit and the targets are defined (e.g. a tabulated orbit or target trajectory). A boundary is detected by the evaluation at `t ± h` throwing; the evaluation at `t` itself is not guarded, so genuine errors still propagate.
- A clear error is raised if the orientation is undefined on both sides (time range shorter than 0.2 s).

The two orientations are combined as `q_B_GCRF(t2) * q_B_GCRF(t1).conj`, following the convention already used by `Orbit::getOrbitalFrame`, and the rotation angle is extracted with `atan2` rather than `acos` (ill-conditioned for small angles, capping the precision at ~1e-11 rad/s). The relative quaternion is rectified first so that rectified inputs straddling a sign flip cannot yield a near-2π angle.

## Behaviour changes

- Each custom pointing state now requires three orbit/orientation evaluations instead of one (Python orientation generators are called three times per state).
- Body frames built with `constructBodyFrame` from a custom pointing profile now carry an angular velocity, so velocities transformed into that frame include the rotational term.

## Tests

- Existing `CustomPointing` test now also compares the angular velocity against the STK-validated VNC local orbital frame (agreement to 3e-11 rad/s, tolerance 1e-10).
- New `CustomPointing_AngularVelocity` test:
  - nadir/velocity pointing on an unperturbed circular orbit vs. the analytical `(0, -n, 0)` (error ~1e-15 rad/s),
  - a custom generator spinning at a constant rate about a fixed body axis (error ~2e-14 rad/s),
  - a constant orientation (exactly zero),
  - a tabulated orbit evaluated at its start, within one step of its start, in the middle, within one step of its end, and at its end (all within 9e-11 rad/s; evaluation outside the range still throws).
- New Python test checks the angular velocity is non-zero.
- Full C++ suite: 1121 passed, 2 pre-existing skips. Full Python suite: 851 passed.


include/OpenSpaceToolkit/Astrodynamics/Trajectory/Model.hpp
phc1990: Took me a bit to understand why we needed the distinction between errors, so that one can determine if we can center/forward/backwards differentiate.

I fully agree with the central/forward/backwards differentiation, however I think we can do without the errors just to be able to tell the difference.

Rather than the 3 errors we could a check functionality like this:

```suggestion

    virtual bool isDefinedAt(const Instant& anInsant) const = 0;

```

Or some other naming if this one it's confusing, maybe `isApplicableAt(...)`

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile.cpp
```diff
@@ -355,16 +362,76 @@ Profile Profile::CustomPointing(
     auto dynamicProviderGenerator = [anOrbit, anOrientationGenerator](const Instant& anInstant) -> Transform
     {
         const State state = anOrbit.getStateAt(anInstant).inFrame(DEFAULT_PROFILE_FRAME);
+        const Quaternion q_B_GCRF = anOrientationGenerator(state);
```
phc1990: Nit: some of these  variables (e.g. `q_B_GCRF`) follow snake_case, I guess it's for readability

src/OpenSpaceToolkit/Astrodynamics/Flight/Profile.cpp
```diff
@@ -355,16 +362,76 @@ Profile Profile::CustomPointing(
...
+        {
+            try
+            {
+                return orientationAt(anInstant + step);
+            }
+            catch (const TrajectoryModel::AfterEndError&)
+            {
+                return std::nullopt;
+            }
+        }();
```
phc1990: Then, rather than rely on error types we can pre-check if it's defined before/at/after and then choose which one of the two defined instants we choose

include/OpenSpaceToolkit/Astrodynamics/Flight/Profile.hpp
```diff
@@ -389,6 +396,12 @@ class Profile
```
phc1990: Nit: maybe we can provide this already in case people have super slow (or super fast) attitude laws, they would be able to choose high step sizes or low step sizes respectively to account for numericall precision or missing rate changes.

```suggestion

   static Profile CustomPointing(
        const trajectory::Orbit& anOrbit, const std::function<Quaternion(const State&)>& anOrientationGenerator, const Duration& anAngularVelocityDifferentiationStepSize
    ); 

```

Then we can have the current constructor call this one with the `DEFAULT_...` value

phc1990 (approved):

